PR #7, opened 2020-08-17
Refactor into dedicated modules

### LINKED ISSUE #1: Dedicated module per package
The current setup solves the issue for now where we were copying over internal packages from one project to another to gain the same functionalities.

The new problem we face however is the fact that if you for example only want to make use of the `ssh` package, you are now automagically pulling in Git related dependencies because of the `git` package.

We should look into creating dedicated modules per package so that they can all have their own set of dependencies and release cycle.

hiddeco:
Fixes #1.


stefanprodan:
I think we need to create a checksum of all packages in CI https://github.com/fluxcd/pkg/blob/master/.github/workflows/build.yaml#L19

hiddeco:
Just for public visibility: the above suggested changes was not required as the `hashFiles('**/go.mod')` includes _any_ matching file.

PR #17, opened 2020-09-09
Factor change predicate out of fluxcd/source-controller

squaremo:
Many of the controllers in gitops toolkit follow the pattern of running on an interval, except when there's a specific annotation saying to run ASAP.

The "except" is encapsulted in a predicate, which filters Update events for those in which the generation has changed OR the specific annotation has changed.

---

Possible gotcha here: the name of the annotation is taken from github.com/fluxcd/source-controller/api, which is its own module; if other code refers to this module, it'll drag in controller-runtime, which wouldn't be the case for source-controller/api. If this is a problem, that const can be put in another module.


stefanprodan:
All API packages are importing `sigs.k8s.io/controller-runtime/pkg/scheme` so I don't see any issues with having the annotation in this package. Can you please run go mod tidy and commit go.sum?

stefanprodan:
I think the logger package could be part of the runtime as well.

stefanprodan (approved):
LGTM

squaremo:
Follow-up:

 - [ ] change imports of logger
 - [ ] use `ChangePredicate` from controllers and remove any duplicates
 - [ ] use `ReconcileAtAnnotation` from controllers and remove any duplicates

stefanprodan:
Can you please run make release-runtime so we can use a v0.0.1 as the package version in controllers.

PR #19, opened 2020-09-10
Fix GH Actions version default

stefanprodan:


hiddeco (approved):
:imp: is in the details

PR #24, opened 2020-09-16
FOSSA scan enabled

idvoretskyi:
Ref.: https://github.com/cncf/foundation/issues/109

Signed-off-by: Ihor Dvoretskyi <ihor@linux.com>

### DIFF
--- a/.github/workflows/fossa.yml
+++ b/.github/workflows/fossa.yml
@@ -1,0 +1,27 @@
+name: FOSSA
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


idvoretskyi:
cc @stefanprodan @dholbach

stefanprodan:
Each directory in this repo is a standalone go package, I think fossa should detect that there are many go.mod files and scan each one?

idvoretskyi:
@stefanprodan ah, I see. No, I assume FOSSA is able to scan code only, so it's irrelevant here.

Thanks for clarifying.

stefanprodan:
All these packages are included in source/kustomize/helm controllers so we scan them there, all good 👍

PR #25, opened 2020-09-29
Introduce apis/meta/v1 package

hiddeco:
* Includes / removes annotations and reason from `runtime`
* Introduces `Condition` type used across controllers


apis/meta/v1/conditions.go
```diff
@@ -0,0 +1,43 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1
+
+const (
+	// ReadyCondition is the name of the Ready condition
+	// implemented by all toolkit resources.
+	ReadyCondition string = "Ready"
```
stefanprodan, line 22: Let’s move the Condition type here as well

hiddeco: There is a side-effect to this: it will no longer show up on the auto generated API documentation (e.g. https://toolkit.fluxcd.io/components/source/api/#source.toolkit.fluxcd.io/v1alpha1.SourceCondition) but becomes a link to the godoc.org page for the struct.

stefanprodan: When we will move to the Condition type from 1.19 will it show up in API docs?

hiddeco: No, same story.

stefanprodan: Ok then I don't see why we would keep that type duplicated in each controller, as we will switch to 1.19 eventually.

stefanprodan (approved):
LGTM

apis/meta/v1/condition_types.go
```diff
@@ -0,0 +1,96 @@
...
+	DependencyNotReadyReason string = "DependencyNotReady"
+
+	// SuspendedReason represents the fact that the reconciliation of a toolkit
+	// resource is suspended.
+	SuspendedReason string = "Suspended"
+)
+
+// InReadyCondition returns if the given Condition slice has a ReadyCondition
+// with a 'True' condition status.
+func InReadyCondition(conditions []Condition) bool {
```
stefanprodan, line 78:
```suggestion
func HasReadyCondition(conditions []Condition) bool {
```

squaremo:
> As they are not really part of the runtime, but rather API elements.

I don't mind so long as they are defined in a single, neutral place.

runtime/consts.go
```diff
@@ -20,28 +20,4 @@ const (
 	// ReconcileAtAnnotation is the annotation used for triggering a
 	// reconciliation outside of the defined schedule.
 	ReconcileAtAnnotation string = "fluxcd.io/reconcileAt"
```
squaremo, line 22: Why didn't this get removed here?

apis/meta/v1/annotation_types.go
```diff
@@ -0,0 +1,23 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1
+
+const (
```
squaremo, line 19: The filename is a bit odd -- no types here (and I wouldn't expect any).

runtime/consts.go
hiddeco: I can't introduce the dependency (including the replace) without Go Modules complaining about a non-existing version.

squaremo: Lol, of course, it's the three step shuffle :-)

apis/meta/v1/annotation_types.go
stefanprodan: Your are reviewing an outdate commit. There is a type there

PR #27, opened 2020-09-29
Introduce `errors` to `runtime` package

hiddeco:
This PR adds commonly found errors in the toolkit controllers so
that they produce (rich) and consistent error (messages).

### DIFF
--- a/runtime/errors/errors.go
+++ b/runtime/errors/errors.go
@@ -1,0 +1,141 @@
+/*
+Copyright 2020 The Flux CD contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


runtime/errors/errors.go
```diff
@@ -0,0 +1,141 @@
...
+// it includes the Kind and NamespacedName of the resource the reconciliation
+// was performed for, and the underlying Err.
+type ReconciliationError struct {
+	Kind           string
+	NamespacedName types.NamespacedName
+	Err            error
+}
+
+func (e *ReconciliationError) Error() string {
+	return fmt.Sprintf("%s '%s' reconciliation failed: %v", e.Kind, e.NamespacedName.String(), e.Err)
```
stefanprodan, line 35: Wouldn't this duplicate the info in logs and events? We log and emit events with Kind, name and namespace as labels.

hiddeco: Possibly, but otherwise the error would lose context which from a "idiomatic Go" point of view is not right.

hiddeco: There is another reason you may want to include this by the way, the helm-controller for example reconciles _both_ `HelmChart` and `HelmRelease` objects, but the event always attaches metadata for the `HelmRelease`. Without this attached to the error the error message would provide little context over what failed with a notification and/or event context.

stefanprodan: Good point!

stefanprodan (approved):

PR #34, opened 2020-10-14
Add unit suffix to reconcile duration metric

seaneagan:
In order to align with the [prometheus metric naming guidelines](https://prometheus.io/docs/practices/naming/) which say
that metric names:

> should have a suffix describing the unit, in plural form.

ref: https://github.com/fluxcd/toolkit/issues/329

Note: a follow up would be needed to update the dashboard added in https://github.com/fluxcd/toolkit/pull/337


hiddeco (approved):
Strong argument / reference in favour of this rename :+1:

stefanprodan (approved):
👍

PR #35, opened 2020-10-19
Introduce `version` package

hiddeco:
Ref: https://github.com/fluxcd/source-controller/issues/127

### DIFF
--- a/version/version_test.go
+++ b/version/version_test.go
@@ -1,0 +1,49 @@
+/*
+Copyright 2020 The Flux CD contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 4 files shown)


version/version_test.go
```diff
@@ -0,0 +1,49 @@
...
+		{"v1.0", true},
+		{"v1", true},
+		{"v1.2.beta", true},
+		{"v1.2-5", true},
+		{"v1.2-beta.5", true},
+		{"\nv1.2", true},
+		{"v1.2.0-x.Y.0+metadata", false},
+		{"v1.2.0-x.Y.0+metadata-width-hypen", false},
+		{"v1.2.3-rc1-with-hypen", false},
+		{"v1.2.3.4", true},
```
stefanprodan, line 38: Can you please add `1.0.0` and `100`

hiddeco: Then we are duplicating tests from the package we import.

stefanprodan (approved):

PR #36, opened 2020-10-19
Document the contract of reconcileAt

squaremo:


hiddeco (approved):
Thanks Michael :coffee: 

In addition to this, we should probably spark a new discussion about the rename, as I think your comments made sense

PR #37, opened 2020-10-20
Reconcile request annotation and helpers

squaremo:
This adds a new annotation name, as proposed in https://github.com/fluxcd/toolkit/discussions/363, and provides some helpers for

 - getting the value in a backward-compatible way
 - recording the value in a CRD status field (so as to detect changes).

### DIFF
--- a/apis/meta/annotations.go
+++ b/apis/meta/annotations.go
@@ -47,3 +47,27 @@ func ReconcileAnnotationValue(annotations map[string]string) (string, bool) {
 	// used each time, this caveat won't matter.
 	return reconcileAt + requestedAt, ok1 || ok2
 }
+
+// ReconcileRequestStatus is a struct to embed in the status type, so
+// that all types using the mechanism have the same field. Use it like
+// this:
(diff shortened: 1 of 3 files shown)


apis/meta/annotations.go
```diff
@@ -22,4 +22,52 @@ const (
 	// interpreted as a timestamp, and any change in value shall trigger a
 	// reconciliation.
```
hiddeco, line 23:
```suggestion
	// reconciliation.
	// DEPRECATED: has been replaced by ReconcileAtAnnotation.
```

This makes GoLand (and other "intelligent" IDEs) show it with strikethrough markup.

squaremo:
I tried this out in image-reflector-controller by putting a `replace` directive in my go.mod, and it is a modest 1-for-1 change in lines of code.

 - [ ] I should update the predicate in this repo so it uses `ReconcileAnnotationValue()`

hiddeco (approved):
Thank you Michael :cherries:

PR #41, opened 2020-10-31
Update k8s to 1.19 in apis/meta

relu:


stefanprodan:
This needs to be done in 2 PRs: first `apis/meta`, followed by semver release of `api/meta`, then bump meta and k8s in `runtime`,  followed by semver release of `runtime`.

relu:
Actually, I just realized this after removing the second commit for the `runtime` mod update, because of the `replace` at the top of the mod file, you can't really do this in two PRs since it's always going to resolve `apis/meta` to the relative path, no matter what version constraint is set in the `require` section.

stefanprodan (approved):

stefanprodan:
@relu I've released apis/meta/v0.2.0 please go ahead with the runtime bump. Thanks!

PR #44, opened 2020-11-01
Adopt k8s standardized conditions in apis/meta

relu:
Use the newly introduced standardized Condition type kubernetes/enhancements#1624

Relates to fluxcd/flux2#225


apis/meta/conditions.go
```diff
@@ -75,48 +49,16 @@ const (
...
-		return nil
-	}
-	out := new(Condition)
-	in.DeepCopyInto(out)
-	return out
-}
-
-// GetCondition returns the Condition from the given slice that matches the
-// given type.
-func GetCondition(conditions []Condition, conditionType string) *Condition {
```
relu, line 96: Uncertain if to keep this and refactor so it's just a wrapper for [FindStatusCondition](https://github.com/kubernetes/apimachinery/blob/v0.19.3/pkg/api/meta/conditions.go#L73) or just ditch it (as is done now) and switch to that one directly where applicable.

stefanprodan: I think this is fine, we should be using FindStatusCondition.

apis/meta/conditions.go
```diff
@@ -75,48 +49,16 @@ const (
...
-
 // HasReadyCondition returns if the given Condition slice has a ReadyCondition
 // with a 'True' condition status.
-func HasReadyCondition(conditions []Condition) bool {
-	condition := GetCondition(conditions, ReadyCondition)
-	if condition == nil {
-		return false
-	}
-	return condition.Status == corev1.ConditionTrue
+func HasReadyCondition(conditions []metav1.Condition) bool {
```
hiddeco, line 54:
```suggestion
func InReadyCondition(conditions []metav1.Condition) bool {
```

The `func` name at present was suggested when I created the PR, but I did not take into account when I accepted the suggestion that it also checks if the `Ready==True`.

relu:
Question: once this is out, will consumer API packages need a version bump (e.g. from v1alpha1 to v1alpha2)? My hunch is that it's not the case considering the interface is preserved (aside from the addition of the [ObservedGeneration](https://github.com/kubernetes/apimachinery/blob/v0.19.3/pkg/apis/meta/v1/types.go#L1388) and that should ensure backward compatibility I assume.

stefanprodan:
@relu this is not a break change so I would keep all the APIs versions as they are.

stefanprodan (approved):
LGTM

Thanks @relu 🏅

hiddeco (approved):
Next package release should be a MINOR update given the breaking changes to the API.

PR #46, opened 2020-11-04
Replace reference to corev1.ConditionUnknown

relu:
Looks like I forgot to use metav1.ConditionUnknown instead in the `RecordCondition`. No biggie since it works just fine considering that the stringified value is used.


hiddeco (approved):
:fishing_pole_and_fish:

PR #47, opened 2020-11-04
Remove FilterOutCondition

relu:
Removing the FilterOutCondition func in favour of k8s meta.RemoveStatusCondition func which implements the same logic.


hiddeco (approved):
LGTM, thank you @relu :100:

PR #48, opened 2020-11-05
Add SetResourceCondition func to apis/meta

relu:
Handle setting `Status.Conditions` for any resource implementing the `v1.Object` interface. This is common in all flux api types.

### DIFF
--- a/apis/meta/conditions.go
+++ b/apis/meta/conditions.go
@@ -67,3 +67,30 @@ func FilterOutCondition(conditions []metav1.Condition, conditionType string) []m
 	}
 	return newConditions
 }
+
+// SetResourceCondition sets the given condition with the given status,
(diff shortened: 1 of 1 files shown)


seaneagan:
@relu what about using an interface instead and implementing it in each controller?

```go
type ObjectWithStatusConditions interface {
  v1.Object
  GetStatusConditions() *[]metav1.Condition
} 
```

I guess struct embedding won't work here: https://github.com/kubernetes-sigs/controller-tools/issues/244

relu:
I was reluctant to do that thinking it wouldn't be a good idea to require adding a getter method to all API types just for this purpose.
It's probably worth it though, considering the repetition of the `Set*Condition` logic we have now for each type.

PR #50, opened 2020-11-17
Remove InReadyCondition from apis/meta

relu:
Useless func now that we have the standardized Condition type and
helpers from k8s.io/apimachinery/pkg/api/meta.


hiddeco (approved):

hiddeco:
@relu oops, forgot to merge this when things had settled in the implementing controller repositories. Can you please rebase?

PR #55, opened 2020-11-25
Correct spelling in string literals

squaremo:
(and use en-US spelling)


squaremo:
Will rebass and sign-off ..

stefanprodan (approved):

PR #56, opened 2020-11-26
Controller helpers

squaremo:
This is an experimental PR to see if some of the bits common to all the GOTK controllers can be factored out into embeddable structs.

Among other things, this will

 - reduce duplication
 - standardise how these common bits are done

It is a bit rough around the edges:

**You have to supply an ObjectReference to both the metrics and the events methods**. This is because all the constituent bits calculated this for themselves, and had different ways of dealing with errors. This way, the error handling is done once, up front, and the ref only has to be calculated once. But it would be nice to do this behind the scenes, in some way.

**You have to pass logs in to the Events helper**. This is a bit awkward, as getting logs to funcs always is. It'd be nice if the setup could just do this for you.

### DIFF
--- a/runtime/controller/events.go
+++ b/runtime/controller/events.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2020 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	kuberecorder "k8s.io/client-go/tools/record"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/fluxcd/pkg/runtime/events"
+)
+
+// Events is a helper struct that adds the capability of sending
+// events to the Kubernetes API and to the GitOps Toolkit notification
+// controller. You use it by embedding it in your reconciler struct:
+//
+//     type MyTypeReconciler {
+//         client.Client
+//         // ... etc.
+//         controller.Events
+//     }
+//
+//  You initialise a suitable value with MakeEvents(); each reconciler
+//  will probably need its own value, since it's specialised to a
+//  particular controller and log.
+type Events struct {
+	EventRecorder         kuberecorder.EventRecorder
+	ExternalEventRecorder *events.Recorder
+	Log                   logr.Logger
+}
+
+func MakeEvents(mgr ctrl.Manager, controllerName string, ext *events.Recorder, log logr.Logger) Events {
+	return Events{
+		EventRecorder:         mgr.GetEventRecorderFor(controllerName),
+		ExternalEventRecorder: ext,
+		Log:                   log,
+	}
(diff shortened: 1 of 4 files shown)


runtime/controller/events.go
```diff
@@ -0,0 +1,79 @@
...
+type runtimeAndMetaObject interface {
+	runtime.Object
+	metav1.Object
+}
+
+// Event emits a Kubernetes event, and forwards the event to the
+// notification controller if configured.
+func (e Events) Event(ref *corev1.ObjectReference, obj runtimeAndMetaObject, severity, msg string) {
+	if e.EventRecorder != nil {
+		e.EventRecorder.Event(obj, "Normal", severity, msg)
```
hiddeco, line 68: Instead of always using `Normal` here, I think it would be better if we serve an `EventTypeWarning` or `EventTypeNormal` from the Kubernetes core package

This also means that the `severity` can become a proper `Reason` string (as we can now determine the severity for the external events based on those two event types), so that the formatting of the events we emit is aligned with what other Kubernetes controllers do.

runtime/controller/metrics.go
```diff
@@ -0,0 +1,81 @@
...
+func (m Metrics) RecordDuration(ref *corev1.ObjectReference, startTime time.Time) {
+	if m.MetricsRecorder != nil {
+		m.MetricsRecorder.RecordDuration(*ref, startTime)
+	}
+}
+
+type readinessMetricsable interface {
+	metav1.Object
+	meta.ObjectWithStatusConditions
+}
```
seaneagan, line 67: Why not just add back `metav1.Object` directly to `meta.ObjectWithStatusConditions` since it [will need it soon anyways](https://github.com/fluxcd/pkg/pull/49/commits/0021192177fb762664896931a8d46a9e424b7f74#diff-43341f9af32900637b6cc623ca582855f331dfab462f34cf8868ccd1a706e5c2L61)?

runtime/controller/metrics.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (m Metrics) RecordReadinessMetric(ref *corev1.ObjectReference, obj readinessMetricsable) {
```
seaneagan, line 69: What about calling this `RecordConditionMetrics` or similar so that if we decide to expose metrics for other conditions besides `Ready` (e.g. all of them) in the future, we don't need a breaking method name change to do so?

runtime/controller/metrics.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	if rc := apimeta.FindStatusCondition(*obj.GetStatusConditions(), meta.ReadyCondition); rc != nil {
+		m.MetricsRecorder.RecordCondition(*ref, *rc, !obj.GetDeletionTimestamp().IsZero())
+	} else {
+		m.MetricsRecorder.RecordCondition(*ref, metav1.Condition{
+			Type:   meta.ReadyCondition,
+			Status: metav1.ConditionUnknown,
+		}, !obj.GetDeletionTimestamp().IsZero())
+	}
+}
```
seaneagan, line 81:
```suggestion
	rc := apimeta.FindStatusCondition(*obj.GetStatusConditions(), meta.ReadyCondition)
	if rc == nil {
		rc = metav1.Condition{
			Type:   meta.ReadyCondition,
			Status: metav1.ConditionUnknown,
		}
	}
	m.MetricsRecorder.RecordCondition(*ref, *rc, !obj.GetDeletionTimestamp().IsZero())
}
```

runtime/controller/metrics.go
seaneagan: Additionally, the only time we should ever need to record the metric I believe is when its value changes, so what about adding a method to this type which both sets the status condition and records the metric, either by using `meta.SetResourceCondition` or replacing it?

blaquezzz (approved):

squaremo:
@hiddeco, @seaneagan: is this useful enough to be worth persevering, wdyt?

hiddeco:
I think that with our comments addressed it would be really useful to get good Kubernetes events and metadata rich notifications. So, yes!

stefanprodan (approved):
LGTM 🚀

squaremo (commented):
Hidde, your commit messages are a joy 🥭

PR #58, opened 2020-12-01
Add gotk_suspend_status metric

sylr:
This metric aims to add the possibility to alert when resources are suspended for too long.


relu (approved):
LGTM :100:

relu:
@stefanprodan while looking at this, I noticed that the metrics exported here are still namespace with `gotk_`. Not sure if this was missed during the renaming refactor or if it was intentionally left out, maybe you can clarify.

stefanprodan (approved):
LGTM

We need to create an issue in flux2 for implementing this in all controllers

PR #59, opened 2020-12-02
Add ability to start git server with TLS

phillebaba:
Add option to run git server with a tls cert.

### DIFF
--- a/gittestserver/server.go
+++ b/gittestserver/server.go
@@ -76,7 +76,7 @@ func (s *GitServer) StartHTTP() error {
 }
 
 // StartHTTPTLS starts the TLS HTTPServer with the given TLS configuration.
-func (s *GitServer) StartHTTPTLS(cert, key, ca []byte, serverName string) error {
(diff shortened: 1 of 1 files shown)


gittestserver/server.go
```diff
@@ -73,6 +75,34 @@ func (s *GitServer) StartHTTP() error {
 	return nil
 }
 
+// StartHTTPTLS starts the TLS HTTPServer with the given TLS configuration.
+func (s *GitServer) StartHTTPTLS(cert, key, ca []byte, serverName string) error {
```
hiddeco, line 79:
```suggestion
func (s *GitServer) StartHTTPS(cert, key, ca []byte, serverName string) error {
```

or

```suggestion
func (s *GitServer) StartHTTPWithTLS(cert, key, ca []byte, serverName string) error {
```

maybe?

phillebaba: Makes fore sense going with option 1

hiddeco (approved):
LGTM, thanks Philip :100:

PR #62, opened 2020-12-09
Fix CreateRepository for GitLab provider

### LINKED ISSUE #22: Adding Deploy Key for GitLab tries to use the wrong project
The list projects method actually does a search and naïvely grabs the first one it finds. e.g. if you have a repo called manifests it will execute a search like this https://gitlab.com/search?search=manifests and grab https://gitlab.com/evolution-x-sakura/local_manifests_sakura as the project to add the key to. This will fail for obvious reasons.

relu:
Ensure projects are discovered using the correct method in the case of group owned projects (ListGroupProjects) vs personal owned projects (ListProjects).

The issue can be reproduced when bootstrapping on GitLab with a personal account and using a repository name that could already belong to another user.

This is from a user report in [slack](https://cloud-native.slack.com/archives/CLAJ40HV3/p1607540221358200).

Fixes #22.

### DIFF
--- a/git/provider_gitlab.go
+++ b/git/provider_gitlab.go
@@ -56,48 +56,18 @@ func (p *GitLabProvider) CreateRepository(ctx context.Context, r *Repository) (b
 		return false, fmt.Errorf("client error: %w", err)
 	}
 
-	var gid *int
-	var projects []*gitlab.Project
-	if !p.IsPersonal {
-		lgo := &gitlab.ListGroupsOptions{
-			Search:         gitlab.String(r.Owner),
-			MinAccessLevel: gitlab.AccessLevel(gitlab.GuestPermissions),
-		}
-		groups, _, err := gl.Groups.ListGroups(lgo, gitlab.WithContext(ctx))
-		if err != nil {
-			return false, fmt.Errorf("failed to list groups, error: %w", err)
-		}
+	gid, projects, err := p.getProjects(ctx, gl, r)
+	if err != nil {
+		return false, fmt.Errorf("failed to list projects, error: %w", err)
+	}
 
-		if len(groups) == 0 {
-			return false, fmt.Errorf("failed to find group named '%s'", r.Owner)
-		}
-		gid = &groups[0].ID
-
-		lpo := &gitlab.ListGroupProjectsOptions{
-			Search: gitlab.String(r.Name),
-		}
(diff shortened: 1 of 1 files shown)


relu:
I've tested this locally with `flux2` with the following use cases:

1. Is able to create personal repo: `flux bootstrap gitlab --owner=personalacc --repository=flux --personal --token-auth --branch=master`
2. Is able to discover the previously created repo (command re-execution).
3. Is able to create group owned repo: `flux bootstrap gitlab --owner=groupacc --repository=flux --token-auth --branch=master`
4. Is able to discover the previously created repo (command re-execution).

git/provider_gitlab.go
```diff
@@ -56,15 +56,37 @@ func (p *GitLabProvider) CreateRepository(ctx context.Context, r *Repository) (b
...
 
 		if len(groups) > 0 {
-			id = &groups[0].ID
+			gid = &groups[0].ID
+		}
+
+		lpo := &gitlab.ListGroupProjectsOptions{
+			Search: gitlab.String(r.Name),
+		}
+		projects, _, err = gl.Groups.ListGroupProjects(*gid, lpo, gitlab.WithContext(ctx))
```
relu, line 78: Note: This method will retrieve only repos in the context of `gid`.

git/provider_gitlab.go
```diff
@@ -56,15 +56,37 @@ func (p *GitLabProvider) CreateRepository(ctx context.Context, r *Repository) (b
...
+		if err != nil {
+			return false, fmt.Errorf("failed to list projects, error: %w", err)
+		}
+	} else {
+		lpo := &gitlab.ListProjectsOptions{
+			Search: gitlab.String(r.Name),
+			Owned:  gitlab.Bool(true),
```
relu, line 85: Note: This ensures the listing includes only repositories owned by the current user. Lacking this was the cause of the original user-reported issue.

stefanprodan:
This fixes #22?

relu:
No, but I can probably fix that in the scope of this PR as well. Wasn't aware of the issue, it's definitely similar in nature.

relu:
I've tested deploy key creation in the scenarios previously mentioned and I can confirm it works with these changes now.

stefanprodan (approved):
LGTM

Thanks @relu 🥇

PR #67, opened 2021-01-13
Ensure correct group is gotten from gitlab search

SomtochiAma:
This PR 
1. ensures that the group and subgroup gotten from the GitLab search is the same specified by the user
2. sets the correct gitlab path in the `getProjects` function and removes `getRepositoryOwner` from the interface.

I tested these changes on a self-hosted GitLab instance and I confirm that I fixed the regression bug introduced in [flux#681](https://github.com/fluxcd/flux2/pull/681) that prevented users from using a custom hostname instead of gitlab.com

### DIFF
--- a/git/provider_gitlab.go
+++ b/git/provider_gitlab.go
@@ -84,31 +84,6 @@ func (p *GitLabProvider) CreateRepository(ctx context.Context, r *Repository) (b
 	return true, nil
 }
 
-// GetRepositoryOwner returns the actual path owner. This is need for Gitlab where the name of a group might differ
-// from its path
-func (p *GitLabProvider) GetRepositoryOwner(ctx context.Context, token string, owner string) (string, error) {
-	gl, err := gitlab.NewClient(token)
-	if err != nil {
-		return "", fmt.Errorf("client error: %w", err)
-	}
-
-	groupName := strings.Split(owner, "/")[0]
-	lgo := &gitlab.ListGroupsOptions{
-		Search:         gitlab.String(groupName),
-		MinAccessLevel: gitlab.AccessLevel(gitlab.GuestPermissions),
-	}
-	groups, _, err := gl.Groups.ListGroups(lgo, gitlab.WithContext(ctx))
-	if err != nil {
-		return "", fmt.Errorf("failed to list groups, error: %w", err)
-	}
-
-	if len(groups) == 0 {
-		return "", fmt.Errorf("failed to find group named '%s'", groupName)
-	}
-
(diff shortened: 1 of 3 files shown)


git/provider_gitlab.go
```diff
@@ -255,3 +258,13 @@ func (p *GitLabProvider) getProjects(ctx context.Context, gl *gitlab.Client, r *
 
 	return gid, projects, nil
 }
+
+func getRightGroup(groups []*gitlab.Group, name string)  *gitlab.Group{
```
hiddeco, line 262:
```suggestion
func findGroupByName(groups []*gitlab.Group, name string)  *gitlab.Group{
```

git/provider_gitlab.go
```diff
@@ -255,3 +258,13 @@ func (p *GitLabProvider) getProjects(ctx context.Context, gl *gitlab.Client, r *
...
+	for _, group := range groups {
+		if group.Name == name {
+			return group
+		}
+	}
+
+	return nil
+}
```
hiddeco, line 270:
```suggestion
}

```

hiddeco:
Other (unsolicited) advice: instead of adding `[WIP]` to the PR title, GitHub has a [draft PR feature](https://github.blog/2019-02-14-introducing-draft-pull-requests/) that reflects the same.

For PRs that already bear the "normal" state, the sidebar offers a

> Still in progress? Convert to draft

link that converts it to a draft.

stefanprodan (approved):
LGTM

Please rebase

git/provider_gitlab.go
```diff
@@ -84,31 +84,6 @@ func (p *GitLabProvider) CreateRepository(ctx context.Context, r *Repository) (b
...
-	if err != nil {
-		return "", fmt.Errorf("failed to list groups, error: %w", err)
-	}
-
-	if len(groups) == 0 {
-		return "", fmt.Errorf("failed to find group named '%s'", groupName)
-	}
-
-	return groups[0].Path, nil
-}
```
hiddeco, line 110: As this interface is public and we want to be nice to our package consumers (even if that's just us), I think it would be good practice to keep this for the time being, and add a `Deprecated` note to the doc block like this:

```
// Deprecated, has become obsolete due to changes in AddTeam.
// 
// GetRepositoryOwner returns the actual path owner. This is need for Gitlab where the name of a group might differ
// from its path
```

(or something that's more descriptive).

hiddeco (approved):
Super b! Thank you @SomtochiAma :100:

PR #68, opened 2021-01-13
Fix false negatives from reconcile-at predicate

squaremo:
The logic of the (newly separated) ReconcileAtChanged predicate should
be:

    if new object has reconcileAt annotation
      if old obj has different or no annotation -> true
    else -> false

However the old and new object were swapped around, leading to

    if old object has reconcileAt annotation
    ...

which results in `false` when the annotation has just been applied,
when it should be `true`.

Signed-off-by: Michael Bridgen <michael@weave.works>


hiddeco (approved):
100% my mistake, and a great catch :fishing_pole_and_fish: 

Thank you Michael :broccoli:

PR #69, opened 2021-01-13
Rename predicate to ReconcileRequestedPredicate

hiddeco:
This changes the name of the predicate (that contained a typo) from
`ReconcilateAtChangedPredicate` to `ReconcileRequestedPredicate`,
which does _not_ contain a typo and is more in-line with the naming
of the `ReconcileRequestAnnotation`.

### DIFF
--- a/runtime/predicates/reconcile_requested.go
+++ b/runtime/predicates/reconcile_requested.go
@@ -1,0 +1,56 @@
+/*
+Copyright 2021 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


relu (approved):
At least you invented a new word, good catch! :100:

runtime/predicates/reconcile_requested.go
```diff
@@ -0,0 +1,56 @@
+/*
+Copyright 2020 The Flux authors
```
relu, line 2: Maybe also 2021?

PR #72, opened 2021-01-18
GitLab: Check both the name and the path when searching for groups

SomtochiAma:
This pull-request checks for the name or the path when finding the group from the search returned by GitLab.
Currently, users of flux v0.6.1, due to this pull request #67, get a `failed to find group` error because we check only the name. In v0.6.next, they will be able to.
Issue ref: [flux2#727](https://github.com/fluxcd/flux2/issues/727)


stefanprodan:
Can you please describe what does this do? Is there an issue behind the change? How does it affect current users? What about users of flux2 v0.6 vs v0.5? Does it affect upgrades from v0.6.1 to v0.6.next?

hiddeco (approved):

stefanprodan (approved):

PR #76, opened 2021-01-21
Drop required access level when listing groups using the GitLab API

### LINKED ISSUE #75: Support using GitLab project tokens
When bootstrapping flux with GitLab, it needs to perform various actions on the GitLab instance:
* Creating a repository (if it doesn't exist, yet)
* Adding a deployment key

During these operations, it will try to find the correct GitLab project using the GitLab API. With the current implementation, the search is restricted to return only groups with an access level >= 10.

When using a project access token in gitlab (from a repository that has already been manually created), the gitlab API will not list project parent groups if search is performed with access level >= 10. This will cause the bootstrapping to fail:

```
ubuntu@myhost:~$ flux bootstrap gitlab \
>     --owner=mygroup/mysubgroup \
>     --repository=myrepo

✗ failed to list projects, error: failed to find group named 'mygroup/mysubgroup'
```

The relevant code sections are:
* https://github.com/fluxcd/pkg/blob/main/git/provider_gitlab.go#L60
* https://github.com/fluxcd/pkg/blob/main/git/provider_gitlab.go#L197

**Solution**
The requirement on the access level can simply be omitted. The GitLab API will only list projects/groups the provided token can see.

defreng:
Resolves #75 

I did run a small test vs our local gitlab instance without any issues:
image

### DIFF
--- a/git/provider_gitlab.go
+++ b/git/provider_gitlab.go
@@ -193,8 +193,7 @@ func (p *GitLabProvider) getProjects(ctx context.Context, gl *gitlab.Client, r *
 	if !p.IsPersonal {
 		groupAndSubGroups := strings.Split(r.Owner, "/")
 		lgo := &gitlab.ListGroupsOptions{
-			Search:         gitlab.String(groupAndSubGroups[0]),
-			MinAccessLevel: gitlab.AccessLevel(gitlab.GuestPermissions),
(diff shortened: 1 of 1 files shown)


SomtochiAma:
This pull request also works with personal access token

stefanprodan (approved):
LGTM

Thanks @defreng for the fix and thanks @SomtochiAma for testing it

PR #77, opened 2021-02-03
Introduce Kustomize API group

hiddeco:
This PR introduces a specific set of structures to work with
Kustomize patches without depending directly on the Kustomize API,
and provide simple validation for the JSON6902 operations.


stefanprodan (approved):
LGTM

Thanks @hiddeco 🍻

PR #80, opened 2021-02-04
git: drop usage of Name for GitLab operations

hiddeco:
During the development of the provider a mistake was made, and the
repository was created using the `Name` field instead of the `Path`.
The result of this is that GitLab will create a `Path` based on the
provided `Name`, with a random (number) suffix added, resulting in
all sorts of issues (and work around attempts) down the line.

By only working with `Path` fields, and assuming that the provided
repository name is a path, the look ups we make are now guaranteed
to always return the same group, subgroup or repository.

### DIFF
--- a/git/provider_gitlab.go
+++ b/git/provider_gitlab.go
@@ -72,7 +72,7 @@ func (p *GitLabProvider) CreateRepository(ctx context.Context, r *Repository) (b
 	}
 
 	cpo := &gitlab.CreateProjectOptions{
-		Name:                 gitlab.String(r.Name),
+		Path:                 gitlab.String(r.Name),
 		NamespaceID:          gid,
 		Visibility:           &visibility,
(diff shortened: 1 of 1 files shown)


git/provider_gitlab.go
```diff
@@ -191,46 +164,35 @@ func (p *GitLabProvider) getProjects(ctx context.Context, gl *gitlab.Client, r *
 		err      error
 	)
 	if !p.IsPersonal {
```
stefanprodan, line 166: Can you please create two function `getPersonalProjects` and `getGroupProjects`

hiddeco: I split them into `getPersonalProject` and `getGroupProject`, as the return of the list when we are looking for a single item made little sense to me.

stefanprodan (approved):
LGTM 

Thanks @hiddeco 🥇

PR #82, opened 2021-02-08
Make SSH server usable for testing

squaremo:
The underlying library, gitkit, expects to have a KeyDir supplied in its *Config when running an SSH server. It's uncertain whether StartSSH is used anywhere (and it wouldn't work, if it is); but for the least disruption, I have added a func for supplying the KeyDir value rather than changing the signature of any existing func.

Since StartSSH blocks when it succeeds, the test has to do a kludge with a timer to be able to exit.

`StartSSH()` doesn't return until `StopSSH()` is called (or there's an error). This makes it difficult to use in tests, since you will usually want to know the address to use before proceeding with the rest of the test, but the listener must be created before it can be obtained.

This PR adds a `ListenSSH()`, which creates the SSH server and listener on localhost, then returns. That means you can call `ListenSSH()`, get the address, start a goroutine to run StartSSH(), then complete the test (and `StopSSH()`).

The semantics of how to use SSH server is different now: instead of StartSSH creating a new server (with a new listener), both ListenSSH() and StartSSH() create a server _if it hasn't been created already_. This is necessary because creating a new listener each time invalidates any address obtained from `SSHAddress()` previously, making the test control flow above impossible.

Another change, for testing convenience: `SSHAddress()` now returns a URL you can use with git, like `HTTPAddress()` does.

### DIFF
--- a/gittestserver/server.go
+++ b/gittestserver/server.go
@@ -84,7 +84,7 @@ func (s *GitServer) StartHTTP() error {
 	return nil
 }
 
-// StartHTTPTLS starts the TLS HTTPServer with the given TLS configuration.
+// StartHTTPS starts the TLS HTTPServer with the given TLS configuration.
 func (s *GitServer) StartHTTPS(cert, key, ca []byte, serverName string) error {
 	s.StopHTTP()
 	service := gitkit.New(s.config)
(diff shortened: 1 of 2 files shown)


hiddeco (approved):
Looks like the yak has been shorn, thank you @squaremo :pineapple:

squaremo:
I need to do a bit more here, because you can't get the address without calling at least `s.sshServer.Listen(...)`, and `s.sshServer.ListenAndServe(...)` won't return; so Listen needs to be explicitly invokable.

hiddeco (approved):

PR #83, opened 2021-02-11
Add method to GitServer to switch auth on

squaremo:
When testing using gittestserver, it's useful to be able to use
authentication. This is especially true when testing against libgit2,
because it expects authentication over SSH, and will balk if it's not
requested.

This commit adds a method to *GitServer for switching auth on. There's
no way to switch auth on for just SSH -- it will also be required for
HTTP. For that reason, a username and password are needed.

The credentials are included in the URL returned by HTTPAddress(), so
anything using that will transparently work whether auth is on or
not. An alternative would be to let uses of *GitServer provide the
credentials if auth is on; but, it is not always convenient to supply
creds for git over HTTP, so those uses would likely have to all just
recreate the code that puts them in the URL anyway.

### DIFF
--- a/gittestserver/server.go
+++ b/gittestserver/server.go
@@ -21,6 +21,7 @@ import (
 	"crypto/x509"
 	"io/ioutil"
 	"net/http/httptest"
+	"net/url"
 	"path/filepath"
 
 	"github.com/sosedoff/gitkit"
@@ -57,6 +58,8 @@ type GitServer struct {
 	config     gitkit.Config
 	httpServer *httptest.Server
 	sshServer  *gitkit.SSH
+	// Set these to configure HTTP auth
+	username, password string
 }
 
 // AutoCreate enables the automatic creation of a non-existing Git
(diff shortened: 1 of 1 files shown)


gittestserver/server.go
```diff
@@ -164,10 +182,21 @@ func (s *GitServer) Root() string {
...
+		u, err := url.Parse(s.httpServer.URL)
+		if err != nil {
+			panic(err)
+		}
+		if s.password != "" {
+			u.User = url.UserPassword(s.username, s.password)
+		} else if s.username != "" {
+			u.User = url.User(s.username)
+		}
+		return u.String()
```
hiddeco, line 199: Doesn't this complicate situations where you want to e.g. mock a `GitRepository` resource including the `Secret` required to authenticate, and do not want to include the credential part of the URL in the URL you define (as those are defined in the secret)?

If so, maybe it would be wise to add a `RepositoryHTTPAddress(repository string)` (and maybe also the `SSH` version) that does not include any credentials for HTTP and only the user for SSH?

squaremo: Yeahp, that's the other branch of that decision. Good suggestion -- in fact perhaps I should have `HTTPAddressWithCredentials()` and adjust _my_ tests to use that.

squaremo: Let me rework this a little.

hiddeco (approved):
Thank you @squaremo 🍎

PR #88, opened 2021-03-01
Transpose git error tidiers from image-automation

squaremo:
This adds the git error tidiers from

  https://github.com/fluxcd/image-automation-controller/pull/115

to fluxcd/pkg/gitutil, so they can be used elsewhere e.g., in
source-controller.


gitutil/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/fluxcd/pkg/gitutil
+
+go 1.14
```
stefanprodan, line 3:
```suggestion
go 1.15
```

We use 1.15 in git, runtime, meta, etc

squaremo: Ah yes, it picked up my local go version which I haven't bothered to update :-S

stefanprodan (approved):
LGTM

PR #91, opened 2021-03-19
Generate deepcopy method for event

SomtochiAma:


hiddeco (approved):
Thank you @SomtochiAma :sun_with_face:

hiddeco (commented):
The instruction I originally gave you are a bit awkward for this package. To resolve the issue that the CI is giving you, the best option is probably to delete the `runtime/events/doc.go` file, and add the following above the `Event` structure:

```go
// Event is a report of an event issued by a controller.
// +kubebuilder:object:generate=true
type Event struct {
	// The object that this event is about.
```

This should result in the `DeepCopy` methods only being generated for that struct, while others are left alone.

stefanprodan (approved):
LGTM

Please squash the commits, thanks @SomtochiAma

hiddeco (approved):
🌈

PR #92, opened 2021-03-25
Add leader election config to runtime

stefanprodan:
This PR adds cmd flags for fine tuning the controller-runtime leader election. To make the leader election more tolerant to transient Kubernetes API outages (rate limits, CNI upgrade, control plane upgrades, etc) the renewal deadline was increased from 10 seconds to one minute and the retry interval was bumped to 5 seconds.

### DIFF
--- a/runtime/leaderelection/leaderelection.go
+++ b/runtime/leaderelection/leaderelection.go
@@ -1,0 +1,73 @@
+/*
+Copyright 2021 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


runtime/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,73 @@
...
+	// ReleaseOnCancel defines if the leader should step down voluntarily
+	// when the Manager ends. This requires the binary to immediately end when the
+	// Manager is stopped, otherwise this setting is unsafe. Setting this significantly
+	// speeds up voluntary leader transitions as the new leader doesn't have to wait
+	// LeaseDuration time first.
+	ReleaseOnCancel bool
+
+	// LeaseDuration is the duration that non-leader candidates will
+	// wait to force acquire leadership. This is measured against time of
+	// last observed ack. Default is 15 seconds.
```
hiddeco, line 48: Given the context of the doc block is this package, I think it should be:

```suggestion
	// last observed ack. Default is 30 seconds.
```

Same goes for the other documented defaults.

hiddeco (approved):
Fixed with the speed of light, thanks @stefanprodan :1st_place_medal:

PR #102, opened 2021-05-18
Add TLS package and helper methods to runtime

phillebaba:
This change is meant to standardize how certificates are read from secrets by the controllers. This logic is currently shared by notification-controller and image-reflector-controller, there might be more.

The original logic is taken from image-reflector-controller.
https://github.com/fluxcd/image-reflector-controller/blob/de3de64adddf8bb00276e12aa0436968bc94ff39/controllers/imagerepository_controller.go#L278-L309

I did however make a change to the behavior so that the functions will return an error if the secret does not contain a ca or cert. Or when the client cert is set but not client key and vice versa. My opinion is that this behavior is more logical than the previous one.

### DIFF
--- a/runtime/tls/doc.go
+++ b/runtime/tls/doc.go
@@ -1,0 +1,18 @@
+/*
+Copyright 2021 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package tls contains helpers to convert Kubernetes secrets to TLS certificates.
+package tls
--- a/runtime/tls/testdata/keys.go
+++ b/runtime/tls/testdata/keys.go
@@ -1,0 +1,51 @@
+/*
+Copyright 2021 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package testdata
+
+var ExampleCA = []byte(`-----BEGIN CERTIFICATE-----
+MIIB7TCCAZKgAwIBAgIUB+17B8PU05wVTzRHLeG+S+ybZK4wCgYIKoZIzj0EAwIw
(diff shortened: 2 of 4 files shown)


phillebaba:
@squaremo do you have any opinion about the change in behavior compared to image-reflector-controller?

squaremo:
> @squaremo do you have any opinion about the change in behavior compared to image-reflector-controller?

Seems reasonable to me. I can't think of a situation in which you'd want a client cert without the key, or vice versa.

runtime/cert/cert.go
```diff
@@ -0,0 +1,86 @@
...
+// If the secret does not contain either a client cert and key or ca cert an
+// error will be returned. If the secret contains only the cert or key an error
+// will be returned.
+func TlsConfigFromSecret(certSecret *corev1.Secret) (*tls.Config, error) {
+	validSecret := false
+	tlsConfig := &tls.Config{}
+
+	clientCert, clientCertOk := certSecret.Data[ClientCert]
+	clientKey, clientKeyOk := certSecret.Data[ClientKey]
+	if clientCertOk && !clientKeyOk {
```
squaremo, line 44: I would usually put mutually exclusive conditions using the same variables like these in a `switch`, because it looks a bit like a logic table:

```
switch {
  case clientCertOk && !clientKeyOk:
    // ...
  case !clientCertOk && clientKeyOk:
    // ...
  case clientCertOk && clientKeyOk:
    // ...
}
```

Modulo the error message, all you care about is that they are both present or neither, so you could also do this:

```
if clientKeyOk != clientCertOk {
  return nil, fmt.Errorf("found one of %s or %s, and expected both or neither", ClientCert, ClientKey)
}
if clientKeyOk && clientCertOk {
  // ...
}
```

If you don't think either of these an improvement, your formulation is totally fine -- just fine-tuning :-)

phillebaba: You are right and I think the second option is smart.

runtime/cert/cert.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+	"net/http"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	ClientCert = "certFile"
+	ClientKey  = "keyFile"
+	CACert     = "caFile"
```
hiddeco, line 32: I would add a suffix to the names, as we also do for `*Condition`, etc. As `*KeyKey` is a bit unfortunate, I would maybe use `Identifier`, `ID`.

hiddeco (commented):
I think this util can also be used in the source-controller for e.g. the Helm repository indexes.

runtime/cert/example_cert.go
```diff
@@ -0,0 +1,51 @@
+/*
```
stefanprodan, line 1: Please add `_test` to this file, we don't want these certs in flux binary or the controllers.

hiddeco: It may be even better to simply use the fixtures from https://cs.opensource.google/go/x/crypto/+/master:ssh/testdata/keys.go

phillebaba: Ok I will remove the certs from this PR.

hiddeco:
@phillebaba this PR is still waiting on an update. Couple of additional comments:

* Due to all the efforts in #101, the introduction of the package requires a `doc.go` file to outline the package.
* Comments are now up to 120 characters in length
* Given it only deals with TLS, I would maybe also call the package `tls` and not `certs`

runtime/cert/example_cert.go
phillebaba: @hiddeco I had a look at the testdata and I am a bit confused about how that can be used to replace the test certificate currently in the PR.

hiddeco (approved):

PR #106, opened 2021-06-08
Update internal `fluxcd/pkg` deps to latest

hiddeco:


stefanprodan:
Please bump Go to 1.16 in GH Actions

PR #109, opened 2021-06-11
Add generic patches in Kustomization

Kissy:
Will allow patching multiple resources instead of a single existing one as
StrategicMerge & JSON6902 are forced to target existing named resources.

This is related to [kustomize-controller PR #364](https://github.com/fluxcd/kustomize-controller/pull/364).

Signed-off-by: Guillaume Le Biller <glebiller@Traveldoo.com>


stefanprodan (approved):
LGTM

Thanks @Kissy

PR #111, opened 2021-06-16
Add controller runtime envtest GH action

stefanprodan:
Use [Envtest binaries manager](https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest) to prep the GH runner.


hiddeco:
If we introduce `setup-envest` in our projects, doesn't this mean that the `kubebuilder` action becomes obsolete because it would only be used for scaffolding APIs, etc?

hiddeco (approved):

PR #113, opened 2021-06-24
Trace log level

### LINKED ISSUE #112: Add trace log level
In https://github.com/fluxcd/image-automation-controller/pull/190 I added some tracing output to the image automation controller. Since that controller uses pkg/runtime/logging, and that doesn't know about `--log-level=trace`, I had to put the tracing in the debug log level. It would be good to recognise the trace level, though.

 - parse `--log-level=trace`
 - add a const for the log level so it can be referred to in `log.V(CONST).Info(...)`

squaremo:
Fixes #112, by adding `trace` to the possible log level values.

The underlying logging library, zap, will accept "custom debug levels" (i.e., levels numerically less than debug -- its levels are in ascending order of importance). It logs these with e.g., `LEVEL(-2)`, but there's no way around that. At least the mechanism works.

### DIFF
--- a/runtime/logger/logger.go
+++ b/runtime/logger/logger.go
@@ -39,6 +39,15 @@ var levelStrings = map[string]zapcore.Level{
 	"error": zapcore.ErrorLevel,
 }
 
+// These are for convenience when doing log.V(...) to log at a
+// particular level. They correspond to the logr equivalents of the
+// zap levels above.
+const (
+	TraceLevel = 2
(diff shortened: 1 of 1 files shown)


stefanprodan:
We’ll need to allow trace in here https://github.com/fluxcd/flux2/blob/main/internal/flags/log_level.go

stefanprodan (approved):
LGTM

Thanks @squaremo ☕

PR #121, opened 2021-07-08
Remove deprecated `ReconcileAtAnnotation`

### LINKED ISSUE #120: Remove deprecated `ReconcilateAtChangedPredicate` from `runtime/predicates`
As suggested in https://github.com/fluxcd/pkg/pull/101#discussion_r664441012

hiddeco:
This includes the removal of the predicate and changes to the
`ReconcileAnnotationValue` function.

In retrospect, it would have been better if we at the time had changed
the `LastHandledReconcileAt` field in the `ReconcileRequestStatus`
to something like `LastHandledReconcileRequest`. But I guess this has
now become "weirdness due to historical reasons".

Fixes #120


stefanprodan (approved):

squaremo (approved):
Looks reasonable to me. Thanks Hidde 🍇

PR #123, opened 2021-07-08
Make it easier to send events without metadata

### LINKED ISSUE #119: Make it easier to send events without metadata
https://github.com/fluxcd/source-controller/commit/26ebbe5a71e20c77e96c88d1e5126a339bb715d1 mentions:

```
  Passing `nil` every time there isn't any metadata for an event quickly
  becomes cumbersome; we should look into an `EventWithMetadata` and/or
  `EventfWithMetadata`, or some other way to _optionally_ provide
  metadata without annoying the consumer.
  ```

### PR DESCRIPTION
**Pull request by hiddeco**
While validating the runtime package changes, it quickly become
cumbersome to pass `nil` for every event that didn't require any
metadata to be attached.

This commit solves this annoyance by renaming the existing methods to
`EventWithMeta` and `EventWithMetaf`, and reintroducing the `Event` and
`Eventf` methods with signatures that do not take any metadata string
map.

Fixes #119

### DIFF
--- a/runtime/controller/events.go
+++ b/runtime/controller/events.go
@@ -61,12 +61,24 @@ func MakeEvents(mgr ctrl.Manager, controllerName string, ext *events.Recorder) E
 }
 
 // Event emits a Kubernetes event, and forwards the event to the ExternalEventRecorder if configured.
-func (e Events) Event(ctx context.Context, obj client.Object, metadata map[string]string, severity, reason, msg string) {
-	e.Eventf(ctx, obj, metadata, severity, reason, msg)
+// Use EventWithMeta or EventWithMetaf if you want to attach metadata to the external event.
+func (e Events) Event(ctx context.Context, obj client.Object, severity, reason, msg string) {
+	e.EventWithMetaf(ctx, obj, nil, severity, reason, msg)
 }
 
 // Eventf emits a Kubernetes event, and forwards the event to the ExternalEventRecorder if configured.
-func (e Events) Eventf(ctx context.Context, obj client.Object, metadata map[string]string, severity, reason, msgFmt string, args ...interface{}) {
+// Use EventWithMeta or EventWithMetaf if you want to attach metadata to the external event.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by stefanprodan — APPROVED** [review]

File: runtime/controller/events.go [hunk 1]
```diff
@@ -61,12 +61,24 @@ func MakeEvents(mgr ctrl.Manager, controllerName string, ext *events.Recorder) E
...
 }
 
 // Eventf emits a Kubernetes event, and forwards the event to the ExternalEventRecorder if configured.
-func (e Events) Eventf(ctx context.Context, obj client.Object, metadata map[string]string, severity, reason, msgFmt string, args ...interface{}) {
+// Use MetaEvent or MetaEventf if you want to attach metadata to the external event.
+func (e Events) Eventf(ctx context.Context, obj client.Object, severity, reason, msgFmt string, args ...interface{}) {
+	e.MetaEventf(ctx, obj, nil, severity, reason, msgFmt, args...)
+}
+
+// MetaEvent emits a Kubernetes event, and forwards the event and metadata to the ExternalEventRecorder if configured.
```
squaremo, line 75: What kind of things end up in metadata? Calling these `MetaEvent` says to me that it's a different kind of event (one that's on the "meta" level); but I suspect that it's just an event that has extra, caller-supplied fields, is that right?

squaremo: If that latter, may I suggest `EventWithMeta` and `EventWithMetaf` as the method names.

stefanprodan: We use the metadata to enrich the event with things like source revision, branch, tag, commit. In the future we would like to make the metadata typed, so it can be used to template alerts.

hiddeco: I was in doubt about the naming too @squaremo, but your suggestion is a combination of current state of the PR and what I suggested to myself in: https://github.com/fluxcd/pkg/issues/119

PR #124, opened 2021-07-08
don't install kubectl as part of kubebuilder tools

makkes:
While the kubectl action defined in this repo installs a recent
version of kubectl, all action workflows use the kubectl binary
installed from kubebuilder which is very old (v1.16.4 for kubebuilder
tools version 2.3.1).

This is from a recent workflow run:

https://github.com/fluxcd/helm-controller/pull/294/checks?check_run_id=3018781524#step:10:9

```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.4", ...
```

### DIFF
--- a/actions/kubebuilder/entrypoint.sh
+++ b/actions/kubebuilder/entrypoint.sh
@@ -6,6 +6,9 @@ VERSION=${1:-2.3.1}
 
 curl -sL https://go.kubebuilder.io/dl/${VERSION}/linux/amd64 | tar -xz -C /tmp/
 
+# remove kubectl as this is supposed to be installed separately
+rm /tmp/kubebuilder_${VERSION}_linux_amd64/kubectl
+
 mkdir -p $GITHUB_WORKSPACE/kubebuilder
 mv /tmp/kubebuilder_${VERSION}_linux_amd64/* $GITHUB_WORKSPACE/kubebuilder/
 ls -lh $GITHUB_WORKSPACE/kubebuilder/bin


actions/kubebuilder/entrypoint.sh
```diff
@@ -8,6 +8,8 @@ curl -sL https://go.kubebuilder.io/dl/${VERSION}/linux/amd64 | tar -xz -C /tmp/
 
 mkdir -p $GITHUB_WORKSPACE/kubebuilder
 mv /tmp/kubebuilder_${VERSION}_linux_amd64/* $GITHUB_WORKSPACE/kubebuilder/
+# remove kubectl as this is supposed to be installed separately
```
hiddeco, line 11: Wouldn't it be better to remove this before doing `mv`?

makkes: you're right. Fixed.

stefanprodan:
Are we sure that deleting kubectl doesn't break CI? I don't know if kubebuilder looks for kubectl in that dir or it uses the one in PATH.

makkes:
> Are we sure that deleting kubectl doesn't break CI? I don't know if kubebuilder looks for kubectl in that dir or it uses the one in PATH.

I must admit I don't really know why GH Actions is preferring `/usr/local/kubebuilder/bin/kubectl` over `/usr/local/bin/kubectl` in the first place as I would assume that `/usr/local/bin` appears before `/usr/local/kubebuilder/bin` in the `PATH`.

hiddeco:
Based on the information in https://book.kubebuilder.io/reference/envtest.html, I would say that it looks for all binaries in `KUBEBUILDER_ASSETS` first, and this may indeed break default behavior.

makkes:
> Based on the information in https://book.kubebuilder.io/reference/envtest.html, I would say that it looks for all binaries in `KUBEBUILDER_ASSETS` first, and this may indeed break default behavior.

In that case I'll have to find a way to prefer /usr/local/bin/ over $RUNNER_WORKSPACE/$(basename $GITHUB_REPOSITORY)/kubebuilder/bin/. 👍🏻

hiddeco:
The problem is that `echo "$PATH" >> $GITHUB_PATH` prepends to the `$PATH` (end of YAML doc), instead of appending. See: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#adding-a-system-path

stefanprodan:
We no longer need kubebuilder in CI, closing this.

PR #126, opened 2021-07-12
Order ReconcilingCondition before ReadyCondition

hiddeco:
This is more logical, as both `StalledCondition` and
`ReconcilingCondition` are abnormal-true types, and should thus only be
present on a resource when the condition is True.


phillebaba (approved):
LGTM

PR #127, opened 2021-07-12
runtime/patch: add integration test using testenv

hiddeco:
This was recorded in 723c6bd1e2a4daba07dee8e49cae853665303f15 as a TODO
because the `testenv` package was not yet available.

The integration tests make use of a `Fake` Kubernetes resource object
that for the time being has been made public from a `testdata` package
in `conditions`. In a future iteration we may want to move it to
`pkg/apis` so that it is available for generic usage during e.g.
testing.


runtime/patch/patch_test.go
```diff
@@ -0,0 +1,919 @@
...
+				},
+			}
+			obj.SetOwnerReferences(refs)
+
+			t.Log("Patching the unstructured object")
+			g.Expect(patcher.Patch(ctx, obj)).To(Succeed())
+
+			t.Log("Validating that the status has been preserved")
+			g.Expect(err).NotTo(HaveOccurred())
+			strV, found, err := unstructured.NestedString(obj.Object, "status", "observedValue")
```
phillebaba, line 101: NIT: Might be better to use a more descriptive variable name.

phillebaba (approved):

PR #129, opened 2021-07-18
Add setup-envtest in Makefile

darkowlzz:
This adds setup-envtest in the Makefile to make it easier to run
envtests without the need to download the kubebuilder binaries
separately.
The default value of KUBEBUILDER_ASSETS can be overridden by the value
set in CI, actions/envtest github action.

Signed-off-by: Sunny <darkowlzz@protonmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@ VER?=0.0.1
 MODULES=$(shell find . -mindepth 2 -maxdepth 4 -type f -name 'go.mod' | cut -c 3- | sed 's|/[^/]*$$||' | sort -u | tr / :)
 targets=$(addprefix test-, $(MODULES))
 root_dir=$(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+ENVTEST_BIN_VERSION?=latest
+KUBEBUILDER_ASSETS?=$(shell $(SETUP_ENVTEST) use -i $(ENVTEST_BIN_VERSION) -p path)
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
@@ -25,8 +27,8 @@ vet-%:
 generate-%: controller-gen
 	cd $(subst :,/,$*); $(CONTROLLER_GEN) object:headerFile="$(root_dir)/hack/boilerplate.go.txt" paths="./..."
 
-test-%: generate-% tidy-% fmt-% vet-%
-	cd $(subst :,/,$*); go test ./... -coverprofile cover.out
+test-%: generate-% tidy-% fmt-% vet-% setup-envtest
+	cd $(subst :,/,$*); KUBEBUILDER_ASSETS=$(KUBEBUILDER_ASSETS) go test ./... -coverprofile cover.out
 
 release-%:
 	$(eval REL_PATH=$(subst :,/,$*))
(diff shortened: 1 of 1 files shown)


squaremo:
This looks handy! Should the test* targets depend on `setup-envtest`, so that `SETUP_ENVTEST` is assigned a value and ultimately, the expression for `KUBEBUILDER_ASSETS` can be evaluated?

darkowlzz:
> Should the test* targets depend on `setup-envtest`, so that `SETUP_ENVTEST` is assigned a value and ultimately, the expression for `KUBEBUILDER_ASSETS` can be evaluated?

Oh right, didn't think about that :smiley: . Thanks, I'll add that.

darkowlzz:
Added `setup-envtest` dependency on `test-*` target.

Makefile
```diff
@@ -51,3 +53,19 @@ CONTROLLER_GEN=$(GOBIN)/controller-gen
...
+
+# Find or download setup-envtest
+setup-envtest:
+ifeq (, $(shell which setup-envtest))
+	@{ \
+	set -e ;\
+	SETUP_ENVTEST_TMP_DIR=$$(mktemp -d) ;\
+	cd $$SETUP_ENVTEST_TMP_DIR ;\
+	go mod init tmp ;\
+	go get sigs.k8s.io/controller-runtime/tools/setup-envtest@latest ;\
```
relu, line 65: Wouldn't it be more reliable to stick to an absolute version?
I think `go install` might be a better option now (since go 1.16) to manage installing these dev utils.

darkowlzz: Hi, thanks for reviewing.
I couldn't find any releases for setup-envtest in https://github.com/kubernetes-sigs/controller-runtime/releases and trying to install it with the overall controller-runtime release results in error

```console
go install sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.9.2: module sigs.k8s.io/controller-runtime@v0.9.2 found, but does not contain package sigs.k8s.io/controller-runtime/tools/setup-envtest
```

I can pin it to a commit. Will that be okay?

Regarding the `go install` option, I followed what's done for controller-gen and gen-crd-api-reference-docs (in the source-controller repo), which is based on the kubebuilder generated makefile. `go install` works well. Should I update the controller-gen target to also use go install? and in other repos as well?

relu: I see that in the [readme](https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest#readme) they also reference `@latest` so maybe that's fine for now.

> Should I update the controller-gen target to also use go install? and in other repos as well?

I'm inclined to say yes, but I think it would make sense to check with the others as well. I think `go install` is more elegant, even though this [scaffolding](https://github.com/kubernetes-sigs/kubebuilder/blob/1831527c09c7ad54bd1782f0e01550b425ef9ae1/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go#L60) is provided by kubebuilder by default. For now, you can probably leave it as is and we can explore the option to use `go install` later.

hiddeco (approved):
This was something that was still on my to-do list, thanks Sunny. :1st_place_medal:

hiddeco:
(It needs a rebase before I can press "merge")

darkowlzz:
Rebased :)

PR #130, opened 2021-07-20
runtime/predicates: Add unit tests

darkowlzz:
Adds unit tests for ReconcileRequestedPredicate.Update.

### DIFF
--- a/runtime/predicates/reconcile_requested_test.go
+++ b/runtime/predicates/reconcile_requested_test.go
@@ -1,0 +1,114 @@
+/*
+Copyright 2021 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


runtime/predicates/reconcile_requested_test.go
```diff
@@ -0,0 +1,98 @@
+package predicates
```
hiddeco, line 1: Can you please add a copyright header? We do this for any Go file in our projects.

darkowlzz: Added.

hiddeco (approved):
:bowing_man::clap: x :100:

hiddeco:
(Rebase again, I am back from vacation so this should happen less often in the future).

darkowlzz:
np, rebased :slightly_smiling_face: Thanks

PR #132, opened 2021-07-25
Add testserver (git + artifact) tests

darkowlzz:
- Adds basic tests for the testservers (gittestserver and artifact/http server) to verify that they run and function correctly.
- Adds the missing auth configuration in the https gittestserver.
- Adds testdata/certs package, based on the testdata from the source-controller repo, and use it for testing gittestserver and httpserver.


hiddeco (approved):
LGTM, thank you @darkowlzz :100:

PR #133, opened 2021-07-26
runtime/conditions: add kstatus condition helpers

hiddeco:
This commit adds getter and setter helpers to `runtime/conditions` to
make it easier to mark resources with kstatus conditions, and observe
their state.

- The `MarkReconciling` and `MarkStalled` functions delete their
  counterpart condition type while marking the resource with the
  kstatus condition, making it easier to adhere to the kstatus
  standards.
- The `IsReady`, `IsStalled` and `IsReconciling` functions take the
  existence of their counterpart condition types into account, ensuring
  conclusive conditional checks.

`MarkReady` was not taken into consideration, it could technically be
added as syntax sugar, but this would first require a discussion about
a counterpart as you likely want to be able to mark something as "not
ready".

To determine a Ready condition at present, the advice is to compose one
using `SetSummary` based on the defined condition types from your own
API, for example:

```go
conditions.SetSummary(
	&obj,
	meta.ReadyCondition,
	conditions.WithConditions(
		foov1.FooCondition,
		foov1.BarCondition,
		foov1.BazCondition,
	),
)
```

### DIFF
--- a/runtime/conditions/getter.go
+++ b/runtime/conditions/getter.go
@@ -57,8 +57,8 @@ func Has(from Getter, t string) bool {
 	return Get(from, t) != nil
 }
 
-// IsTrue is true if the condition with the given type is True, otherwise it return false if the condition is not True
-// or if the condition does not exist (is nil).
+// IsTrue is true if the condition with the given type is True, otherwise it is false if the condition is not True or if
+// the condition does not exist (is nil).
 func IsTrue(from Getter, t string) bool {
 	if c := Get(from, t); c != nil {
 		return c.Status == metav1.ConditionTrue
@@ -66,8 +66,8 @@ func IsTrue(from Getter, t string) bool {
 	return false
 }
 
-// IsFalse is true if the condition with the given type is False, otherwise it return false if the condition is not
-// False or if the condition does not exist (is nil).
+// IsFalse is true if the condition with the given type is False, otherwise it is false if the condition is not False or
+// if the condition does not exist (is nil).
 func IsFalse(from Getter, t string) bool {
 	if c := Get(from, t); c != nil {
 		return c.Status == metav1.ConditionFalse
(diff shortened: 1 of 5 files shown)


scottrigby (approved):
These are excellent! Adding a to-do for ensuring this is part of the forthcoming Guide on writing a Flux controller ([see notes](https://docs.google.com/document/d/1M0y74iwpqSA-Eu0p3efMD-UCDzEYG23eJsmXXEWihaM/edit#heading=h.6d1ykjtxgc8m)). 💯 lgtm

runtime/patch/patch.go
```diff
@@ -71,7 +71,7 @@ import (
 //					Conditions: []string{
 //						meta.ReadyCondition,
 //						meta.ReconcilingCondition,
-//						meta.ProgressingReason,
+//						meta.StalledCondition,
```
scottrigby, line 74: Also glad we found this little typo. Yay for fixing docs 🥳

runtime/conditions/getter.go
```diff
@@ -83,6 +83,24 @@ func IsUnknown(from Getter, t string) bool {
 	return true
 }
 
+// IsReady is true if IsStalled and IsReconciling are False, and meta.ReadyCondition is True, otherwise it is false if
+// the condition is not True or if it does not exist (is nil).
+func IsReady(from Getter) bool {
+	return !IsStalled(from) && !IsReconciling(from) && IsTrue(from, meta.ReadyCondition)
```
stefanprodan, line 89: Why isn't this just `IsTrue(from, meta.ReadyCondition)` did we changed the behaviour of `Ready`?

hiddeco: Because `Stalled` and `Reconciling` are successors to `Ready`, and should take precedence if they exist over `Ready`.

> For resources that have not adopted the recommended conditions described above [...]

https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus

stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #140, opened 2021-08-09
Add Access Control List API types and conditions

stefanprodan:
Ref: https://github.com/fluxcd/flux2/issues/1704


apis/acl/conditions.go
```diff
@@ -0,0 +1,31 @@
...
+	// AccessDeniedCondition indicates that access to a resource has been denied by an ACL assertion.
+	// The Condition adheres to an "abnormal-true" polarity pattern, and MUST only be present on the resource if the
+	// Condition is True.
+	AccessDeniedCondition string = "AccessDenied"
+)
+
+// These constants define the Condition reasons for when the GitOps Toolkit components perform ACL assertions.
+const (
+	// AccessDeniedReason indicates that access to a resource has been denied by an ACL assertion.
+	AccessDeniedReason string = "AccessDenied"
```
hiddeco, line 30: Would there be more context defining / informative reasons for when the Condition type is `AccessDenied`?

stefanprodan: Not right now, as our ACL is based on namespace labels only, but if we add more matches, such as caller labels, then we could have 2 different reasons.

stefanprodan: Note that I've added here the reason so that it gets copied to `Ready`, as our helpers doesn't copy the condition type to the Ready reason.

hiddeco (approved):

PR #141, opened 2021-08-09
Add ACL authorization helper to runtime

stefanprodan:
Ref: https://github.com/fluxcd/flux2/issues/1704

### DIFF
--- a/runtime/acl/acl.go
+++ b/runtime/acl/acl.go
@@ -1,0 +1,76 @@
+/*
+Copyright 2021 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package acl
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	aclapi "github.com/fluxcd/pkg/apis/acl"
+)
+
+// Authorization is an ACL helper for asserting access to cross-namespace references.
+type Authorization struct {
+	client client.Client
+}
+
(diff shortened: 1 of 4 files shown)


runtime/acl/acl.go
```diff
@@ -0,0 +1,70 @@
...
+	aclapi "github.com/fluxcd/pkg/apis/acl"
+)
+
+// Authorization is an ACL helper for asserting access to cross-namespace references.
+type Authorization struct {
+	Client client.Client
+}
+
+// HasAccessTo asserts if an object has access to a cross-namespace reference based on the ACL defined on the referenced object
+func (a *Authorization) HasAccessTo(ctx context.Context, object client.Object, reference types.NamespacedName, acl *aclapi.AccessFrom) (bool, error) {
```
hiddeco, line 38:
```suggestion
// ObjHasAccessTo asserts if an object has access to a cross-namespace reference based on the ACL defined on the referenced object.
func (a *Authorization) ObjHasAccessTo(ctx context.Context, object client.Object, reference types.NamespacedName, acl *aclapi.AccessFrom) (bool, error) {
```

runtime/acl/acl.go
```diff
@@ -0,0 +1,70 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	aclapi "github.com/fluxcd/pkg/apis/acl"
+)
+
+// Authorization is an ACL helper for asserting access to cross-namespace references.
+type Authorization struct {
+	Client client.Client
+}
```
hiddeco, line 35: Wondering if we want to make the client publicly accessible, or instead rely on a factory method and make it private, ensuring we only expose `HasAccessToRef` (and not the full client).

runtime/acl/acl.go
```diff
@@ -0,0 +1,76 @@
...
+// Authorization is an ACL helper for asserting access to cross-namespace references.
+type Authorization struct {
+	client client.Client
+}
+
+// NewAuthorization takes a controller runtime client and return an Authorization object that allows asserting
+// access to cross-namespace references.
+func NewAuthorization(kubeClient client.Client) *Authorization {
+	return &Authorization{client: kubeClient}
+}
```
hiddeco, line 41:
```suggestion
// NewAuthorization takes a controller-runtime client and returns an Authorization object that allows asserting
// access to cross-namespace references.
func NewAuthorization(kubeClient client.Client) *Authorization {
	return &Authorization{client: kubeClient}
}
```

Also not sure about the `kubeClient`, but I assume you opted for this because using `client` would be a duplicate to the field name, and it makes it easier to know "what" client while just looking at the method signature.

hiddeco (approved):
One last remark on the new doc-block, but other than that LGTM. Thank you @stefanprodan :100:

runtime/acl/acl.go
stefanprodan: We have a client pkg in runtime, I just wanted to point out it's not that :)

PR #144, opened 2021-08-17
Account for reconciling condition in patch example

squaremo:
At present the example of how to use patch.Helper unconditionally clears the reconciling condition on a "clean" exit (no error or explicit requeue). This is on the assumption that the condition is used only within the Reconcile method to indicate that the resource is being processed.

However, in general it's possible for Reconcile to exit cleanly without the reconciliation having reached a ready state, but without definitively failing either. For example, a Deployment would need to visit Reconcile several times during a rolling upgrade before it has rolled out a new pod spec.

On the expectation that people may borrow the example, I think it is better to treat the Reconciling condition as an indication that there is more work to be done.

### DIFF
--- a/runtime/patch/patch.go
+++ b/runtime/patch/patch.go
@@ -50,6 +50,9 @@ import (
 // This ensures any modifications made to the spec and the status (conditions) object of the resource are always
 // persisted at the end of a reconcile run.
 //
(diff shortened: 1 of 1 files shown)


hiddeco (approved):
Thank you @squaremo :apple:

PR #146, opened 2021-08-24
Remove use of deprecated io/ioutil package

allenporter:
Remove deprecated calls and replace with suggestions from the deprecation notice.

Update test instructions since `make test` doesn't work. Now uses `make all`, which runs tests for all targets in all subdirectories.

Issue: https://github.com/fluxcd/flux2/issues/1658


stefanprodan (approved):
LGTM 

Thanks @allenporter ☕

PR #148, opened 2021-09-16
testenv: Remove klog init flags

darkowlzz:
This change allows test environments with glog to work with testenv.

### Background

image-reflector-controller uses badger for database. [dgraph-io/badger](https://github.com/dgraph-io/badger) go package depends on  [dgraph-io/ristretto](https://github.com/dgraph-io/ristretto), which depends on glog. When glog is used, it sets `log_dir` flag, [refer](https://github.com/golang/glog/blob/23def4e6c14b4da8ac2ed8007337bc5eb5007998/glog_file.go#L41) .
Due to the glog transitive dependency of badger, when badger is used in testenv based tests, the testenv package init() initialization of the klog flags and badger set the glog flags conflict, resulting in some redefined flags.

This can be reproduced by adding the following in `TestMain()`:

```golang
    badgerDir, err := ioutil.TempDir(os.TempDir(), "badger")
    if err != nil {
        panic(fmt.Sprintf("Failed to create temporary directory for badger: %v", err))
    }

    badgerDB, err := badger.Open(badger.DefaultOptions(badgerDir))
    if err != nil {
        panic(fmt.Sprintf("Failed to create new Badger database: %v", err))
    }
```
with `"github.com/dgraph-io/badger/v3"` import.

Failure output:
```console
/tmp/go-build1837318169/b001/controllers.test flag redefined: log_dir
panic: /tmp/go-build1837318169/b001/controllers.test flag redefined: log_dir

goroutine 1 [running]:
flag.(*FlagSet).Var(0xc00003c240, {0x19d5ad8, 0x26f5090}, {0x17b67fd, 0x7}, {0x17f2153, 0x2f})
        ~/.gimme/versions/go1.17.1.linux.amd64/src/flag/flag.go:879 +0x2f4
flag.(*FlagSet).StringVar(...)
        ~/.gimme/versions/go1.17.1.linux.amd64/src/flag/flag.go:762
k8s.io/klog/v2.InitFlags(0x1494a27)
        ~/go/pkg/mod/k8s.io/klog/v2@v2.8.0/klog.go:427 +0x55
github.com/fluxcd/pkg/runtime/testenv.init.0()
        ~/go/pkg/mod/github.com/fluxcd/pkg/runtime@v0.13.0-rc.2/testenv/testenv.go:56 +0x1b
FAIL    github.com/fluxcd/image-reflector-controller/controllers        0.027s
FAIL
```

Based on an [example](https://github.com/kubernetes/klog/blob/f8e668dbaa5f6f0e6a5c24ffd7667263840d79ae/examples/coexist_glog/coexist_glog.go) in klog repo, glog and klog can coexist with separate FlagSets.

With this change, the image-reflector-controller is able to use testenv.

### DIFF
--- a/runtime/testenv/testenv.go
+++ b/runtime/testenv/testenv.go
@@ -52,13 +52,6 @@ var (
 	env *envtest.Environment
 )
 
-func init() {
-	klog.InitFlags(nil)
-	logger := klogr.New()
-	log.SetLogger(logger)
-	ctrl.SetLogger(logger)
-}
-
 var (
 	cacheSyncBackoff = wait.Backoff{
 		Duration: 100 * time.Millisecond,
@@ -118,7 +111,32 @@ func WithCRDPath(path ...string) Option {
 //
 // NOTE: This function should be called only once for each package you are running tests within, usually the environment
 // is initialised in a suite_test.go or <package>_test.go file within a `TestMain` function.
+//
+// When a testenv Environment is created, it initializes the
+// controller-runtime's deferred logger with a default logger based on klog. In
+// order to override this behavior, the controller-runtime logger can be
+// initialized before creating testenv Environment.
+//
+//  import (
+//      "testing"
+//
+//      "github.com/fluxcd/pkg/runtime/testenv"
+//      ctrl "sigs.k8s.io/controller-runtime"
+//      "sigs.k8s.io/controller-runtime/pkg/log/zap"
+//  }
+//
+//  func TestMain(m *testing.M) {
+//      zlog := zap.New(zap.UseDevMode(true))
+//      ctrl.SetLogger(zlog)
+//
+//      testEnv = testenv.New()
+//      ...
+//  }
+//
 func New(o ...Option) *Environment {
+	// Set a default logger if not set already.
+	log.SetLogger(klogr.New())
+
 	opts := options{}
 	for _, apply := range o {
 		apply(&opts)


runtime/testenv/testenv.go
```diff
@@ -53,7 +54,14 @@ var (
...
 
 func init() {
-	klog.InitFlags(nil)
+	// NOTE: The following enables glog and klog to coexist. If a test depends
+	// on a package that uses glog, klog flag initialization here would result
+	// in redeclaration of some flags. The following creates a new FlagSet and
+	// uses it to initializes klog.
+	// Refer: https://github.com/kubernetes/klog/blob/f8e668dbaa5f6f0e6a5c24ffd7667263840d79ae/examples/coexist_glog/coexist_glog.go
+	klogFlags := flag.NewFlagSet("klog", flag.ExitOnError)
+	klog.InitFlags(klogFlags)
```
squaremo, line 63: Do you need to tell klogr about the flags?

squaremo: OK; I think it is worth pointing out in the signpost comment that it's effectively throwing away the flag values, so that future readers will not skip through and assume they are passed. Since we're talking about global variables here, it would be a reasonable assumption given the comment that say, `klog.InitFlags(...)` _assigns_ the flagset to be used by klogr. (Or does it actually do this? There you go, I'm already confused ...)

darkowlzz: :smile: yes, it is a bit confusing. I spent some more time going through the klog repo and understanding how it's supposed to be used. It looks like `klog.InitFlags()` is independent of klogr. There's a private variable called `logging` that's used for setting the flag values on the given `FlagSet` to `InitFlags()`.
klogr just imports klog and uses the global klog. There's no API to pass a klog instance. If any values are set on the global klog using the global flagset, it'll  be available to klogr. No way to pass a different klog.

The problem in this case is that since the testenv's init gets called at the very beginning, there's nothing that can be done in testenv package from preventing someone trying to register another flag later. In the case of badger, since they use glog, even they don't have a way to not do this without dropping the glog dependency. The only way around this that I can think of is to avoid filling the global flags with klog init flags for this particular situation.

I'll update the comment to mention that the klog flags are not passed to the klogr that's set as the controller runtime logger. It can be manually passed by creating a custom klog and  klogr, and overwriting the set controller-runtime logger.

darkowlzz:
I spent more time validating the idea of overriding the controller-runtime(c-n) logger set by the testenv `init()` and found that the c-n deferred logger can only be set once. Importing testenv package sets the c-n logger in its `init()`, subsequent calls to `ctrl.SetLogger()` becomes no-op. Since this is a library, I think it'll be better to not hijack the logger by just importing the package and allow the users to set their own logger. To enable that, I've moved the logger initialization code into a new function `initLogger()`, which is called when a new testenv `Environment` is created. This enables the users to set their custom logger before creating testenv. The testenv's attempt to set the c-n logger becomes no-op if it's already set.

Also, added godoc example to show how a custom logger can be set before creating testenv.

runtime/testenv/testenv.go
```diff
@@ -52,11 +53,23 @@ var (
...
+	// because they aren't set on the global FlagSet. To prevent conflicting
+	// with any flags registered later during the test setup, the flags are left
+	// empty. In the tests, a custom logger can be created with all the
+	// necessary flags and assigned to the controller-runtime logger explicitly
+	// before creating testenv Environment.
+	klogFlags := flag.NewFlagSet("klog", flag.ExitOnError)
+	klog.InitFlags(klogFlags)
 	logger := klogr.New()
 	log.SetLogger(logger)
-	ctrl.SetLogger(logger)
```
darkowlzz, line 59: This is redundant. It internally calls `log.SetLogger()`.

runtime/testenv/testenv.go
squaremo: I've looked at the klog code now too. The flag values aren't thrown away as I speculated, they are used to populate a global struct, as you say. And `klogr` is just an adapter that uses the global. So that's all fine from our point of view, since all we want to do is avoid using the same flagset as glog.

hiddeco (approved):
Reasoning seems to make sense, thanks for getting to the bottom of it, and for including an example :1st_place_medal:

runtime/testenv/testenv.go
```diff
@@ -52,11 +53,23 @@ var (
...
-func init() {
-	klog.InitFlags(nil)
+func initLogger() {
+	// NOTE: Set the klog flags on a separate FlagSet to avoid conflicting with
+	// any flags set by the application code. For example, an application
+	// importing glog would conflict with the flags set by klog, resulting in
+	// redeclaration of the same flags.
+	// This is based on klog's coexist with glog example, refer:
+	// https://github.com/kubernetes/klog/blob/f8e668dbaa5f6f0e6a5c24ffd7667263840d79ae/examples/coexist_glog/coexist_glog.go
+	// The klog flags set below are not available to the klogr logger below
```
squaremo, line 63: This, we have discovered, is not correct. `klogr` uses the public API of `klog`, which in turn makes use of the flagset to set itself up.

darkowlzz: Removed it with a statement about how to override this logger setup.

runtime/testenv/testenv.go
```diff
@@ -52,11 +53,20 @@ var (
...
+	// any flags set by the application code. For example, an application
+	// importing glog would conflict with the flags set by klog, resulting in
+	// redeclaration of the same flags.
+	// This is based on klog's coexist with glog example, refer:
+	// https://github.com/kubernetes/klog/blob/f8e668dbaa5f6f0e6a5c24ffd7667263840d79ae/examples/coexist_glog/coexist_glog.go
+	// To override this logger setup, a custom logger can be created and
+	// assigned to the controller-runtime logger explicitly before creating
+	// testenv Environment.
+	klogFlags := flag.NewFlagSet("klog", flag.ExitOnError)
+	klog.InitFlags(klogFlags)
```
squaremo, line 67: Do you need to call `klogFlags.Parse()`? I don't think klog calls it.

darkowlzz: Searching for `Parse()` in klog repo, I only see them in the example code. So, yes, you're probably right about that.

darkowlzz: I tried setting a parse here `klogFlags.Parse(os.Args[1:])` , but it results in the following error :
```console
flag provided but not defined: -test.paniconexit0
Usage of klog:
  -add_dir_header
        If true, adds the file directory to the header of the log messages
  -alsologtostderr
        log to standard error as well as files
...
```
Because some extra flags get set in the test run that klog flagset isn't aware of.
I can't think of a way to remove unknown flag args before passing to this Parse() in a library. Maybe this is the reason it's not parsed here in the first place.
This leaves it up to the user to construct their own flagset and configure their logger. The library only does the minimum default setup.

darkowlzz: So that makes me feel like dropping `klog.InitFlags()` completely since the flags are not parsed and used at all.

darkowlzz: And if we do that, we can remove all the extra comments about coexisting with glog. It'll just work with glog without doing anything.

squaremo: I thought that might be the case. And if it can't be done here, I don't think a user of the library has much of a chance either. It worked before because the test environment didn't mandate using `klogr`; could this package go back to that arrangement?

darkowlzz: I think it did mandate klogr before but didn't care to parse the flags and left no way to change it because of running the logger setup in the init.
I can just remove the separate flagset code and `klog.InitFlags()`.  That will be equivalent to the previous behavior.

darkowlzz:
Renamed the title of the PR and the details in the commit message because it no longer has to do with glog directly.

runtime/testenv/testenv.go
```diff
@@ -118,7 +111,32 @@ func WithCRDPath(path ...string) Option {
...
+//      zlog := zap.New(zap.UseDevMode(true))
+//      ctrl.SetLogger(zlog)
+//
+//      testEnv = testenv.New()
+//      ...
+//  }
+//
 func New(o ...Option) *Environment {
+	// Set a default logger if not set already.
+	log.SetLogger(klogr.New())
```
squaremo, line 138: If this is equivalent to ctrl.SetLogger, it'd be better to use the same in both the example given and the code. (And if it's not, why do they differ?)

squaremo (approved):
I think this is the right way to go now, and I'm down to nit picks, so ✅ 
Thanks for your patience Sunny!

PR #149, opened 2021-09-20
runtime/conditions: Add merge option with generations

### LINKED ISSUE #142: runtime/conditions: take Condition generation into account during merge
At present the [merge operation](https://github.com/fluxcd/pkg/blob/11abd078e26f7eaf8c4c415fa1897e1d41993592/runtime/conditions/merge.go#L58) does not take the generation of conditions into account. To ensure the most relevant (and recent) condition is merged, it should be possible to take the generation recorded in the condition into account.

It could be possible to make this an opt-out (or opt-in) option using `mergeOptions`, see: https://github.com/fluxcd/pkg/blob/11abd078e26f7eaf8c4c415fa1897e1d41993592/runtime/conditions/merge_strategies.go#L33

darkowlzz:
Adds merge option `WithGenerations()` to consider the condition's observed
generation while merging. When withGenerations is set, the condition
groups are created based on the generation of the conditions. The
condition groups with the latest generations are grouped together and
sorted based on their merge priority. The condition groups with old
generations don't contain the up-to-date information and hence not
considered.
The merge behavior remains the same when no generation is specified.

Fixes #142

### DIFF
--- a/runtime/conditions/merge.go
+++ b/runtime/conditions/merge.go
@@ -41,15 +41,17 @@ type localizedCondition struct {
 // it is should be reflected in the target condition.
 //
 // More specifically:
-// 1. Conditions are grouped by status and polarity.
+// 1. Conditions are grouped by status, polarity and observed generation (optional).
 // 2. The resulting condition groups are sorted according to the following priority:
 //   - P0 - Status=True, NegativePolarity=True
 //   - P1 - Status=False, NegativePolarity=False
-//   - P2 - Condition=True, NegativePolarity=False
+//   - P2 - Status=True, NegativePolarity=False
 //   - P3 - Status=False, NegativePolarity=True
 //   - P4 - Status=Unknown
 // 3. The group with highest priority is used to determine status, and other info of the target condition.
 // 4. If the polarity of the highest priority and target priority differ, it is inverted.
+// 5. If the observed generation is considered, the condition groups with the latest generation get the highest
+// priority.
 //
 // Please note that the last operation includes also the task of computing the Reason and the Message for the target
 // condition; in order to complete such task some trade-off should be made, because there is no a golden rule for
@@ -98,6 +100,10 @@ func getConditionGroups(conditions []localizedCondition, options *mergeOptions) 
 		for i := range groups {
 			if groups[i].status == condition.Status &&
 				groups[i].negativePolarity == stringInSlice(options.negativePolarityConditionTypes, condition.Type) {
+				// If withLatestGeneration is true, add to group only if the generation match.
+				if options.withLatestGeneration && groups[i].generation != condition.ObservedGeneration {
+					continue
+				}
 				groups[i].conditions = append(groups[i].conditions, condition)
 				added = true
 				break
@@ -108,8 +114,22 @@ func getConditionGroups(conditions []localizedCondition, options *mergeOptions) 
 				conditions:       []localizedCondition{condition},
 				status:           condition.Status,
 				negativePolarity: stringInSlice(options.negativePolarityConditionTypes, condition.Type),
+				generation:       condition.ObservedGeneration,
 			})
 		}
+	}
+
+	// If withLatestGeneration is true, form a conditionGroups of the groups
+	// with the latest generation.
+	if options.withLatestGeneration {
+		latestGen := groups.latestGeneration()
+		latestGroups := conditionGroups{}
+		for _, g := range groups {
+			if g.generation == latestGen {
+				latestGroups = append(latestGroups, g)
+			}
+		}
+		groups = latestGroups
 	}
 
 	// sort groups by priority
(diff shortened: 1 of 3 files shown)


runtime/conditions/merge.go
```diff
@@ -208,3 +216,12 @@ func stringInSlice(s []string, val string) bool {
 	}
 	return false
 }
+
+func generationInSlice(s []int64, val int64) bool {
```
hiddeco, line 220:
```suggestion
func int64InSlice(s []int64, val int64) bool {
```

runtime/conditions/merge_strategies.go
```diff
@@ -231,3 +233,10 @@ func getFirstCondition(g conditionGroups, priority []string) *localizedCondition
 		return &topGroup.conditions[0]
 	}
 }
+
+// WithGenerations instructs merge to consider the conditions in the given observed generations only.
```
hiddeco, line 237: I think this should be:

```
// WithGenerations instructs merge to consider the conditions with the given observed generations only.

File: runtime/conditions/merge_test.go [hunk 3]
```diff
@@ -179,6 +179,64 @@ func TestMergeRespectPriority(t *testing.T) {
...
+			negativeConditions: []string{true1.Type},
+			conditions: []*metav1.Condition{
+				conditionWithGeneration(false1, 1),
+				conditionWithGeneration(true1, 2),
+				conditionWithGeneration(unknown1, 3),
+			},
+			mergeOpts: []MergeOption{WithGenerations(2, 3)},
+			want:      FalseCondition("foo", "reason true1", "message true1"),
+		},
+	}
```
hiddeco, line 222: Think that if there are multiple generations, the latest (highest) generation should always take precedence as this is the most up-to-date information available.

runtime/conditions/merge.go
darkowlzz: I had that before :smiley: but changed it to be more specific. Will make it `in64InSlice`.

runtime/conditions/merge_test.go
darkowlzz: Yeah, that'd be more correct. I'll implement that. Thanks.

runtime/conditions/merge.go
```diff
@@ -41,15 +41,16 @@ type localizedCondition struct {
 // it is should be reflected in the target condition.
 //
 // More specifically:
-// 1. Conditions are grouped by status and polarity.
+// 1. Conditions are grouped by status, polarity and observed generation(optional).
```
hiddeco, line 44:
```suggestion
// 1. Conditions are grouped by status, polarity and observed generation (optional).
```

runtime/conditions/merge.go
```diff
@@ -41,15 +41,16 @@ type localizedCondition struct {
...
 // 2. The resulting condition groups are sorted according to the following priority:
 //   - P0 - Status=True, NegativePolarity=True
 //   - P1 - Status=False, NegativePolarity=False
-//   - P2 - Condition=True, NegativePolarity=False
+//   - P2 - Status=True, NegativePolarity=False
```
hiddeco, line 48: :+1:

runtime/conditions/merge.go
```diff
@@ -41,15 +41,16 @@ type localizedCondition struct {
...
 //   - P3 - Status=False, NegativePolarity=True
 //   - P4 - Status=Unknown
 // 3. The group with highest priority is used to determine status, and other info of the target condition.
 // 4. If the polarity of the highest priority and target priority differ, it is inverted.
+// 5. If observed generation is considered, the condition groups with the latest generation get the highest priority.
```
hiddeco, line 53:
```suggestion
// 5. If the observed generation is considered, the condition groups with the latest generation get the highest priority.
```

runtime/conditions/merge.go
```diff
@@ -98,6 +99,10 @@ func getConditionGroups(conditions []localizedCondition, options *mergeOptions)
 		for i := range groups {
 			if groups[i].status == condition.Status &&
 				groups[i].negativePolarity == stringInSlice(options.negativePolarityConditionTypes, condition.Type) {
+				// If with generations is set, add to group only if the generations match.
```
hiddeco, line 102:
```suggestion
				// If withGenerations is true, add to group only if the generations match.
```

runtime/conditions/merge_strategies.go
```diff
@@ -231,3 +233,10 @@ func getFirstCondition(g conditionGroups, priority []string) *localizedCondition
 		return &topGroup.conditions[0]
 	}
 }
+
+// WithGenerations instructs merge to consider the conditions with the given observed generations only.
+func WithGenerations() MergeOption {
```
hiddeco, line 238: Wondering if something like `WithLatestGeneration()` would be more descriptive? Also, the comment seems to be outdated.

darkowlzz: Yeah, I think being explicit about it will be better.

hiddeco (approved):
LGTM, thank you @darkowlzz :100:

PR #152, opened 2021-09-30
Add server-side apply package

stefanprodan:
Package ssa contains utilities for managing Kubernetes resources using sever-side apply.

Adapted from https://github.com/stefanprodan/kustomizer/tree/v1.1.0/pkg/manager.

Used in https://github.com/fluxcd/kustomize-controller/pull/426.


ssa/changeset.go
```diff
@@ -0,0 +1,84 @@
...
+// ChangeSetEntry defines the result of an action performed on an object.
+type ChangeSetEntry struct {
+	// Subject represents the Object ID in the format 'kind/namespace/name'.
+	Subject string
+	// Action represents the action type taken by the reconciler for this object.
+	Action string
+	// Diff contains the YAML diff resulting from server-side apply dry-run.
+	Diff string
+}
+
```
hiddeco, line 81: Would provide an example here of the returned string.

ssa/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/fluxcd/pkg/ssa
+
+go 1.16
+
+require (
+	github.com/google/go-cmp v0.5.6
+	k8s.io/apimachinery v0.22.2
+	sigs.k8s.io/cli-utils v0.25.1-0.20210608181808-f3974341173a
```
hiddeco, line 8: `cli-utils` has a loose release cycle right (and therefore pinned to a commit)?

stefanprodan: Yes this needs to be pinned everywhere otherwise we can't build flux due to kustomize v2.

hiddeco: Would maybe add a comment (super sensitive about (documenting) version pins at the moment after libgit2 fiasco).

stefanprodan: Added

ssa/changeset.go
```diff
@@ -0,0 +1,84 @@
...
+*/
+
+package ssa
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Action resents the action type performed by the reconciliation process.
```
hiddeco, line 25:
```suggestion
// Action represents the action type performed by the reconciliation process.
```

ssa/changeset.go
```diff
@@ -0,0 +1,84 @@
...
+func (c *ChangeSet) String() string {
+	var b strings.Builder
+	for _, entry := range c.Entries {
+		b.WriteString(entry.String() + "\n")
+	}
+	return strings.TrimSuffix(b.String(), "\n")
+}
+
+func (c *ChangeSet) ToMap() map[string]string {
+	res := make(map[string]string)
```
hiddeco, line 65:
```suggestion
	res := make(map[string]string, len(c.Entries))
```

ssa/manager_apply.go
```diff
@@ -0,0 +1,170 @@
...
+// Apply performs a server-side apply of the given object if the matching in-cluster object is different or if it doesn't exist.
+// Drift detection is performed by comparing the server-side dry-run result with the existing object.
+// When immutable field changes are detected, the object is recreated if 'force' is set to 'true'.
+func (m *ResourceManager) Apply(ctx context.Context, object *unstructured.Unstructured, force bool) (*ChangeSetEntry, error) {
+	existingObject := object.DeepCopy()
+	_ = m.client.Get(ctx, client.ObjectKeyFromObject(object), existingObject)
+
+	dryRunObject := object.DeepCopy()
+	if err := m.dryRunApply(ctx, dryRunObject); err != nil {
+		if force && strings.Contains(err.Error(), "immutable") {
```
hiddeco, line 40: How safe is this check? Could I create a false positive by e.g. creating a resource with the name `immutable`, and then let it fail because of something not related to immutability?

ssa/manager_apply.go
```diff
@@ -0,0 +1,170 @@
...
+	sort.Sort(SortableUnstructureds(objects))
+	changeSet := NewChangeSet()
+	var toApply []*unstructured.Unstructured
+	for _, object := range objects {
+		existingObject := object.DeepCopy()
+		_ = m.client.Get(ctx, client.ObjectKeyFromObject(object), existingObject)
+
+		dryRunObject := object.DeepCopy()
+		if err := m.dryRunApply(ctx, dryRunObject); err != nil {
+			if force && strings.Contains(err.Error(), "immutable") {
```
hiddeco, line 80: Given this is used twice, I would factor it out to ensure the logic in both methods continuous to be aligned.

ssa/manager_apply.go
stefanprodan: Yes but for now there is typed error for immutable errors, I'll add a comment here to drop the string contains once Flux min req becomes Kubernetes 1.21.

hiddeco: Was not specifically looking for a typed error, but maybe an `AND` condition check which confirms another part of the common error message is present in the string after the `Contains(err, "immutable")` passes.

hiddeco (approved):
LGTM, awesome work @stefanprodan :100: :1st_place_medal:

PR #153, opened 2021-10-05
gittestserver: Add InitRepo() to initialize a repo

darkowlzz:
`GitServer.InitRepo()` can be used to initialize a repo in the git server
with a given fixture. The repo is created internally using go-git over
the file protocol, irrespective of the git server.

This is needed for https://github.com/fluxcd/source-controller/pull/445 to use the git test server.


stefanprodan (approved):
LGTM

Thanks @darkowlzz ☕

Issue #139, opened 2021-08-03
Mark events to prevent alerting spam


stefanprodan:
Taken from https://github.com/fluxcd/source-controller/pull/411

To improve observability through events, without spamming the alerting system, I propose we introduce a new `eventSeverity` type called `trace`. The `trace` events are emitted as Kubernetes events under `Normal`, these events are dropped when forwarding to notification controller. Every source reconciliation should issue **a single** `info` event that contains the source URL and revision (later on we should consider adding the commit message too for GitRepositories).

PR #161, opened 2021-10-10
External event recorder adhere to EventRecorder

### LINKED ISSUE #134: External event recorder should adhere to `EventRecorder` interface
The Kubernetes core has the following `record.EventRecorder` interface:

```go
// EventRecorder knows how to record events on behalf of an EventSource.
type EventRecorder interface {
	// Event constructs an event from the given information and puts it in the queue for sending.
	// 'object' is the object this event is about. Event will make a reference-- or you may also
	// pass a reference to the object directly.
	// 'type' of this event, and can be one of Normal, Warning. New types could be added in future
	// 'reason' is the reason this event is generated. 'reason' should be short and unique; it
	// should be in UpperCamelCase format (starting with a capital letter). "reason" will be used
	// to automate handling of events, so imagine people writing switch statements to handle them.
	// You want to make that easy.
	// 'message' is intended to be human readable.
	//
	// The resulting event will be created in the same namespace as the reference object.
	Event(object runtime.Object, eventtype, reason, message string)

	// Eventf is just like Event, but with Sprintf for the message field.
	Eventf(object runtime.Object, eventtype, reason, messageFmt string, args ...interface{})

	// AnnotatedEventf is just like eventf, but with annotations attached
	AnnotatedEventf(object runtime.Object, annotations map[string]string, eventtype, reason, messageFmt string, args ...interface{})
}
````

As this interface includes everything we need, including adding "metadata" using `AnnotatedEventf`, it would likely be better if our own `event.Recorder` would adhere to the above interface.

As a second step, the `controllers.Events` helper could be refactored into a "`DelegatingRecorder`" which delegates events to all underlying recorder implementations.

### PR DESCRIPTION
**Pull request by darkowlzz**
:warning: Removes `controller.Events`.

This change updates the external event recorder to be compatible with the k8s EventRecorder interface and embed the k8s event recorder. Since the `events.Recorder` implements the k8s EventRecorder, it can be directly used in reconcilers that embed the k8s EventRecorder interface. This makes the `controller.Events` redundant. The k8s event recorder is embedded in the `events.Recorder` to be able to post events to k8s and any external recorders.

This also introduces a new event type `EventTypeTrace` to map with the severity type trace.
Since the kubernetes API only accepts Normal and Warning event types. Sending a Trace event results in:

```console
E1018 10:37:56.133861 1768805 event.go:334] Unsupported event type: 'Trace'
```

In order to keep the trace events only in k8s, the trace events are passed to k8s as Normal events. Trace events are not passed to the external event recorders, refer https://github.com/fluxcd/pkg/issues/139.

Since the `EventRecorder` interface methods don't return any error, the external event recorder has a logger to log any event recording related errors.

Replacing `corev1.ObjectReference` with `runtime.Object` due to the function signature change required updating the existing tests objects to be `runtime.Object` compatible.

Closes #134

### DIFF
--- a/runtime/events/recorder.go
+++ b/runtime/events/recorder.go
@@ -120,6 +120,9 @@ func (r *Recorder) AnnotatedEventf(
 		r.Log.Error(err, "failed to get object reference")
 	}
 
+	// Add object info in the logger.
+	log := r.Log.WithValues("name", ref.Name, "namespace", ref.Namespace, "reconciler kind", ref.Kind)
+
 	// Convert the eventType to severity.
 	severity := eventTypeToSeverity(eventtype)
 
@@ -135,7 +138,7 @@ func (r *Recorder) AnnotatedEventf(
 
 	if r.Client == nil {
 		err := fmt.Errorf("retryable HTTP client has not been initialized")
-		r.Log.Error(err, "unable to record event")
+		log.Error(err, "unable to record event")
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


darkowlzz:
Rebased and added object information in the recorder logger.

The error log from the retry test contains the object info, name/namespace:

```console
E1019 00:36:16.462960 2396418 recorder.go:195]  "msg"="unable to record event" "error"="POST http://127.0.0.1:33683 giving up after 3 attempt(s): unexpected HTTP status 500 Internal Server Error" "name"="webapp" "namespace"="gitops-system" 
```

PR #164, opened 2021-10-11
Change the metric bucket into a more useful range.

dieend:
We encountered a delay in reconciliation which end up taking to more than 15 minutes. The current histogram metrics configuration doesn't help us to see the actual value.

This change will set the minimum range from 1nanoseconds, to 1millisecond. This is assuming 1 millisecond is blazingly fast that we don't care anything beyond that.

It'll also change the maximum range from 10s, to 1 hour. It's assuming 1 hour is terribly slow, and we don't care anything beyond that.

Ideally, this configuration should be configurable by user, as each company have their own preferences.

### DIFF
--- a/runtime/metrics/recorder.go
+++ b/runtime/metrics/recorder.go
@@ -57,10 +57,10 @@ func NewRecorder() *Recorder {
 		),
 		durationHistogram: prometheus.NewHistogramVec(
 			prometheus.HistogramOpts{
-				Name:    "gotk_reconcile_duration_seconds",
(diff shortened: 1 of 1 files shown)


dieend:
@stefanprodan this slipped through my hand. I've updated the PR, can we re-run the test?

dieend:
Hi @stefanprodan any chance you can take a look?

stefanprodan (approved):
LGTM

Thanks @dieend

stefanprodan:
@dieend can you please rebase and force push? Thanks

PR #177, opened 2021-10-28
Run build in container

relu:
Run the build inside a container


.github/actions/run-tests/Dockerfile
```diff
@@ -0,0 +1,22 @@
+ARG BASE_VARIANT=bullseye
+ARG GO_VERSION=1.16.8
```
stefanprodan, line 2: Can we not pin the patch version? In CI we do `1.16.x`

stefanprodan (approved):

PR #178, opened 2021-11-01
Rationalise the return value of ACL checks

squaremo:
The return type of HasAccessToRef() is `(bool, error)`, but the error is used to signal both a negative result (no access!) and a failure to complete the check (client request failed), leaving the `bool` redundant.

As a result, it's not clear to the caller whether it should report that access was denied, or that there was some (probably transitory) problem.

~Instead, use the bool to signal whether access was possible, and only return an error when there was an error while checking.~

EDIT: return a special type of error for access denied, and remove the bool return value.

### DIFF
--- a/runtime/acl/acl.go
+++ b/runtime/acl/acl.go
@@ -29,6 +29,24 @@ import (
 	aclapi "github.com/fluxcd/pkg/apis/acl"
 )
 
+// AccessDeniedError represents a failed access control list check.
+type AccessDeniedError string
+
+func (e AccessDeniedError) Error() string {
+	return string(e)
+}
+
+func accessDeniedErrorf(f string, args ...interface{}) error {
+	return AccessDeniedError(fmt.Sprintf(f, args...))
+}
+
+// IsAccessDenied returns true if the supplied error is an access denied error; e.g., as returned by
+// HasAccessToRef.
+func IsAccessDenied(e error) bool {
+	_, ok := e.(AccessDeniedError)
+	return ok
+}
+
 // Authorization is an ACL helper for asserting access to cross-namespace references.
 type Authorization struct {
 	client client.Client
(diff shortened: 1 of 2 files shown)


stefanprodan (commented):
The reason I put these in place is to help out Flux users after we transition to ACL. Removing the detailed errors would make it hard for end users to figure out what’s missing.

squaremo:
> The reason I put these in place is to help out Flux users after we transition to ACL.

Yeah, that is a casualty of using the error just for errors. To pass more information on why access is not possible, we could

 - return only an error and use a particular type for "access denied", so it can be distinguished from other errors
 - return an enumeration (of `OK|DeniedBecauseX|DeniedBecauseY|...`) instead of a bool

The latter is a bit cleaner, if you almost always want to distinguish between access denied and other errors (which I think we do, otherwise I wouldn't have posted this PR!).

stefanprodan:
> return only an error and use a particular type for "access denied", so it can be distinguished from other errors

I vote for this option :)

stefanprodan:
Maybe we could `errors.NewForbidden` from `k8s.io/apimachinery/pkg/errors`?

squaremo:
Thinking about it, usually you don't want to disclose the reason for an authorisation failure, since that's leaking information. In this case, it would tell a person who can read the conditions of the referring object something about the object being referred to (e.g., whether it exists or not), which they might otherwise not be able to see.

stefanprodan:
Yes i chose to leak the objects name to make ACLs user-friendly, a tenant reading the error message would know what to ask for from cluster admins. Do we consider the objects name (namespace and source) sensitive information?

stefanprodan (approved):
LGTM

Thanks @squaremo

squaremo:
> Do we consider the objects name (namespace and source) sensitive information?

No (since you have to use those to refer to it in the first place); but you might consider acknowledging its existence, or the nature of its ACL, to be giving information away.

PR #179, opened 2021-11-03
gittestserver: Add http middleware support

darkowlzz:
Add http middleware support to be able to change the behavior of the
git server. Usage could be for returning certain response and error or
introducing some delay in the server, etc.
The middlewares can be chained together based on the needs.

Helper middlewares can be added in the future for certain behaviors
frequently needed in tests.

### DIFF
--- a/gittestserver/server.go
+++ b/gittestserver/server.go
@@ -20,6 +20,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"net/http"
 	"net/http/httptest"
 	"net/url"
 	"os"
@@ -60,6 +61,9 @@ func NewGitServer(docroot string) *GitServer {
 	}
 }
 
+// HTTPMiddleware is a git http server middleware.
+type HTTPMiddleware func(http.Handler) http.Handler
+
 // GitServer is a git server for testing purposes.
 // It can serve git repositories over HTTP and SSH.
 type GitServer struct {
@@ -68,6 +72,12 @@ type GitServer struct {
 	sshServer  *gitkit.SSH
 	// Set these to configure HTTP auth
 	username, password string
+	httpMiddlewares    []HTTPMiddleware
+}
+
+// AddHTTPMiddlewares adds http middlewares to the git server.
+func (s *GitServer) AddHTTPMiddlewares(httpMiddlewares ...HTTPMiddleware) {
+	s.httpMiddlewares = append(s.httpMiddlewares, httpMiddlewares...)
 }
 
 // AutoCreate enables the automatic creation of a non-existing Git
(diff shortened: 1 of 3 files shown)


gittestserver/server.go
```diff
@@ -60,6 +61,9 @@ func NewGitServer(docroot string) *GitServer {
 	}
 }
 
+// Middleware is a git http server middleware.
+type Middleware func(http.Handler) http.Handler
```
hiddeco, line 65:
```suggestion
type HTTPMiddleware func(http.Handler) http.Handler
```

as we already implement both HTTP/S and SSH, it is better to be specific imo.

This would also need changing of the `AddMiddleware` method.

darkowlzz: Yeah, that'd be good. I was wondering if middlewares are usually used for SSH as well, but couldn't find much about it. Seems to be a thing mostly in http servers.

gittestserver/server.go
```diff
@@ -68,6 +72,12 @@ type GitServer struct {
 	sshServer  *gitkit.SSH
 	// Set these to configure HTTP auth
 	username, password string
+	httpMiddlewares    []HTTPMiddleware
+}
+
+// AddHTTPMiddlewares adds http middlewares to the git server.
+func (s *GitServer) AddHTTPMiddlewares(httpMiddlewares ...HTTPMiddleware) {
```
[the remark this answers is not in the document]
darkowlzz: You're right. Good catch. Updated to append the handlers.
Thanks.

darkowlzz:
There seems to be a test flake in `TestAclAuthorization` - https://github.com/fluxcd/pkg/runs/4097180945?check_suite_focus=true#step:6:95 .

```console
--- FAIL: TestAclAuthorization (11.40s)
    --- FAIL: TestAclAuthorization/When_in_same_namespace_it_ignores_the_ACL_and_grands_access (0.02s)
        acl_test.go:106: 
            Unexpected error:
                <*errors.withStack | 0xc0004824b0>: {
                    error: <*errors.withMessage | 0xc000360180>{
                        cause: <*cache.ErrCacheNotStarted | 0x259c668>{},
                        msg: "object /, Kind=, /test1 is not being added to the testenv client cache",
                    },
                    stack: [0x1592831, 0x15943b6, 0x522d4f, 0x474021],
                }
                object /, Kind=, /test1 is not being added to the testenv client cache: the cache is not started, can not read objects
            occurred
```

Looks like the test starts immediately without waiting for the setup to be ready, https://github.com/fluxcd/pkg/blob/main/runtime/acl/acl_test.go#L36-L41, resulting in the above failure.
Usually we add a `<-testEnv.Manager.Elected()` after starting the test environment go-routine, which hopefully waits provides enough time for cache to be ready. That may help.

hiddeco (approved):
This looks good to me, thanks @darkowlzz :bow:

PR #180, opened 2021-11-05
Remove file support from Kustomize Patch type docs

sbernheim:
The Kustomize Patch API type only supports inlined patch
definitions and does not support reading patches from a
file. Updating the godoc comments for the type to remove
any mention of reading patch definitions from a file and
make it clear that only inline patches are supported.

### DIFF
--- a/apis/kustomize/kustomize_types.go
+++ b/apis/kustomize/kustomize_types.go
@@ -83,10 +83,11 @@ type Selector struct {
 	LabelSelector string `json:"labelSelector,omitempty"`
 }
 
-// Patch contains either a StrategicMerge or a JSON6902 patch, either a file or inline, and the target the patch should
+// Patch contains an inline StrategicMerge or JSON6902 patch, and the target the patch should
 // be applied to.
 type Patch struct {
-	// Patch contains the JSON6902 patch document with an array of operation objects.
+	// Patch contains an inline StrategicMerge patch or an inline JSON6902 patch with 
+	// an array of operation objects.
 	// +required
 	Patch string `json:"patch,omitempty"`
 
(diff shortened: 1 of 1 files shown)


sbernheim:
@stefanprodan @scottrigby - I mentioned this via Slack earlier today.  These doc comments led me to believe that FluxCD Kustomization supports reading patch content from a file.  Until/unless we add support for that option, the docs should not indicate that it's supported.

scottrigby (approved):
Looks good! just need to push DCO

stefanprodan (approved):
LGTM

Thanks @sbernheim ☕

hiddeco (approved):
My guess on why this was initially documented is that I didn't reject the possibility that some implementation _would_ support file references. However, all controller implementations do not provide this option, and it is thus more confusing than that it adds value.

Thank you Herr @sbernheim, and for the trust in my "usual high standard of work" :see_no_evil:

hiddeco:
It needs a rebase, and probably the run of `make all` or something to make the CI not fall over generated doc changes.

sbernheim:
@hiddeco - I did a rebase and force push just now.  I'm not sure why `make all` would be necessary for a two-line godoc change, but I gave it a try.  Unfortunately, I appear to be missing some dependencies needed for testing the pkg project in my local setup so most goals failed.

If this rebased PR doesn't do the trick, I'll have to circle back and try again when I'm not just awake in the middle of the night after sleeping most of the day. 🤒

sbernheim:
Thank you, @scottrigby!

Multumesc, @stefanprodan!

En bedankt, Meneer @hiddeco!

PR #182, opened 2021-11-07
ssa: Add helper to detect SOPS secrets

stefanprodan:
This PR adds a helper function named `IsEncryptedSecret` to be used in kustomize-controller to detect SOPS encrypted secrets.

### DIFF
--- a/ssa/utils.go
+++ b/ssa/utils.go
@@ -206,6 +206,16 @@ func IsKustomization(object *unstructured.Unstructured) bool {
 	return false
 }
 
+// IsEncryptedSecret checks if the given object is a Kubernetes Secret encrypted with Mozilla SOPS.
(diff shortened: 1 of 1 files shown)


ssa/utils.go
```diff
@@ -206,6 +206,16 @@ func IsKustomization(object *unstructured.Unstructured) bool {
 	return false
 }
 
+// IsEncryptedSecret checks if the given object is a Kubernetes Secret encrypted with Mozilla SOPS.
+func IsEncryptedSecret(object *unstructured.Unstructured) bool {
```
hiddeco, line 210: Would not introduce this in this package, but rather keep it together with SOPS logic in kustomize-controller.

hiddeco: (Or in a generic SOPS package, but in any case, I wouldn't blend SSA with other implementation specifics).

PR #192, opened 2021-12-03
ssa: Force apply on invalid and conflict errors

stefanprodan:
Ref: https://github.com/fluxcd/kustomize-controller/issues/506


hiddeco (commented):
:100: for adding the reference.

hiddeco (approved):

stefanprodan:
I have some doubts that `kubectl apply --force` is doing the right thing when using webhooks. Assuming you have some validation controller that marks all objects as invalid due to some bug, using force will wipe clean the cluster even if there is no conflict around immutability.

PR #205, opened 2021-12-22
Export IsImmutableError, and some clean-up

souleb:
We want to rely on this function behavior in flux-cli


stefanprodan (approved):
LGTM

Thanks @souleb

PR #208, opened 2022-01-06
Preserve value difference when masking secret data

souleb:
This enable an existing secret object and its updated version to preserve changes when masking sensitive values


stefanprodan (approved):
LGTM

Thanks @souleb 🏅

PR #150, opened 2021-09-24
Fuzzing: Initial commit

AdamKorcz:
This is a draft for work in progress for adding fuzzing to the shared library.

In case anyone wants to comment, ask or test it out, I share the progress in this PR.

The fuzzers can be built and run by:
```bash
git clone https://github.com/AdamKorcz/pkg --branch fuzz1
cd pkg/fuzz
sudo docker build . --tag flux-fuzzing-image
```
To run a fuzzer, modify and uncomment this line in the Dockerfile:
```dockerfile
#RUN /fuzzers/FuzzLibGit2Error
```

____________________________________________________________
Cross referencing https://github.com/fluxcd/source-controller/pull/443

PR #212, opened 2022-01-10
Refactor Fuzz implementation

pjbgf:
Following-up from the security audit, I reviewed [the initial Fuzz PR](https://github.com/fluxcd/pkg/pull/150) and created this PR that supersedes it.

Main changes:
- Structure the fuzz implementation to be closer to what fuzz testing might look like once leveraging the [golang 1.18 built-in fuzz feature](https://go.dev/doc/fuzz/).
- Add Makefile target to enable smoke testing fuzzers locally.
- Add github workflow to test and upload crash results.
- Add documentation and fix issue with the `FuzzEventf` target.

The changes aim to make it easier to test fuzzers locally. And decrease the amount of changes to adopt built-in support for fuzzing.

Oss-fuzz aims to add support to the golang native fuzzing once [1.18 is released officially](https://github.com/google/oss-fuzz/issues/7020#issuecomment-995282338).


ssh/go.mod
```diff
@@ -3,6 +3,19 @@ module github.com/fluxcd/pkg/ssh
 go 1.17
 
 // fix for CVE-2020-29652
-require golang.org/x/crypto v0.0.0-20201216223049-8b5274cf687f
+require golang.org/x/crypto v0.0.0-20210322153248-0c34fe9e7dc2
```
stefanprodan, line 6: This brings back CVE-2020-29652

.github/workflows/cifuzz.yaml
```diff
@@ -0,0 +1,24 @@
+name: CIFuzz
+on: [pull_request]
```
hiddeco, line 2: Is this limitation on purpose, or do we also want to e.g. test branches? Also: do we want to test _all_ pull requests, or just the ones targeting specific branches?

pjbgf: The goal for this workflow is just to ensure the fuzz targets are operational, and to identify any low-hanging fuzz (within 60s of execution). 

Assuming that _all_ changes to `main` are done via PRs, I think we want to target pull requests against `main` only. I will update that now.

hiddeco (approved):
This looks good to me, thank you very much @pjbgf :bow:

runtime/events/events_fuzzer.go
```diff
@@ -0,0 +1,133 @@
...
+
+func ensureDependencies() error {
+	// only install dependencies when running inside a container
+	if _, err := os.Stat("/.dockerenv"); os.IsNotExist(err) {
+		return nil
+	}
+
+	if os.Getenv("KUBEBUILDER_ASSETS") == "" {
+		cmd := exec.Command("/usr/bin/bash", "-c", `go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest && \
+		/root/go/bin/setup-envtest use -p path 1.21.x`)
```
stefanprodan, line 54: Can we move the Kubernetes version into the Makefile?

stefanprodan:
@pjbgf would it possible to have a dedicated module for fuzz that imports all our packages and the fuzz specific ones? This would keep the runtime dependencies to a minimum.

runtime/events/events_fuzzer.go
pjbgf: @stefanprodan good point. I changed it to use the value from  `ENVTEST_BIN_VERSION`. 
The default value of `1.23` is to keep the calls from `oss-fuzz` deterministic. PTAL

pjbgf:
> @pjbgf would it be possible to have a dedicated module for fuzz that imports all our packages and the fuzz specific ones? This would keep the runtime dependencies to a minimum.

Sure, will move them all to the fuzz directory and move them back at executing time.

stefanprodan (approved):
LGTM

Thanks @pjbgf 🏅

PR #218, opened 2022-01-13
Fix fuzzer build and improve fuzz tests on pull requests

pjbgf:
Changes:
- Fix issues happening at [oss-fuzz](https://github.com/google/oss-fuzz/runs/4804502678?check_suite_focus=true) blocking https://github.com/google/oss-fuzz/pull/7131.
- Use the project specific smoke test to ensure the fuzzer correctness instead of using the official oss-fuzz GitHub action, which does not support well the multi-repository approach that FluxCD requires.


stefanprodan (approved):
LGTM

Thanks @pjbgf 🏅

PR #219, opened 2022-01-14
Add action to generate SBOM for container images

stefanprodan:
This action should be used in all the GitOps Toolkit controllers, in the release workflow, to publish SBOM to GitHub releases.

### DIFF
--- a/actions/sbom/action.yml
+++ b/actions/sbom/action.yml
@@ -1,0 +1,29 @@
+name: Generate SBOM for container images
+description: A GitHub Action for generating SBOM in spdx-json format
+author: Stefan Prodan
+branding:
+  color: blue
+  icon: command
+inputs:
+  version:
+    description: "Syft version"
+    required: false
+  image:
+    description: "container image URL"
+    required: true
+  output:
+    description: "SBOM file path"
+    required: true
+    default: "/tmp/sbom.spdx.json"
+runs:
+  using: composite
+  steps:
+    - name: "Install Anchore Syft"
+      shell: bash
+      run: |
+        curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | \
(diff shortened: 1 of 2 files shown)


actions/sbom/action.yml
```diff
@@ -0,0 +1,29 @@
+name: Generate SBOM for container images
+description: A GitHub Action for generating SBOM in spdx-json format
+author: Stefan Prodan
+branding:
+  color: blue
+  icon: command
+inputs:
+  version:
+    description: "Syft version"
+    required: false
```
hiddeco, line 10: Isn't this required input?

stefanprodan: It is not, we get the latest version by default.

stefanprodan: Pinning to a version it is better done in each workflow I think.

hiddeco: Ack, when I tried the snippet from the CLI PR, I got the impression a function argument was required. All good then :+1:

hiddeco (approved):

actions/sbom/action.yml
```diff
@@ -0,0 +1,29 @@
...
+    required: true
+    default: "/tmp/sbom.spdx.json"
+runs:
+  using: composite
+  steps:
+    - name: "Install Anchore Syft"
+      shell: bash
+      run: |
+        curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | \
+          sh -s -- -b /usr/local/bin ${{ inputs.version }}
```
pjbgf, line 25: It would be safer to pin this to a version and verify its checksum.

stefanprodan: Yeap, please add this to the `sbom-action` branch.

pjbgf (approved):
LGTM

PR #220, opened 2022-01-14
Add checksum verification for syft installer script

pjbgf:
cc @stefanprodan


stefanprodan:
@pjbgf this fails with "sha256sum: 'standard input': no properly formatted SHA256 checksum lines found"

pjbgf:
Mistyped the checksum - should be all good now.

stefanprodan (approved):
Thank you!

PR #221, opened 2022-01-17
Add tar.WithMaxUntarSize

pjbgf:
- Add new tar package to make Untar more extensible/testable.
- Set default `maxUntarSize` to `100Mb`.
- Tests core scenarios around `tar.Untar` usage.
- Removes `summary` as this is not used across Flux2.

This PR is to be follow-up by another that either deprecates the use of `untar` or redirects `untar.Untar` to the new pkg.

### DIFF
--- a/tar/tar.go
+++ b/tar/tar.go
@@ -1,0 +1,172 @@
+// Copyright 2017 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Copyright 2020 The FluxCD contributors. All rights reserved.
+// Adapted from: golang.org/x/build/internal/untar
+
+// Package tar provides ways to manage tarball files.
+package tar
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"time"
+
+	securejoin "github.com/cyphar/filepath-securejoin"
+)
+
+const (
+	// DefaultMaxUntarSize defines the default (100MB) max amount of bytes that Untar will process.
+	DefaultMaxUntarSize = 100 << (10 * 2)
+
+	// UnlimitedUntarSize defines the value which disables untar size checks for maxUntarSize.
+	UnlimitedUntarSize = -1
+)
+
+type tarOpts struct {
+	// maxUntarSize represents the limit size (bytes) for archives being decompressed by Untar.
+	// When max is a negative value the size checks are disabled.
+	maxUntarSize int
+}
+
+// Untar reads the gzip-compressed tar file from r and writes it into dir.
+//
+// If dir is a relative path, it cannot ascend from the current working dir.
+// If dir exists, it must be a directory.
+func Untar(r io.Reader, dir string, inOpts ...TarOption) (err error) {
+	opts := tarOpts{
+		maxUntarSize: DefaultMaxUntarSize,
+	}
+	opts.applyOpts(inOpts...)
+
+	dir = filepath.Clean(dir)
+	if !filepath.IsAbs(dir) {
+		cwd, err := os.Getwd()
+		if err != nil {
+			return err
+		}
+
+		dir, err = securejoin.SecureJoin(cwd, dir)
+		if err != nil {
+			return err
+		}
+	}
+
+	fi, err := os.Lstat(dir)
+	// Dir does not need to exist, as it can later be created.
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("cannot lstat '%s': %w", dir, err)
+	}
+
+	if err == nil && !fi.IsDir() {
+		return fmt.Errorf("dir '%s' must be a directory", dir)
+	}
+
+	madeDir := map[string]bool{}
+	zr, err := gzip.NewReader(r)
+	if err != nil {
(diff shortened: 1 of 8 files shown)


untar/untar.go
```diff
@@ -17,9 +17,35 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 )
 
+var (
+	maximumArchiveSize int = 1 << (10 * 3) // 1Gb
```
pjbgf, line 25: Added 1Gb to decrease the likelihood of breaking existing usage, but it feels too large for a default value.
Keen to here opinions on better values here.

tar/tar.go
```diff
@@ -0,0 +1,144 @@
...
+	"time"
+)
+
+type Tar struct {
+	maxUntarSize int
+}
+
+func NewTar() *Tar {
+	return &Tar{
+		maxUntarSize: 1 << (10 * 3), // 1Gb
```
pjbgf, line 29: Added 1Gb to decrease the likelihood of breaking existing usage, but it feels too large for a default value.
Keen to here opinions on better values here.

hiddeco (commented):
Think that all returned errors can be wrapped (`%v` -> `%w`).

tar/tar.go
hiddeco: I would set the default to `0` (no limit) to make it fully backwards compatible, and instead break it on the implementation side. Which makes this an "API addition only" change.

pjbgf: Fair point, I took a slight different approach by setting the default `maxUntarSize` to 100Mb for consumers of the new API. But consumers with automatically have no limit when using the old API. PTAL

tar/tar.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// WithMaxUntarSize sets the limit size for archives being decompressed by Untar.
+func (u *Tar) WithMaxUntarSize(max int) *Tar {
+	u.maxUntarSize = max
+	return u
+}
+
+// Untar reads the gzip-compressed tar file from r and writes it into dir.
+func (u *Tar) Untar(r io.Reader, dir string) (summary string, err error) {
```
stefanprodan, line 41: Can we add a `TarGzip` function that takes a directory and generates a tgz file? We have that in the tests and I think it would be nice to offer both functions in this new package.

tar/tar.go
```diff
@@ -0,0 +1,145 @@
...
+	madeDir := map[string]bool{}
+	defer func() {
+		td := time.Since(t0)
+		if err == nil {
+			summary = fmt.Sprintf("Extracted tarball into %s: %d files, %d dirs (%v)", dir, nFiles, len(madeDir), td)
+		} else {
+			summary = fmt.Sprintf("Error extracting tarball into %s after %d files, %d dirs, %v: %v", dir, nFiles, len(madeDir), td, err)
+		}
+	}()
+	zr, err := gzip.NewReader(r)
```
laozc, line 53: Can we have a flag to support non-compressed tarball?

tar/tar.go
pjbgf: I created [an issue](https://github.com/fluxcd/pkg/issues/323) so this can be done as a follow-up task.

tar/tar.go
pjbgf: This package is mostly aimed at the Flux project needs. Is there any specific use case you have in mind in which we would need non-compressed tarballs?

tar/tar.go
```diff
@@ -0,0 +1,178 @@
...
+			// beforehand. Thus, don't check for errors; the next
+			// write will fail with the same error.
+			dir := filepath.Dir(abs)
+			if !madeDir[dir] {
+				if err := os.MkdirAll(filepath.Dir(abs), 0o755); err != nil {
+					return summary, err
+				}
+				madeDir[dir] = true
+			}
+			wf, err := os.OpenFile(abs, os.O_RDWR|os.O_CREATE|os.O_TRUNC, mode.Perm())
```
darkowlzz, line 127: Looks like upstream has some updated around this code https://github.com/golang/build/blame/master/internal/untar/untar.go#L82-L93 . Since this is used in flux CLI which can be used on macOS, do we need the updates here too?

tar/tar.go
```diff
@@ -0,0 +1,167 @@
...
+	maxUntarSize int
+}
+
+// Untar reads the gzip-compressed tar file from r and writes it into dir.
+//
+// If dir is a relative path, it cannot ascend from the current working dir.
+// If dir exists, it must be a directory.
+func Untar(r io.Reader, dir string, inOpts ...TarOption) (err error) {
+	opts := tarOpts{
+		maxUntarSize: defaultMaxUntarSize,
```
stefanprodan, line 43: I don't think we should be enforcing a max size of 100MB, we'll break Flux for many users who have huge app repos.

pjbgf: The suggestion here is for the default within the library. For the controllers the plan was to make this flag configurable, with no limit as default.

stefanprodan: How would you pass `no limit` to this function?

hiddeco: Negative integer (or at least, that is what I proposed offline).

stefanprodan: Let's add a const for no limit.

pjbgf: Added `tar.UnlimitedUntarSize` to represent the no limit value, although any negative integer is also acceptable.

stefanprodan (approved):
LGTM 

Thanks @pjbgf

PR #222, opened 2022-01-18
runtime/events: Handle empty webhook address and log event in debug mode

darkowlzz:
- Handle empty webhook address passed to the event recorder by skipping attempts to post the event to external event endpoint.
- Log events when debug logs are enabled. This is part of simplifying logs and events by making events as the front-end/go-to option. With this, while debugging, the events are also visible in the logs.

### DIFF
--- a/runtime/events/recorder.go
+++ b/runtime/events/recorder.go
@@ -79,8 +79,10 @@ var _ kuberecorder.EventRecorder = &Recorder{}
 // given webhook. The recorder performs automatic retries for connection errors and 500-range response codes from the
 // external recorder.
 func NewRecorder(mgr ctrl.Manager, log logr.Logger, webhook, reportingController string) (*Recorder, error) {
-	if _, err := url.Parse(webhook); err != nil {
-		return nil, err
+	if webhook != "" {
+		if _, err := url.Parse(webhook); err != nil {
+			return nil, err
+		}
 	}
 
 	httpClient := retryablehttp.NewClient()
@@ -135,6 +137,12 @@ func (r *Recorder) AnnotatedEventf(
 
 	// Forward the event to the Kubernetes recorder.
 	r.EventRecorder.AnnotatedEventf(object, annotations, eventtype, reason, messageFmt, args...)
+
+	// If no webhook address is provided, skip posting to event recorder
+	// endpoint.
+	if r.Webhook == "" {
+		return
+	}
 
 	if r.Client == nil {
 		err := fmt.Errorf("retryable HTTP client has not been initialized")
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

Thanks @darkowlzz

darkowlzz:
Updated the event debug log to log the annotations as individual key value pairs instead of a nested value under "annotations".
The normal and warning events are logged as:

```json
{"level":"info","ts":"2022-01-18T22:39:21.955+0530","msg":"sync222 webapp","name":"webapp","namespace":"gitops-system","reconciler kind":"ConfigMap","test":"true","test2":"foo","test3":"bar","reason":"sync"}
{"level":"error","ts":"2022-01-18T22:39:21.955+0530","msg":"sync333 webapp","name":"webapp","namespace":"gitops-system","reconciler kind":"ConfigMap","test":"true","test2":"foo","test3":"bar","error":"sync"}
```

`"test":"true","test2":"foo","test3":"bar"` being the annotated values of the event.

stefanprodan:
@darkowlzz what happens if the an annotation name collides with an exiting key, like `level` or `msg`?

darkowlzz:
> @darkowlzz what happens if the an annotation name collides with an exiting key, like `level` or `msg`?

@stefanprodan it'll overwrite. That's a good concern. Good reason to keep annotations nested.
Tried it and looks like that results in duplicate entries for the same keys:

```json
{"level":"info","ts":"2022-01-19T23:19:08.732+0530","msg":"sync webapp","name":"webapp","namespace":"gitops-system","reconciler kind":"ConfigMap","test":"true","level":"testlevel","msg":"testmsg","reason":"sync"}
{"level":"error","ts":"2022-01-19T23:19:08.732+0530","msg":"sync webapp","name":"webapp","namespace":"gitops-system","reconciler kind":"ConfigMap","test":"true","level":"testlevel","msg":"testmsg","error":"sync"}
```
New values for `level` and `msg` are appended at the end.

May not be a good thing for json log parsers. Keeping the annotations nested would avoid such issues.

stefanprodan:
@darkowlzz better to keep the annotations nested then

darkowlzz:
Updated to have nested annotations:
```json
{"level":"info","ts":"2022-01-19T23:48:24.417+0530","msg":"sync webapp","name":"webapp","namespace":"gitops-system","reconciler kind":"ConfigMap","reason":"sync","annotations":{"level":"testlevel","msg":"testmsg","test":"true"}}
{"level":"error","ts":"2022-01-19T23:48:24.417+0530","msg":"sync webapp","name":"webapp","namespace":"gitops-system","reconciler kind":"ConfigMap","annotations":{"level":"testlevel","msg":"testmsg","test":"true"},"error":"sync"}
```
Thanks @stefanprodan

hiddeco (approved):

PR #224, opened 2022-01-19
Export the Action type in pkg/kustomize

souleb:
If implemented this will permit consuming the Action type from other
packages.


stefanprodan (approved):
LGTM

Thanks @souleb

PR #229, opened 2022-01-21
ssa: Refactor server-side apply helpers

stefanprodan:
Changes:
- use `DeepEqual` to detect drift in metadata maps
- fix typo in `FieldManager` API
- remove `ReadKubernetesObjects` function (static validation should happen before SSA)


darkowlzz (approved):
LGTM!

PR #230, opened 2022-01-25
runtime: Add flag to disable cross-namespace refs

stefanprodan:
Part of: https://github.com/fluxcd/flux2/issues/2337


darkowlzz (approved):
LGTM!

PR #233, opened 2022-02-03
ssa: Transfer the ownership of managed fields from others managers

SomtochiAma:
This pull request allows the designated manager to transfer ownership of fields from others managers to itself. Instead of replacing a 3rd party manager, we merge the fields into our manager allowing it to take ownership of fields that it previously didn't owned.

Ref:  [#fluxcd/kustomize-controller](https://github.com/fluxcd/kustomize-controller/issues/558).


stefanprodan (approved):
LGTM

Thanks @SomtochiAma 🏅

PR #235, opened 2022-02-09
convert StringData to data

souleb:
If applied, will make sure that stringData is converted to Data before
applying.


ssa/utils.go
```diff
@@ -475,3 +488,14 @@ func containsItemString(s []string, e string) bool {
...
 }
+
+func convertStringDataToData(secret *corev1.Secret) {
+	if secret.Data == nil {
+		secret.Data = make(map[string][]byte)
+	}
+
+	for k, v := range secret.StringData {
+		secret.Data[k] = []byte(v)
+		delete(secret.StringData, k)
```
stefanprodan, line 499: I would set StringData to nil after the iteration is done.

ssa/manager_diff_test.go
```diff
@@ -23,9 +23,10 @@ import (
 	"testing"
 	"time"
 
+	"sigs.k8s.io/yaml"
```
stefanprodan, line 26: Move this to section below please

PR #248, opened 2022-03-03
Update `helmtestserver` to optionally sign packaged charts

aryan9600:
Adds a new method `PackageSignedChartWithVersion` which generates a PGP keypair and uses it to sign the packaged chart, along with a minor refactor. Adds tests for `helmtestserver`.

### DIFF
--- a/helmtestserver/server.go
+++ b/helmtestserver/server.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 The Flux authors
+Copyright 2020, 2022 The Flux authors
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -17,14 +17,21 @@ limitations under the License.
 package helmtestserver
 
 import (
+	"crypto/rand"
+	"encoding/hex"
 	"os"
 	"path/filepath"
 
+	"golang.org/x/crypto/openpgp"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/repo"
 	"sigs.k8s.io/yaml"
 
 	"github.com/fluxcd/pkg/testserver"
+)
+
+const (
+	keyRingName = "TestUser"
 )
 
 // NewTempHelmServer returns a HTTP HelmServer with a newly created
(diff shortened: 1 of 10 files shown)


helmtestserver/server.go
```diff
@@ -69,9 +75,56 @@ func (s *HelmServer) PackageChart(path string) error {
...
+	return s.packageChart(path, version, publicKeyPath)
+}
+
+func (s *HelmServer) packageChart(path, version, publicKeyPath string) error {
 	pkg := action.NewPackage()
-	pkg.Destination = s.HTTPServer.Root()
+	pkg.Destination = s.Root()
 	pkg.Version = version
+	if publicKeyPath != "" {
+		secretKeyPath := fmt.Sprintf("%s/%s", s.Root(), "secret.pgp")
```
darkowlzz, line 93: Would be better to have some randomness here to avoid multiple packaging processes from stepping over one another when a single test server is shared.

helmtestserver/server.go
```diff
@@ -69,9 +75,56 @@ func (s *HelmServer) PackageChart(path string) error {
 // with the given version, to be served by the HelmServer. It returns an
 // error in case of a packaging failure.
 func (s *HelmServer) PackageChartWithVersion(path, version string) error {
+	return s.packageChart(path, version, "")
+}
+
+// PackageSignedChartWithVersion attempts to package the chart at the given path
+// with the given version and sign it using a PGP keyring, to be served by the HelmServer.
```
darkowlzz, line 82: Maybe good to mention that it's a generated keyring.

helmtestserver/server.go
```diff
@@ -17,16 +17,22 @@ limitations under the License.
...
+	"golang.org/x/crypto/openpgp"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/repo"
 	"sigs.k8s.io/yaml"
 
 	"github.com/fluxcd/pkg/testserver"
 )
 
+const (
+	KeyRingName = "TestUser"
```
darkowlzz, line 33: I don't think this needs to be public.

helmtestserver/server.go
darkowlzz: Would be also nice to talk about the `publicKeyPath`. That it write the public key of the signed package, which can be used to verify the package.

helmtestserver/server.go
darkowlzz: Maybe also consider adding a `defer` to just delete the secret after use, since it's not needed at all later.

helmtestserver/server_test.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+func TestPackageChart(t *testing.T) {
+	server, err := NewTempHelmServer()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(server.Root())
+	publicKeyPath := fmt.Sprintf("%s/%s", server.Root(), "pub.pgp")
+	packagedChartPath := fmt.Sprintf("%s/%s", server.Root(), "helmchart-0.1.0.tgz")
```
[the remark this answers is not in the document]
darkowlzz: Also, let's use `filepath.Join()` here for constructing paths.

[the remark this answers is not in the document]
aryan9600: > easy/automatic cleanup with the root.
Yes, that's the reason for creating these files within the server root. I don't see how it matters to keep these files in a separate temp dir, since they'll all be cleaned at the end of the test anyway.

helmtestserver/server.go
aryan9600: It's used twice, in `generateKeyring` and `packageChart`. Making it private 👍

helmtestserver/server.go
```diff
@@ -69,9 +76,58 @@ func (s *HelmServer) PackageChart(path string) error {
 // with the given version, to be served by the HelmServer. It returns an
 // error in case of a packaging failure.
 func (s *HelmServer) PackageChartWithVersion(path, version string) error {
+	return s.packageChart(path, version, "")
+}
+
+// PackageSignedChartWithVersion attempts to package the chart at the given path
+// with the given version and sign it using a internally generated PGP keyring, to be served
```
hiddeco, line 83:
```suggestion
// with the given version and sign it using an internally generated PGP keyring, to be served
```

helmtestserver/server_test.go
```diff
@@ -0,0 +1,66 @@
+package helmtestserver
```
hiddeco, line 1: Requires copyright header.

helmtestserver/server_test.go
```diff
@@ -0,0 +1,82 @@
+/*
+Copyright 2020 The Flux authors
```
darkowlzz, line 2:
```suggestion
Copyright 2022 The Flux authors
```

darkowlzz (approved):
LGTM!
Please update the year in the license header and rebase.

PR #253, opened 2022-03-29
runtime/conditions: Update observed gen on Set()

darkowlzz:
Conditions `Set()` should update the observed generation of the condition
to the latest object version to reflect the observation on the latest
generation of the object.


pjbgf (approved):
LGTM

hiddeco (approved):

PR #255, opened 2022-03-31
[Kustomize] Retrieve the right key to validate if substitution is disabled

souleb:
Using `kustomizev1.GroupVersion.Group` this will get the key to validate if any label or annotation have been set to disable substitution.


stefanprodan (approved):
LGTM

Thanks @souleb

PR #256, opened 2022-04-11
runtime/controller: add flags to configure exponential back-off retry

pjbgf:
Add two new flags to enable users to configure exponential back-off retry. 
The default values are set to 750ms for minimum retry time, and 15min for max.

### DIFF
--- a/runtime/controller/rate_limiter.go
+++ b/runtime/controller/rate_limiter.go
@@ -1,0 +1,62 @@
+/*
+Copyright 2022 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


hiddeco (approved):

runtime/controller/rate_limiter.go
```diff
@@ -0,0 +1,62 @@
...
+	// MaxRetryDelay represents the maximum amount of time in which a
+	// reconciliation object will wait before a retry.
+	MaxRetryDelay time.Duration
+}
+
+// BindFlags will parse the given pflag.FlagSet for the controller and
+// set the RateLimiterOptions accordingly.
+func (o *RateLimiterOptions) BindFlags(fs *pflag.FlagSet) {
+	flag.DurationVar(&o.MinRetryDelay, flagMinRetryDelay, defaultMinRetryDelay,
+		"The minimum amount of time in which a reconciliation object will wait before a retry.")
```
aryan9600, line 51:
```suggestion
		"The minimum amount of time for which a reconciliation object will wait before a retry.")
```

aryan9600 (changes requested):

runtime/controller/rate_limiter.go
stefanprodan, line 51: Maybe?
```suggestion
		"The minimum amount of time that an object reconciliation will wait before a retry.")
```

stefanprodan:
@pjbgf this was publish in `runtime/v0.14.0`

PR #257, opened 2022-04-12
Add SecretKeyReference

nstogner:
Allow users to specify Secret keys. To be used in:

- https://github.com/fluxcd/kustomize-controller/issues/613
- https://github.com/fluxcd/helm-controller/issues/458

Signed-off-by: Nick Stogner <nicholas.stogner@gmail.com>


stefanprodan (approved):
LGTM

Thanks @nstogner

apis/meta/reference_types.go
```diff
@@ -54,3 +54,15 @@ type NamespacedObjectKindReference struct {
 	// +optional
 	Namespace string `json:"namespace,omitempty"`
 }
+
+// SecretKeyReference contains enough information to locate the referenced Kubernetes Secret object in the same
+// namespace. Optionally a key can be specified.
```
hiddeco, line 59:
```suggestion
// SecretKeyReference contains enough information to locate the referenced Kubernetes Secret object in the same
// namespace. Optionally, a key can be specified.
```

It might also be good for (developer) consumers to know in what way this type differs from `SecretKeySelector`, and when to pick this type over the version from K8s core.

nstogner: Would you like to add that information to the documentation above the type? Would like make its way into user-facing docs?

hiddeco: Think this might already be the case for some of the types in this module. I am also happy with it being added to the `doc.go` file, or a dedicated doc block at the top of this document.

nstogner: Added above the type to keep it local.

stefanprodan:
@nstogner can you please squash the commits into a single one, signoff that commit, rebase with upstream `main` and force push.

hiddeco (approved):

PR #262, opened 2022-04-15
kustomize: introduce secure FS implementation

hiddeco:
This implementation functions as a drop-in replacement for Kustomize's
own `fsOnDisk`, and asserts any path it handles to be inside root.

In essence, the whole file system is now restricted like
loader.RestrictionRootOnly would, but while allowing root to differ
from the top Kustomization directory.

The main reason to put the constraint in the file system implementation
is because the current Krusty API does not allow to configure a custom
load restrictor, but does allow injecting a custom FS.

### DIFF
--- a/kustomize/filesys/fs_secure.go
+++ b/kustomize/filesys/fs_secure.go
@@ -1,0 +1,232 @@
+/*
+Copyright 2022 The Flux authors
+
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM 

Thanks @hiddeco 🥇

darkowlzz (approved):
LGTM!

PR #265, opened 2022-04-26
Remove cyclic dependency to kustomize-controller API 

stefanprodan:


kustomize/kustomize_varsub.go
```diff
@@ -60,7 +59,7 @@ func SubstituteVariables(
 		return nil, err
 	}
 
-	key := fmt.Sprintf("%s/substitute", kustomizev1.GroupVersion.Group)
+	key := "kustomize.toolkit.fluxcd.io/substitute"
```
hiddeco, line 62: I feel this should at least be moved to a `const`.

stefanprodan: Done!

hiddeco (approved):

pjbgf (approved):

PR #266, opened 2022-05-04
Add WithSSHConfig and PublicKeyLookupFunc

pjbgf:
This replaces the original gitkit fork with github.com/fluxcd/gitkit.
Having the capability of setting `ssh.ServerConfig` amongst other things,
enables the use of unit tests to reproduce scenarios that otherwise can
only be experienced in E2E tests.


stefanprodan:
I’m for making an official fork on gitkit in fluxcd org.

stefanprodan:
Fork here https://github.com/fluxcd/gitkit

gittestserver/go.mod
```diff
@@ -6,8 +6,12 @@ require (
 	github.com/go-git/go-billy/v5 v5.3.1
 	github.com/go-git/go-git/v5 v5.4.2
 	github.com/sosedoff/gitkit v0.3.0
+	golang.org/x/crypto v0.0.0-20220427172511-eb4f295cb31f
 )
 
+// Replaces gitkit with fluxcd's fork, which extends the server settings that can be configured.
+replace github.com/sosedoff/gitkit => github.com/fluxcd/gitkit v0.4.0
```
stefanprodan, line 13: I would use `fluxcd/gitkit` directly, if we manage to port all our changes to upstream then we can roll back to sosedoff.

pjbgf: Sure, I have renamed the fork and updated the reference here. PTAL

stefanprodan (approved):
LGTM 

Thanks @pjbgf

PR #268, opened 2022-05-06
ssh: Update PreferredKexAlgos based on golang.org/x/crypto/ssh 

pjbgf:
Aligns preferred algorithms with upstream golang.org/x/crypto/ssh, resulting in adding support to two new kex algorithms on PreferredKexAlgos:
- curve25519-sha256
- diffie-hellman-group14-sha256

Removes references to unused algorithms that are no longer recommended for use:
- diffie-hellman-group1-sha1
- diffie-hellman-group-exchange-sha1

This code does not seem to be called anywhere. An alternative change may be to simply removing it from the module so we don't need to update it going forwards.

### DIFF
--- a/ssh/go.mod
+++ b/ssh/go.mod
@@ -4,7 +4,7 @@ go 1.17
 
 // Fix for CVE-2020-29652: https://github.com/golang/crypto/commit/8b5274cf687fd9316b4108863654cc57385531e8
 // Fix for CVE-2021-43565: https://github.com/golang/crypto/commit/5770296d904e90f15f38f77dfc2e43fdf5efc083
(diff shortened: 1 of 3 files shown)


pjbgf:
@SomtochiAma are you aware of any place where this code is being used?

hiddeco (approved):

pjbgf:
@SomtochiAma thank you for the confirmation. It seems like this is not being used within the fluxcd org. 

I will merge this PR, as if this is being used, the changes will be beneficial. However, we probably should start looking into decommissioning this code in the future if we don't find whether it is needed/used.

Issue #269, opened 2022-05-09
runtime: Use server-side throttling


stefanprodan:
Flux should disable client-side throttling, when server-side throttling is enabled. We could adapt/make use of https://github.com/kubernetes-sigs/cli-utils/pull/584 in pkg/runtime/client.

PR #270, opened 2022-05-10
runtime: Use server-side throttling when available

aryan9600:
Checks if the PriorityAndFairness flow control filter is enabled to
disable client side throttling, leveraging server side throttling
instead, since client-go respects the `Retry-After` response header set
by the APF filter.

Fixes: #269 

### DIFF
--- a/runtime/client/client.go
+++ b/runtime/client/client.go
@@ -18,7 +18,6 @@ package client
 
 import (
 	"context"
-	"log"
 
 	"github.com/spf13/pflag"
 	"k8s.io/client-go/rest"
@@ -72,12 +71,9 @@ func (o *Options) BindFlags(fs *pflag.FlagSet) {
 func GetConfigOrDie(opts Options) *rest.Config {
 	config := ctrl.GetConfigOrDie()
 	enabled, err := flowcontrol.IsEnabled(context.Background(), config)
-	if err != nil {
-		log.Fatalf("could not check if the server has PriorityAndFairness flow control filter enabled: %s", err)
-	}
-	// A negative QPS and Burst indicates that the client should not have a rate limiter.
-	// Ref: https://github.com/kubernetes/kubernetes/blob/v1.24.0/staging/src/k8s.io/client-go/rest/config.go#L354-L364
-	if enabled {
+	if err == nil && enabled {
+		// A negative QPS and Burst indicates that the client should not have a rate limiter.
+		// Ref: https://github.com/kubernetes/kubernetes/blob/v1.24.0/staging/src/k8s.io/client-go/rest/config.go#L354-L364
 		config.QPS = -1
 		config.Burst = -1
 		return config
--- a/runtime/client/client.go
+++ b/runtime/client/client.go
@@ -17,8 +17,12 @@ limitations under the License.
 package client
 
 import (
+	"context"
+	"log"
+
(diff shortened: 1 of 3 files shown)


runtime/client/client.go
```diff
@@ -61,9 +65,23 @@ func (o *Options) BindFlags(fs *pflag.FlagSet) {
 		"The maximum burst queries-per-second of requests sent to the Kubernetes API.")
 }
 
-// GetConfigOrDie wraps ctrl.GetConfigOrDie and sets the configured Options, returning the modified rest.Config.
+// GetConfigOrDie wraps ctrl.GetConfigOrDie and checks if the Kubernetes apiserver
+// has PriorityAndFairness flow control filter enabled. If true, it returns a rest.Config
+// with client side throttling disabled. Otherwise, it returns a modified rest.Config
+// configured with the provided Options.
 func GetConfigOrDie(opts Options) *rest.Config {
```
hiddeco, line 65: While working on https://github.com/fluxcd/helm-controller/pull/480, I discovered this func does not have a non-panicing counterpart.

I think this would be useful to have in situations where you work with the option, but not the configuration (like the above PR).

runtime/client/client.go
```diff
@@ -61,9 +65,23 @@ func (o *Options) BindFlags(fs *pflag.FlagSet) {
...
 	config := ctrl.GetConfigOrDie()
+	enabled, err := flowcontrol.IsEnabled(context.Background(), config)
+	if err != nil {
+		log.Fatalf("could not check if the server has PriorityAndFairness flow control filter enabled: %s", err)
```
darkowlzz, line 76: Is this intentionally a fatal log? If we can't determine flow control, shouldn't it continue with the existing behavior instead of ending the whole program?

aryan9600: Yes it is. The function name indicates that it forces the program to exit if it isn't successful. As @hiddeco said in his above comment, it'd be wise to have a counterpart which errors out instead and have this logic there. Do you think the flowcontrol check should only be in the other function or does it make sense to have it in this function as well in it's current form?

darkowlzz: I don't think we should be introducing such fatal code when we have the control to handle it gracefully. In this case, as this is a new feature, if this is requested against an old API server which doesn't support this API, it'll receive an error and that'll cause the whole program to stop. Keeping with the current behavior, if there's no flow control support, just continue with the client side configurations.

darkowlzz: > Otherwise, it returns a modified rest.Config configured with the provided Options.

It also doesn't match with the description of the function.

stefanprodan:
@aryan9600 please test this code with the [Flux RBAC](https://github.com/fluxcd/flux2/blob/main/manifests/rbac/controller.yaml) and see if we need to allow any new endpoints. Looking at the flowcontrol code seems that access to the `version` API is required.

aryan9600:
I don't think that's the case, since only the `hostURL` is returned and the api versioned path is dropped [here](https://github.com/kubernetes-sigs/cli-utils/blob/v0.30.0/pkg/flowcontrol/flowcontrol.go#L105). I'll check once anyway to be sure.

stefanprodan (approved):
LGTM

Thanks @aryan9600

stefanprodan:
@aryan9600 please rebase

PR #271, opened 2022-05-11
runtime: Add Feature Gates

pjbgf:
Adds `features.FeatureGates` that acts as an helper to managing feature switches.

Controllers can set their supported features and then at runtime verify which ones are enabled/disabled.
Callers have to call `BindFlags()`, and then call `SupportedFeatures()` to set the supported features and their default values.

Example usage for source-controller:

#### `main.go`
```golang
import (
	feathelper "github.com/fluxcd/pkg/runtime/features"
	"github.com/fluxcd/source-controller/internal/features"
)
	...
        var featureGates feathelper.FeatureGates
	...
	featureGates.BindFlags(flag.CommandLine)

	flag.Parse()

	err := featureGates.WithLogger(setupLog).
		SupportedFeatures(features.FeatureGates())
	if err != nil {
		setupLog.Error(err, "unable to load feature gates")
		os.Exit(1)
	}

```

#### `source-controller/internal/features/features.go`

```golang
// Package features sets the feature gates that
// source-controller supports, and their default
// states.
package features

import feathelper "github.com/fluxcd/pkg/runtime/features"

const (
	// OptimizedGitClones decreases resource utilization for GitRepository
	// reconciliations. It supports both go-git and libgit2 implementation
	// when checking out target branches or tags.
	//
	// When enabled, avoids full clone operations by first checking whether
	// the last revision is still the same at the target repository,
	// and if that is so, skips the reconciliation.
	OptimizedGitClones = "optimized-git-clones"
)

var features = map[string]bool{
	// OptimizedGitClones
	// opt-in from v0.25
	OptimizedGitClones: false,
}

// DefaultFeatureGates contains a list of all supported feature gates and
// their default values.
func FeatureGates() map[string]bool {
	return features
}

// Enabled verifies whether the feature is enabled or not.
func Enabled(feature string) (bool, error) {
	return feathelper.Enabled(feature)
}

```

#### `Checking`

```golang
	enabled, err := features.Enabled(features.OptimizedGitClones)
	...
```

### DIFF
--- a/runtime/features/features.go
+++ b/runtime/features/features.go
@@ -1,0 +1,85 @@
+/*
+Copyright 2022 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package features
+
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
+	"github.com/spf13/pflag"
+	cliflag "k8s.io/component-base/cli/flag"
+)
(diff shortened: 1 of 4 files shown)


stefanprodan (approved):
LGTM

Thanks @pjbgf 🏅

PR #286, opened 2022-06-07
Update `KnownKey.Matches` to initialize it's own hasher

aryan9600:
Previously, `KnownKey.Matches()` accepted a SHA256 hasher as an argument,
which could lead to unintended bugs when calling it in a loop. This
eliminates that, by intializing a new hasher itself instead of relying
on the caller for the same.
Enables us to fix a regression in the source-controller: https://github.com/fluxcd/image-automation-controller/issues/378


pjbgf (approved):
LGTM

thanks @aryan9600!

PR #289, opened 2022-06-14
Add pkg for masking token

SomtochiAma:
This pull requests adds a pkg for masking a token in the string.
The code was moved from [the notification-controller](https://github.com/fluxcd/notification-controller/blob/main/internal/server/event_handlers.go#L322) to be used by other controllers.

### DIFF
--- a/masktoken/masktoken.go
+++ b/masktoken/masktoken.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2022 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


stefanprodan:
Please rebase your branch with upstream main

pjbgf (approved):
Thanks @SomtochiAma 

LGTM

masktoken/masktoken.go
```diff
@@ -0,0 +1,38 @@
...
+*/
+
+package masktoken
+
+import (
+	"fmt"
+	"regexp"
+)
+
+func MaskTokenFromString(log string, token string) string {
```
hiddeco, line 24: I feel a bit uncomfortable about this low-level package not returning a proper error. I would think it is up to the caller to decide what needs to happen if an error happens during the masking, instead of writing the error into the output (which does not give you anything valuable programmatically to rely upon, especially since the string is not exported).

In addition to this, it is not highlighted in the doc block as behavior either.

SomtochiAma: You make a good point. In the notification controller, it was a default that the new error was returned. I can see how that can be different for other packages.
I will make an update and add a proper doc block.

PR #290, opened 2022-06-14
maskToken: Return regex compile error 

SomtochiAma:
Ref: https://github.com/fluxcd/pkg/pull/289#discussion_r896826660

### DIFF
--- a/masktoken/masktoken.go
+++ b/masktoken/masktoken.go
@@ -21,18 +21,21 @@ import (
 	"regexp"
 )
 
-func MaskTokenFromString(log string, token string) string {
+// MaskTokenFromString redacts all matches for the given token from the provided string,
+// replacing them with "*****".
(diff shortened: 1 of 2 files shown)


masktoken/masktoken.go
```diff
@@ -21,18 +21,19 @@ import (
 	"regexp"
 )
 
-func MaskTokenFromString(log string, token string) string {
+// MaskTokenFromString redacts a given token from a string. Precisely it replaces it with "*****".
+// It can be used to remove sensitive information from error messages.
```
hiddeco, line 25:
```suggestion
// MaskTokenFromString redacts all matches for the given token from the provided string,
// replacing them with  "*****".
// The token is expected to be [explain details about format/regexp behavior].
// This can for example be used to remove sensitive information from error messages.
```

PR #291, opened 2022-06-14
Consolidate scattered Git packages into `pkg/git`

aryan9600:
Adds a new a package: `git`, which defines an interface `GitClient` for all git clients for standardized usage. Currently, only a client for `go-git` has been implemented. A follow up PR shall implement a client for `libgit2` (since this PR is already quite big).
It also adds e2e test using https://github.com/fluxcd/gitkit as a git provider. More e2e tests using GitLab and BitBucket shall follow in upcoming PRs


git/git.go
```diff
@@ -0,0 +1,145 @@
...
+
+	"github.com/ProtonMail/go-crypto/openpgp"
+)
+
+type Hash []byte
+
+// String returns the SHA1 Hash as a string.
+func (h Hash) String() string {
+	return string(h)
+}
```
hiddeco, line 36: With an eye on SHA2 support being available in `git`. I think it would be good to at least prepare this for the future by e.g. making the `Hash` a separate generic type, and the current `Hash []byte` a `SHA1 [20]byte`.

Or as other option: to remove the indications of it being strictly a SHA1 hash.

git/git.go
```diff
@@ -0,0 +1,145 @@
...
+	r := []rune(subject)
+	if len(r) > 50 {
+		return fmt.Sprintf("%s...", string(r[0:50]))
+	}
+	return subject
+}
+
+// GitClient defines all the methods that a git implementation must cover.
+// An instance of GitClient should have a 1:1 mapping with a Git repository.
+type GitClient interface {
```
hiddeco, line 111: Wondering if it would be an idea to split this interface into multiple categories (e.g. read/write).

git/git.go
```diff
@@ -0,0 +1,145 @@
...
+	Init(url, branch string) (bool, error)
+	Checkout(ctx context.Context, url string, checkoutOpts CheckoutOptions) (*Commit, error)
```
aryan9600, line 113: Does it make more sense to name this `Clone()` instead and rename `SwitchBranch()` to `CheckoutBranch()`?

git/git.go
pjbgf: +1 making this more composable would make it easier to test on the different use cases.

git/git.go
```diff
@@ -0,0 +1,145 @@
...
+	Push(ctx context.Context, url string) (*Commit, error)
+	SwitchBranch(ctx context.Context, branch string)
+	Write(path string, reader io.Reader) error
+	Commit(commitInfo Commit, ent *openpgp.Entity) (string, error)
```
hiddeco, line 117:
```suggestion
	Commit(info Commit, signer *openpgp.Entity) (string, error)
```

git/git.go
aryan9600: ATM, the `GoGitClient` is stateful, i.e. it stores repository and related metadata. Seeing how projects like IAC and flux2 perform both read and write operations to the same repository, it might be tricky to design an API which splits this interface, since both would require access to the repository. In that case, we might need to make the clients stateless. @pjbgf @hiddeco

git/git.go
pjbgf: It is hard to say what this would look like as the new hash function support seems to come as part of Git protocol V3.

The goal number 3 of the [hash function transition](https://git-scm.com/docs/hash-function-transition/) states that:
Users can use SHA-1 and SHA-256 identifiers for objects interchangeably.

Suggesting that an object could be referred by both, meaning that we could end up having both.

On top of that, the document suggests the use of SHA3-256 instead of SHA-256 (from the SHA2 family). Both of which would result in the same output length, but from the same input they yield different outputs.

All the above, makes me think that the second option - removing indications of it being stricly SHA1 - may be a first good step.

git/git.go
pjbgf: @aryan9600 that is a good point, specially considering that some use cases do not require that statefulness - i.e. SC and I am assuming flux CLI as well.

By segregating this interface into smaller interfaces, we don't force the API consumers to implement everything when they only need a `Clone` operation, for example.

Another point is how we are planning to manage the release of lower level state that might need to be kept depending on the lower-level implementation (e.g. libgit2)? It would be great if that could be encapsulated, but I am not sure we can without making the implementation a lot less inefficient (i.e. creating a new repository instance for every single interface call).

git/git.go
```diff
@@ -0,0 +1,146 @@
...
+
+	for _, r := range keyRing {
+		reader := strings.NewReader(r)
+		keyring, err := openpgp.ReadArmoredKeyRing(reader)
+		if err != nil {
+			return "", fmt.Errorf("unable to read armored key ring: %w", err)
+		}
+		signer, err := openpgp.CheckArmoredDetachedSignature(keyring, bytes.NewBuffer(c.Encoded), bytes.NewBufferString(c.Signature), nil)
+		if err == nil {
+			return fmt.Sprintf("%X", signer.PrimaryKey.Fingerprint[12:20]), nil
```
pjbgf, line 93: It would be good to validate that we don't go out of bounds on the Fingerprint slice.

git/git.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+	}
+	return "", fmt.Errorf("unable to verify commit with any of the given key rings")
+}
+
+// ShortMessage returns the first 50 characters of a commit subject.
+func (c *Commit) ShortMessage() string {
+	subject := strings.Split(c.Message, "\n")[0]
+	r := []rune(subject)
+	if len(r) > 50 {
```
pjbgf, line 103:
```suggestion
	if len(r) >= 50 {
```

hiddeco: The limit is _up to_ 50, so the current condition is correct.

git/git.go
```diff
@@ -0,0 +1,167 @@
...
+type GitClient interface {
+	// Init initializes a repository at the configured path with the origin
+	// remote set to url on the provided branch.
+	Init(ctx context.Context, url, branch string) error
+	// Clone clones a repository from the provided url using the options provided.
+	Clone(ctx context.Context, url string, checkoutOpts CheckoutOptions) (*Commit, error)
+	// Push pushes the current branch of the repository to origin.
+	Push(ctx context.Context) error
+	// SwitchBranch switches the active branch of the repository to the
+	// provided branch.
```
pjbgf, line 120: The `go-git` implementation creates a new branch if one does not exist. Please expand the documentation to that effect.

git/git_test.go
```diff
@@ -0,0 +1,304 @@
+/*
+Copyright 2021 The Flux authors
```
pjbgf, line 2: Across the new files:

```suggestion
Copyright 2022 The Flux authors
```

git/gogit/git_client.go
```diff
@@ -0,0 +1,296 @@
...
+
+// NewGoGitClient returns a new GoGitClient
+func NewGoGitClient(path string, authOpts *git.AuthOptions) *GoGitClient {
+	return &GoGitClient{
+		path:     path,
+		authOpts: authOpts,
+	}
+}
+
+func (g *GoGitClient) Init(ctx context.Context, url, branch string) error {
```
pjbgf, line 53: Do we need `ctx` for init?

git/gogit/git_client.go
```diff
@@ -0,0 +1,296 @@
...
+		URLs: []string{url},
+	}); err != nil {
+		return err
+	}
+
+	branchRef := plumbing.NewBranchReferenceName(branch)
+	if err = r.CreateBranch(&config.Branch{
+		Name:   branch,
+		Remote: extgogit.DefaultRemoteName,
+		Merge:  branchRef,
```
pjbgf, line 74: The branch names are less relevant on the local repository (which is mostly ephemeral), than on the remote repository. But I suspect we want to be using the same branch name in both.

git/gogit/transport.go
```diff
@@ -0,0 +1,110 @@
...
+				if err != nil {
+					return nil, err
+				}
+				pk.HostKeyCallback = callback
+			}
+			customPK := &CustomPublicKeys{
+				pk: pk,
+			}
+			return customPK, nil
+		}
```
pjbgf, line 65: `git.SSH` requires `Identity` and `KnownHosts`, we could return an error to that effect here.

git/options.go
```diff
@@ -0,0 +1,156 @@
...
+	// SemVer tag expression to checkout, takes precedence over Tag.
+	SemVer string `json:"semver,omitempty"`
+
+	// Commit SHA1 to checkout, takes precedence over Tag and SemVer,
+	// can be combined with Branch with some Implementations.
+	Commit string
+
+	// RecurseSubmodules defines if submodules should be checked out,
+	// not supported by all Implementations.
+	RecurseSubmodules bool
```
pjbgf, line 48: `RecurseSubmodules` and `ShallowClone` could be extracted into a implementation dependent state machine field - not necessarily as part of this PR.

git/gogit/git_client.go
aryan9600: We would need it for the libgit2 client.

git/gogit/transport.go
aryan9600: this check is actually redundant, since we validate `opts` in `NewAuthOptions` before returning it, so I am removing it

git/go.mod
```diff
@@ -0,0 +1,38 @@
+module github.com/fluxcd/pkg/git
+
+go 1.17
+
+require (
+	github.com/Masterminds/semver/v3 v3.1.1
+	github.com/ProtonMail/go-crypto v0.0.0-20220517143526-88bb52951d5b
+	github.com/fluxcd/gitkit v0.5.1
+	github.com/fluxcd/pkg/gittestserver v0.5.4
```
stefanprodan, line 9: All `fluxcd/pkg` packages must be referenced on disk, see runtime for an example.

git/git.go
```diff
@@ -0,0 +1,181 @@
...
+	Clone(ctx context.Context, url string, checkoutOpts CheckoutOptions) (*Commit, error)
+	// IsClean returns whether the working tree is clean.
+	IsClean() (bool, error)
+	// Head returns the hash of the current HEAD of the repo.
+	Head() (string, error)
+	// Path returns the path of the repository.
+	Path() string
+	// Free frees any resources that need to be freed at the end of
+	// the client's lifecycle.
+	Free()
```
hiddeco, line 121: I would rename this to something like `Cleanup`, which is more Go idiomatic naming instead of the C-like "free".

pjbgf (commented):
It just needs a rebase, apart from that LGTM.

pjbgf (approved):

hiddeco:
Is the observation that with this implementation, it is no longer possible to configure e.g. an in-memory FS for go-git, correct?

git/gogit/git_client.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	var r *extgogit.Repository
+	var err error
+	if g.useMemFS {
+		storage := filesystem.NewStorage(osfs.New(g.path), cache.NewObjectLRUDefault())
+		r, err = extgogit.Init(storage, memfs.New())
+	} else {
+		r, err = extgogit.PlainInit(g.path, false)
+	}
```
pjbgf, line 70: It would be great if instead of making this decisions here, we relied on the `Storer` interface and used the same implementation regardless of whether the operation was against a filesystem or memory.

git/gogit/git_client.go
```diff
@@ -0,0 +1,291 @@
...
+var _ git.GitClient = &GoGitClient{}
+
+// NewGoGitClient returns a new GoGitClient
+func NewGoGitClient(path string, authOpts *git.AuthOptions, useMemFS bool) *GoGitClient {
+	return &GoGitClient{
+		path:     path,
+		authOpts: authOpts,
+		useMemFS: useMemFS,
+	}
+}
```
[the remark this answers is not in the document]
hiddeco: Agree with this, as this would allow you to mock some state and then inject it into the client for testing purposes.

git/git.go
```diff
@@ -0,0 +1,181 @@
...
+	Head() (string, error)
+	// Path returns the path of the repository.
+	Path() string
+	// Cleanup frees any resources that need to be freed at the end of
+	// the client's lifecycle.
+	Cleanup()
+}
+
+// GitWriter knows how to perform write operations to a repository.
+type GitWriter interface {
```
hiddeco, line 125:
```suggestion
// RepositoryWriter knows how to perform local and remote write operations to a
// Git repository.
type RepositoryWriter interface {
```

git/git.go
```diff
@@ -0,0 +1,181 @@
...
+	subject := strings.Split(c.Message, "\n")[0]
+	r := []rune(subject)
+	if len(r) > 50 {
+		return fmt.Sprintf("%s...", string(r[0:50]))
+	}
+	return subject
+}
+
+// GitReader knows how to perform read operations for a repository.
+type GitReader interface {
```
hiddeco, line 110:
```suggestion
// RepositoryReader knows how to perform local and remote read operations
// on a Git repository.
type RepositoryReader interface {
```

git/git.go
```diff
@@ -0,0 +1,181 @@
...
+	// the client's lifecycle.
+	Cleanup()
+}
+
+// GitClient defines all the methods that a git client must cover.
+// An instance of GitClient should have a 1:1 mapping with a Git repository.
+type GitClient interface {
+	GitReader
+	GitWriter
+}
```
hiddeco, line 150:
```suggestion
type Client interface {
	RepositoryReader
	RepositoryWriter
}
```

git/git.go
```diff
@@ -0,0 +1,181 @@
...
+	// are overwritten.
+	WriteFile(path string, reader io.Reader) error
+	// Commit commits any changes made to the repository.
+	Commit(info Commit, signer *openpgp.Entity) (string, error)
+	// Cleanup frees any resources that need to be freed at the end of
+	// the client's lifecycle.
+	Cleanup()
+}
+
+// GitClient defines all the methods that a git client must cover.
```
hiddeco, line 145:
```suggestion
// Client defines all the methods that a Git client must cover.
```

git/git.go
```diff
@@ -0,0 +1,181 @@
...
+// An instance of GitClient should have a 1:1 mapping with a Git repository.
```
hiddeco, line 146: What does "1:1 mapping with a Git repository" refer to?

hiddeco (commented):
Some naming nits and questions around phrasing in doc blocks. Other than that, it looks good to me if you add the default behavior (or make the `GoGitClient` over-writable with options, might be a nicer API due to not having to pass `nil, nil`) like we discussed offline. 🥇 

💡 It is generally advised to not prefix things with the same name as the package name, as this would be a double definition of context.

git/git.go
hiddeco: or even `RepositoryClient`, or `Repository`.

pjbgf: +1 on `RepositoryReader`/`RepositoryWriter`. As for the `GitClient` interface, I think `RepositoryClient` and `Repository` are better alternatives.

hiddeco: Think I am most in favor of `Repository`, as I am not sure the `Client` suffix adds that much value with the reader/writer split.

aryan9600: I think the `Client` suffix does add value, because the term `Repository` could hamper readability by misguiding users into thinking this represents a Git repository, where as this is something which can perform operations to a repository.

hiddeco: Ack.

git/git.go
darkowlzz: This change made me look into the go-crypto package because these numbers aren't easy to understand without any domain knowledge. I found that there's [KeyIdString()](https://github.com/ProtonMail/go-crypto/blob/e85cedf506cd13da5f975b5949c6fe608d34d91f/openpgp/packet/public_key.go#L783-L785), which does exactly the same thing. So, we don't really need to add the above check if we just use that library's helper.

git/go.mod
```diff
@@ -0,0 +1,45 @@
...
+replace (
+	github.com/fluxcd/pkg/gittestserver => ../gittestserver
+	github.com/fluxcd/pkg/gitutil => ../gitutil
+	github.com/fluxcd/pkg/ssh => ../ssh
+	github.com/fluxcd/pkg/version => ../version
+)
+
+require (
+	github.com/Masterminds/semver/v3 v3.1.1
+	github.com/ProtonMail/go-crypto v0.0.0-20220517143526-88bb52951d5b
```
darkowlzz, line 14: I suspect that these comments https://github.com/fluxcd/source-controller/blob/0bf9de319dded36d8723fbc492b9fcae35f9c80c/go.mod#L13-L16 are relevant here as well.

git/git.go
darkowlzz: I may be missing some context of previous discussions but `Client`, `Reader`, `Writer` sound more like how we name things in go, unless there's a possibility of other APIs here that are not about repository. But maybe in that case, these APIs can be a subpackage of git, `git/repository`.

git/git.go
```diff
@@ -0,0 +1,183 @@
...
+	// Cleanup frees any resources that need to be freed at the end of
+	// the client's lifecycle.
+	Cleanup()
+}
+
+// RepositoryWriter knows how to perform local and remote write operations
+// to a Git repository.
+type RepositoryWriter interface {
+	// Init initializes a repository at the configured path with the origin
+	// remote set to url on the provided branch.
```
darkowlzz, line 129:
```suggestion
	// Init initializes a repository at the configured path with the remote
	// origin set to url on the provided branch.
```
remote of name "origin".

git/git.go
```diff
@@ -0,0 +1,183 @@
...
+	Init(ctx context.Context, url, branch string) error
+	// Push pushes the current branch of the repository to origin.
+	Push(ctx context.Context) error
```
darkowlzz, line 132: Do we never deal with other remotes? Only origin?

git/git.go
```diff
@@ -0,0 +1,183 @@
...
+	// SwitchBranch switches the active branch of the repository to the
```
darkowlzz, line 133: maybe
```suggestion
	// SwitchBranch switches the current branch of the repository to the
```
because active and stale branches mean something else, for example in the github branches page https://github.com/fluxcd/pkg/branches.

git/git.go
```diff
@@ -0,0 +1,183 @@
...
+	// provided branch. If the branch doesn't exist, it is created.
+	SwitchBranch(ctx context.Context, branch string) error
+	// WriteFile creates a new file at the given path, reads from the reader and
+	// writes the content to the file. If the file already exists, it's contents
```
darkowlzz, line 137:
```suggestion
	// writes the content to the file. If the file already exists, its contents
```
or just
```suggestion
	// writes the content to the file. If the file already exists, it's overwritten
```

git/git.go
```diff
@@ -0,0 +1,183 @@
...
+	// are overwritten.
+	WriteFile(path string, reader io.Reader) error
+	// Commit commits any changes made to the repository.
+	Commit(info Commit, signer *openpgp.Entity) (string, error)
```
darkowlzz, line 141: Shouldn't signer be an optional parameter?
Maybe we can have options for some other methods above as well. For things that we assume defaults, like "origin" being the default remote when pushing, "main" being the default branch when initializing, bare repo option etc.

git/git.go
```diff
@@ -0,0 +1,183 @@
...
+// a Git repository.
+type RepositoryClient interface {
+	RepositoryReader
+	RepositoryWriter
+}
+
+const (
+	// GoGitClient for performing Git operations using go-git.
+	GoGitClient = "go-git"
+)
```
darkowlzz, line 157: How about moving the above interfaces and constants into a separate file, maybe `client.go`?
At the moment, there's a mix of concrete type implementations, interfaces, variables and functions all together.
Just an observation, maybe due to unfamiliar with the file.

git/gogit/repository_client.go
```diff
@@ -0,0 +1,318 @@
...
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/storage"
+	"github.com/go-git/go-git/v5/storage/filesystem"
+	"github.com/go-git/go-git/v5/storage/memory"
+
+	"github.com/fluxcd/pkg/git"
+)
+
+// GoGitClient implements git.GitClient
+type GoGitClient struct {
```
darkowlzz, line 42: Just
```suggestion
type Client struct {
```
would match with the rest of the things like `ClientOption`.

git/git.go
```diff
@@ -0,0 +1,183 @@
...
+	Clone(ctx context.Context, url string, checkoutOpts CheckoutOptions) (*Commit, error)
+	// IsClean returns whether the working tree is clean.
+	IsClean() (bool, error)
+	// Head returns the hash of the current HEAD of the repo.
+	Head() (string, error)
+	// Path returns the path of the repository.
+	Path() string
+	// Cleanup frees any resources that need to be freed at the end of
+	// the client's lifecycle.
+	Cleanup()
```
darkowlzz, line 122: Maybe `Free()` would be more accurate/better? The description also uses "free" to describe it.
Usually we think of freeing resources. Cleanup sounds more like deleting some files.

git/git.go
darkowlzz: Also makes me wonder, don't we need a way to add more other remotes? Maybe in IAC for pushing to a fork?

git/e2e/gitkit_test.go
```diff
@@ -0,0 +1,290 @@
...
+	"github.com/fluxcd/pkg/ssh"
+)
+
+var letterRunes = []rune("abcdefghijklmnopqrstuvwxyz1234567890")
+
+func TestGitKitE2E(t *testing.T) {
+	gitServer, err := gittestserver.NewTempGitServer()
+	if err != nil {
+		log.Fatal(err)
+	}
```
darkowlzz, line 51: Maybe use gomega here and below for all error checks.

git/git.go
aryan9600: It was `Free()` initially, but changed it after @hiddeco's suggestion: https://github.com/fluxcd/pkg/pull/291#discussion_r911678770

git/git.go
aryan9600: As far as I understand, we don't.

git/git.go
aryan9600: I can't think about when the IAC would need to push to a fork?

git/git.go
aryan9600: `signer` _is_ an optional parameter, I'll update the method docs to reflect this.

aryan9600: You can define the default branch of the repo through `Init(url, branch)`. I don't see why we need to support bare repos as bare repos are primarily meant for git servers and not for users.

git/git.go
hiddeco: Based on this I had another discussion with Sunny. We feel with the recent changes to the interface names, etc. The following might be better:

```go
type Closer interface{
	Close()
}
```

```go
type DiscardCloser struct{}

func (c *DiscardCloser) Close() {}
```

```go
type Client struct {
	*DiscardCloser
        ...
}

File: git/gogit/repository_client.go [hunk 32]
```diff
@@ -0,0 +1,318 @@
...
+
+	for _, clientOpt := range clientOpts {
+		if err := clientOpt(g); err != nil {
+			return nil, err
+		}
+	}
+
+	return g, nil
+}
+
```
hiddeco, line 73:
```suggestion
func WithStorer(s storage.Storer) ClientOption {
	return func(c *GoGitClient) error {
		c.storer = s
	}
}

func WithWorkTreeFS(wt billy.Filesystem) ClientOption {
	return func(c *GoGitClient) error {
		c.worktreeFS = wt
	}
}
```

---

Or the alternative:

```go
type ClientOption interface {
	apply(c *GoGitClient) error
}

type WithStorer storage.Storer

func(o WithStorer) apply(c *GoGitClient) error {
	c.storer = storage.Storer(o)
}
```

pjbgf: I would +1 on the `WithStorer` and `WithWorkTreeFS` approach.

hiddeco: I would still keep the other methods for consumer convenience, and in addition, ensure that at least both are set at the end of the init. As it's now possible to pass an option that does not set both.

git/client.go
```diff
@@ -0,0 +1,87 @@
...
+
+// DiscardRepositoryCloser is a RepositoryCloser which discards calls to Close().
+type DiscardRepositoryCloser struct{}
+
+func (c *DiscardRepositoryCloser) Close() {}
+
+const (
+	// GoGitClient represents a RepositoryClient that uses go-git for
+	// performing Git operations.
+	GoGitClient = "go-git"
```
hiddeco, line 86: I would expect this to be declared in `gogit`, as e.g. `gogit.ClientName`.

git/gogit/repository_client.go
```diff
@@ -0,0 +1,344 @@
...
+	"github.com/go-git/go-git/v5/storage/filesystem"
+	"github.com/go-git/go-git/v5/storage/memory"
+
+	"github.com/fluxcd/pkg/git"
+)
+
+const (
+	// GoGitClient represents a RepositoryClient that uses go-git for
+	// performing Git operations.
+	GoGitClient = "go-git"
```
hiddeco, line 45:
```suggestion
	ClientName = "go-git"
```

I feel `gogit.ClientName` would prevent the double context definition, and is also a more natural name for a string value. Where as `gogit.GoGitClient` could also be client object.
`gogit.GoGitClient`

git/client.go
```diff
@@ -0,0 +1,81 @@
...
+// at the end of the lifecycle of a RepositoryWriter/RepositoryReader.
+// When this is not required by the implementation, it can simply embed an
+// anonymous pointer to DiscardRepositoryCloser.
+type RepositoryCloser interface {
+	// Close closes any resources that need to be closed at the end of
+	// a Git repository client's lifecycle.
+	Close()
+}
+
+// RepositoryClient knows how to perform local and remote operations to
```
hiddeco, line 71:
```suggestion
// RepositoryClient knows how to perform local and remote operations on
```

hiddeco (approved):
Other than the minor doc nit, I am really happy with this. Great job @aryan9600 :1st_place_medal:

pjbgf (approved):
Grand job @aryan9600! :clap: :clap: 


Only a minor point missing, as pointed out by @darkowlzz we should replace `signer.PrimaryKey.Fingerprint[12:20]` with [KeyIdString()](https://github.com/ProtonMail/go-crypto/blob/e85cedf506cd13da5f975b5949c6fe608d34d91f/openpgp/packet/public_key.go#L783-L785).

PR #292, opened 2022-06-16
add new pkg `transport` for http transports

aryan9600:
This PR moves `github.com/fluxcd/source-controller/internal/transport` into `pkg/http/transport` to enable wider usage.

### DIFF
--- a/http/transport/go.mod
+++ b/http/transport/go.mod
@@ -1,0 +1,3 @@
+module github.com/fluxcd/pkg/http/transport
+
+go 1.17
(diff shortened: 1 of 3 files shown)


transport/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/fluxcd/pkg/transport
+
+go 1.18
```
pjbgf, line 3: We may want to keep the same golang version across all submodules, and upgrade them at the same time.

```suggestion
go 1.17
```

stefanprodan (approved):
LGTM

Thanks @aryan9600

PR #299, opened 2022-07-20
gittestserver: Add ReadOnly mode

pjbgf:


hiddeco (dismissed):
Something appears to be off in CI.

hiddeco (approved):

PR #300, opened 2022-07-28
git: Add a new `RepositoryClient` using libgit2

aryan9600:
This PR adds a new package `pkg/git/libgit2`, which implements `git.RepositoryClient` using `libgit2`. The package uses managed smart transports instead of the underlying transports provided by `libgit2`/`git2go`. It needs to link with `libgit2` (which doesn't need to be compiled with `openssl` and `libssh2`). 

Follow up steps:
* Split `pkg/git` into `pkg/git/{repository,gogit,libgit2}`


git/libgit2/repository_client.go
```diff
@@ -0,0 +1,472 @@
...
+}
+
+func NewClient(path string, authOpts *git.AuthOptions) *Client {
+	return &Client{
+		path:     path,
+		authOpts: authOpts,
+	}
+}
+
+var _ git.RepositoryClient = &Client{}
```
darkowlzz, line 66: I wasn't around two weeks ago to emphasize on the naming discussion in https://github.com/fluxcd/pkg/pull/291#discussion_r921514846, but looking at the implementations, it still feel like we don't need `Repository` prefix. `RepositoryClient` interface could just be called `Client` and the other interfaces could just be `Reader` and `Writer`.

hiddeco: I think this would be best paired with your other suggestion in that comment, which would make `git/{repostory,libgit2,gogit}` all individual modules.

aryan9600: Yes, agreed, that's a good idea. Let's tackle that later in a separate PR? This PR is already very large.

Makefile
```diff
@@ -44,7 +70,7 @@ release-%:
 	git push origin "$(REL_PATH)/v$(VER)"
 
 # Find or download controller-gen
-CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
+CONTROLLER_GEN = $(GOBIN)/controller-gen
 .PHONY: controller-gen
 controller-gen: ## Download controller-gen locally if necessary.
 	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.7.0)
```
darkowlzz, line 76: Can you also update controller-gen to 0.9.2?

Makefile
```diff
@@ -3,37 +3,63 @@ MODULES=$(shell find . -mindepth 2 -maxdepth 4 -type f -name 'go.mod' | cut -c 3
...
+		cd $(subst :,/,$*) && go fmt ./... ;\
+	fi
 
 vet-%:
-	cd $(subst :,/,$*); go vet ./...
+	@if [ "$(DIR)" = "git" ]; then \
+		cd git && make vet ;\
+	else \
+		cd $(subst :,/,$*) && go vet ./... ;\
+	fi
```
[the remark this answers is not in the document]
darkowlzz: Since vet requires libgit2/git2go CGO dependencies to be able to analyze the code, let's keep the conditional make target for vet.

git/go.mod
```diff
@@ -19,32 +19,89 @@ require (
...
 	github.com/fluxcd/pkg/ssh v0.5.0
 	github.com/fluxcd/pkg/version v0.1.0
 	github.com/go-git/go-billy/v5 v5.3.1
 	github.com/go-git/go-git/v5 v5.4.2
+	github.com/go-logr/logr v1.2.3
+	github.com/libgit2/git2go/v33 v33.0.9
 	github.com/onsi/gomega v1.19.0
 	golang.org/x/crypto v0.0.0-20220525230936-793ad666bf5e
+	golang.org/x/net v0.0.0-20220607020251-c690dde0001d
+	sigs.k8s.io/controller-runtime v0.12.3
```
darkowlzz, line 33: Let's keep this to the same version that other flux packages and repositories use, v0.11.2, else they'll have to override it and downgrade to the previous version.

git/libgit2/transport/http.go
```diff
@@ -0,0 +1,479 @@
...
+		if req.Method == "POST" {
+			if len(content) == 0 {
+				// a copy of the request body needs to be saved so
+				// it can be reused in case of redirects.
+				if content, err = io.ReadAll(self.reader); err != nil {
+					return err
+				}
+			}
+			req.Body = io.NopCloser(bytes.NewReader(content))
+			req.ContentLength = int64(len(content))
```
darkowlzz, line 432: Compared to the previous implementation, this is different. As far as I know and remember, this piece of code was taken from the upstream implementation of this in git2go.
Based on net/http docs:
> // If body is of type *bytes.Buffer, *bytes.Reader, or
> // *strings.Reader, the returned request's ContentLength is set to its
> // exact value (instead of -1), GetBody is populated (so 307 and 308
> // redirects can replay the body), and Body is set to NoBody if the
> // ContentLength is 0.

and some web search, using `NopCloser()` results in content length to not be set. And not setting content length can be an issue with HTTP/2. Was this causing an issue anywhere?

I'm finding it hard to find reasons for certain changes in the implementation when compared with what's in SC and their implications.
Don't think we need to leave in-line comment about everything, but not sure what's a better way to convey the reasons for the changes now and for the future.

git/libgit2/transport/http.go
```diff
@@ -0,0 +1,479 @@
...
+				}
+			}
+		}
+
+		return nil
+	}
+
+	stream := newManagedHttpStream(t, req, client)
+	stream.recvReply.Add(1)
+	stream.sendRequestBackground()
```
darkowlzz, line 222: In the previous implementation, this is set only when the request method is POST. I couldn't understand the reason for this change.

git/libgit2/transport/http.go
```diff
@@ -0,0 +1,479 @@
...
+)
+
+// registerManagedHTTP registers a Go-native implementation of an
+// HTTP(S) transport that doesn't rely on any lower-level libraries
+// such as OpenSSL.
+func registerManagedHTTP() error {
+	for _, protocol := range []string{"http", "https"} {
+		_, err := git2go.NewRegisteredSmartTransport(protocol, true, httpSmartSubtransportFactory)
+		if err != nil {
+			return fmt.Errorf("failed to register transport for '%s': %v", protocol, err)
```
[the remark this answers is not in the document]
aryan9600: From what I understand, we don't use `%q` because it doesn't play well with the logger. I think @stefanprodan might know about this

git/libgit2/transport/http.go
```diff
@@ -0,0 +1,479 @@
...
+			req.ContentLength = int64(len(content))
+		}
+
+		self.owner.logger.V(logger.TraceLevel).Info("new request", "method", req.Method, "postUrl", req.URL)
+		resp, err = self.client.Do(req)
+		if err != nil {
+			return err
+		}
+		buf := make([]byte, 0)
+		resp.Body.Read(buf)
```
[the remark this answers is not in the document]
aryan9600: Debug code creeped in, please ignore, thanks for spotting.

git/libgit2/transport/init.go
```diff
@@ -0,0 +1,62 @@
...
+			return
+		}
+
+		if err = registerManagedSSH(); err == nil {
+			enabled = true
+		}
+	})
+
+	return err
+}
```
pjbgf, line 62: libgit2 will become a self-contained module (on a follow-up PR), which solely supports Managed Transport. 
If both preceding assumptions are correct, I think the transport initialisation should either happen at `init()`.
Alternatively, the implementation will need to be aware of the transport state, to error when it wasn't initialised properly.

git/libgit2/transport/callback_test.go
```diff
@@ -0,0 +1,139 @@
...
+//
+// Expected hash (used in the cases) can get found with:
+// ssh-keyscan -p 2022 source.developers.google.com | ssh-keygen -l -f -
+var knownHostsFixtureWithPort = `[source.developers.google.com]:2022 ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBB5Iy4/cq/gt/fPqe3uyMy4jwv1Alc94yVPxmnwNhBzJqEV5gRPiRk5u4/JJMbbu9QUVAguBABxL7sBZa5PH/xY=`
+
+// This is an incorrect known hosts entry, that does not aligned with
+// the normalized format and therefore won't match.
+var knownHostsFixtureUnormalized = `source.developers.google.com:2022 ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBB5Iy4/cq/gt/fPqe3uyMy4jwv1Alc94yVPxmnwNhBzJqEV5gRPiRk5u4/JJMbbu9QUVAguBABxL7sBZa5PH/xY=`
+
+func TestKnownHostsCallback(t *testing.T) {
```
pjbgf, line 30: Test should go against `CheckKnownHost` instead.

pjbgf (changes requested):

Makefile
```diff
@@ -1,39 +1,52 @@
...
 all:
 	$(MAKE) $(targets)
 
 tidy-%:
-	cd $(subst :,/,$*); go mod tidy -compat=1.17
+	cd $(subst :,/,$*) && go mod tidy -compat=1.17
 
 fmt-%:
-	cd $(subst :,/,$*); go fmt ./...
+	cd $(subst :,/,$*) && go fmt ./...
```
darkowlzz, line 26: Don't think we need to change this.

Makefile
```diff
@@ -1,39 +1,52 @@
...
+		cd $(subst :,/,$*) && go vet ./... ;\
+	fi
 
 generate-%: controller-gen
-	# Run schemapatch to validate all the kubebuilder markers before generation.
-	cd $(subst :,/,$*); $(CONTROLLER_GEN) schemapatch:manifests="./" paths="./..."
-	cd $(subst :,/,$*); $(CONTROLLER_GEN) object:headerFile="$(root_dir)/hack/boilerplate.go.txt" paths="./..."
+# Run schemapatch to validate all the kubebuilder markers before generation
+# Skip git as this isn't required for that package and increases
+# the complexity unnecessarily (due to libgit2 prequisites)
```
darkowlzz, line 38: These can be reverted.

git/libgit2/transport/callback.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+// certificateCallback constructs a dummy CertificateCallback.
+func certificateCallback() git2go.CertificateCheckCallback {
+	// returning a nil func can cause git2go to panic.
+	return func(cert *git2go.Certificate, valid bool, hostname string) error {
+		return nil
+	}
+}
```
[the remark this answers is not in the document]
aryan9600: I feel it's better to have it like this, because it allows for neat documentation, but I don't mind changing it.

git/libgit2/transport/callback_test.go
```diff
@@ -0,0 +1,144 @@
...
```
pjbgf, line 45:
```suggestion
func TestCheckKnownHost(t *testing.T) {
```

Makefile
```diff
@@ -1,15 +1,18 @@
 VER?=0.0.1
 MODULES=$(shell find . -mindepth 2 -maxdepth 4 -type f -name 'go.mod' | cut -c 3- | sed 's|/[^/]*$$||' | sort -u | tr / :)
 targets=$(addprefix test-, $(MODULES))
-root_dir=$(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+root_dir=$(shell git rev-parse --show-toplevel)
 
-# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
+# Use $GOBIN from the enviornment if set, otherwise use ./bin
```
pjbgf, line 6:
```suggestion
# Use $GOBIN from the environment if set, otherwise use ./bin
```

git/libgit2/transport/http.go
```diff
@@ -0,0 +1,488 @@
...
+
+	for {
+		req := &http.Request{
+			Method: self.req.Method,
+			URL:    self.req.URL,
+			Header: self.req.Header,
+		}
+		req = req.WithContext(self.owner.ctx)
+
+		if req.Method == "POST" {
```
pjbgf, line 433:
```suggestion
		if req.Method == http.MethodPost {
```

git/libgit2/transport/http.go
```diff
@@ -0,0 +1,488 @@
...
+
+		self.owner.logger.V(logger.TraceLevel).Info("new request", "method", req.Method, "postUrl", req.URL)
+		resp, err = self.client.Do(req)
+		if err != nil {
+			return err
+		}
+
+		// GET requests will be automatically redirected.
+		// POST require the new destination, and also the body content.
+		if req.Method == "POST" && resp.StatusCode >= 301 && resp.StatusCode <= 308 {
```
pjbgf, line 453:
```suggestion
		if req.Method == http.MethodPost && resp.StatusCode >= 301 && resp.StatusCode <= 308 {
```

git/libgit2/transport/options.go
```diff
@@ -0,0 +1,70 @@
...
+
+// RemoveTransportOptions removes the registerd TransportOptions object
+// mapped to the provided id.
+func RemoveTransportOptions(transportOptsURL string) {
+	m.Lock()
+	delete(transportOpts, transportOptsURL)
+	m.Unlock()
+}
+
+func GetTransportOptions(transportOptsURL string) (*TransportOptions, bool) {
```
pjbgf, line 61: Maybe?
```suggestion
// GetTransportOptions returns a TransportOptions that matches the transportOptsURL.
// No  matches returns nil with false.
func GetTransportOptions(transportOptsURL string) (*TransportOptions, bool) {
```

git/libgit2/transport/ssh.go
```diff
@@ -0,0 +1,433 @@
...
+			defer cancel()
+		}
+
+		closedAlready := atomic.LoadInt32(t.closedSessions)
+		for {
+			select {
+			case <-ctx.Done():
+				t.logger.V(logger.TraceLevel).Error(ctx.Err(), "context channel Done closed")
+				if atomic.LoadInt32(t.closedSessions) < closedAlready {
+					t.logger.V(logger.TraceLevel).Info("closing transport", "address", addr)
```
pjbgf, line 267: I think `addr` is redundant here, as this is already set on line 175.
But adding `cmd` here could be useful when debugging cross-action or out-of-order closures.

```suggestion
					t.logger.V(logger.TraceLevel).Info("closing transport", "command", cmd)
```

git/libgit2/transport/http.go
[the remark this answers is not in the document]
pjbgf: To make the package consistent, please replace the 3 instances of %q we have with `%s`:
```
gogit/clone.go:441
transport/http.go:185
transport/ssh.go:91
```
The change on the gogit/clone.go is outside the scope of this PR. So happy for that to happen either on this or on the follow-up PR in which we segregate the implementations on their own modules.

pjbgf (approved):
LGTM

great work @aryan9600 getting this into shape. :clap:

darkowlzz:
Please also rebase for the final review.

git/libgit2/transport/http.go
```diff
@@ -0,0 +1,488 @@
...
+			}
+			t.TLSClientConfig = &tls.Config{
+				RootCAs: certPool,
+			}
+		}
+	}
+
+	req.Header.Set("User-Agent", "git/2.0 (flux-libgit2)")
+	if t.Proxy != nil {
+		t.ProxyConnectHeader.Set("User-Agent", "git/2.0 (flux-libgit2)")
```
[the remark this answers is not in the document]
darkowlzz: Missed using the constant here?

git/libgit2/transport/http_test.go
```diff
@@ -0,0 +1,286 @@
...
+			action: git2go.SmartServiceActionUploadpack,
+			transport: &http.Transport{
+				Proxy:              http.ProxyFromEnvironment,
+				ProxyConnectHeader: map[string][]string{},
+			},
+			assertFunc: func(g *WithT, req *http.Request, _ *http.Client) {
+				g.Expect(req.URL.String()).To(Equal("https://final-target/abc/git-upload-pack"))
+				g.Expect(req.Method).To(Equal("POST"))
+				g.Expect(req.Header).To(BeEquivalentTo(map[string][]string{
+					"User-Agent":   {"git/2.0 (flux-libgit2)"},
```
darkowlzz, line 61: We can use the user agent constant here and below.

git/libgit2/transport/http_test.go
```diff
@@ -0,0 +1,286 @@
...
+					"Content-Type": {"application/x-git-upload-pack-request"},
```
darkowlzz, line 62: It may be better to create more constants for all these different types of content types, long strings, error-prone.

git/libgit2/transport/http_test.go
```diff
@@ -0,0 +1,286 @@
...
+			},
+			wantedErr: nil,
+		},
+		{
+			name:      "UploadpackLs: URL, method and headers are correctly set",
+			action:    git2go.SmartServiceActionUploadpackLs,
+			transport: &http.Transport{},
+			assertFunc: func(g *WithT, req *http.Request, _ *http.Client) {
+				g.Expect(req.URL.String()).To(Equal("https://final-target/abc/info/refs?service=git-upload-pack"))
+				g.Expect(req.Method).To(Equal("GET"))
```
darkowlzz, line 73: Like how a lot of the code started using `http.Method*` in this PR, we can use them in the tests as well.

git/utils_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestCleanPath(t *testing.T) {
```
darkowlzz, line 27:
```suggestion
func TestSecurePath(t *testing.T) {
```

git/libgit2/repository_client.go
```diff
@@ -0,0 +1,537 @@
...
+			callbackErr = fmt.Errorf("ref %s rejected: %s", refname, status)
+		}
+		return nil
+	}
+	err = l.remote.Push([]string{fmt.Sprintf("refs/heads/%s:refs/heads/%[1]s", branch)}, &git2go.PushOptions{
+		RemoteCallbacks: callbacks,
+		ProxyOptions:    git2go.ProxyOptions{Type: git2go.ProxyTypeAuto},
+	})
+	if err != nil {
+		if strings.Contains(err.Error(), "early EOF") {
```
darkowlzz, line 377: We also need a test for this similar to the one in https://github.com/fluxcd/image-automation-controller/pull/410 maybe adapted to the tests in this package.

darkowlzz (approved):
LGTM!

darkowlzz:
Needs another rebase.

PR #302, opened 2022-08-01
Add autologin flag options

darkowlzz:
Add runtime/autologin package for autologin flag options.

Tested it with image-reflector-controller.

Part of https://github.com/fluxcd/image-reflector-controller/issues/264.

### DIFF
--- a/runtime/autologin/registry.go
+++ b/runtime/autologin/registry.go
@@ -1,0 +1,60 @@
+/*
+Copyright 2022 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package autologin
+
+import "github.com/spf13/pflag"
+
+const (
+	flagAWSForECR   = "aws-autologin-for-ecr"
+	flagAzureForAKS = "azure-autologin-for-acr"
+	flagGCPForGCR   = "gcp-autologin-for-gcr"
+)
+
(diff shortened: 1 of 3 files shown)


runtime/autologin/registry.go
```diff
@@ -0,0 +1,60 @@
...
+
+// BindFlags will parse the given pflag.FlagSet for auto-login option flags and
+// set the Options accordingly.
+func (o *Options) BindFlags(fs *pflag.FlagSet) {
+	fs.BoolVar(&o.AWSForECR, flagAWSForECR, false,
+		"(AWS) Attempt to get credentials for images in Elastic Container Registry, when no secret is referenced")
+	fs.BoolVar(&o.AzureForAKS, flagAzureForAKS, false,
+		"(Azure) Attempt to get credentials for images in Azure Container Registry, when no secret is referenced")
+	fs.BoolVar(&o.GCPForGCR, flagGCPForGCR, false,
+		"(GCP) Attempt to get credentials for images in Google Container Registry, when no secret is referenced")
```
stefanprodan, line 59: I know this would entail a breaking change, but I would remove the registry service name from the flags so they can be reused for any OIDC based auth, like in the future Git.

stefanprodan (approved):

runtime/autologin/registry.go
darkowlzz: Yeah, I thought about it. Even the actual login code may be common for the different services. But wasn't sure what others thought about it.

stefanprodan: Let's postpone this, I would make this flags true by default and rename them just before GA.

darkowlzz: But since we are more likely to move from flags to per object auto-login configuration, we'll need other set of more generic auto-login flags that can be set to enable login for multiple services in specific providers.
This is making me wonder if we should extract these flags from image-reflector at all if the plan is for others to have per object auto-login configuration and also change image-reflector to do the same before GA.

stefanprodan: Agreed, let's not merge this now as these flags in their current form will only be used in IRC.

stefanprodan:
Closing this as the flags were removed in https://github.com/fluxcd/image-reflector-controller/pull/311

PR #305, opened 2022-08-01
Add OCI client

stefanprodan:
Move the `flux2/internal/oci` code to the `fluxcd/pkg/oci` package under `client`. The client offers basic helper functions for pushing, pulling, tagging and listing OCI artifacts. The client is current usable only in the CLI, future versions of this client should allow swapping the default auth with the OIDC providers from `oci/auth`.


darkowlzz (approved):
Left a comment about adding some docs in the code in meta.go.
Overall, compared with what's in flux2 package, it looks good to me.

darkowlzz (approved):
LGTM!

PR #306, opened 2022-08-02
oci: Use go 1.18 and azure sdk v1

darkowlzz:
Update oci package to use go 1.18, and azcore and azidentity to v1.1.0.
Updated the integration tests accordingly.
Update actions/envtest and build workflow to use go 1.18.


hiddeco (approved):

stefanprodan (approved):
LGTM

Thanks a lot @darkowlzz

PR #307, opened 2022-08-05
Add `org.opencontainers.image.created` to OCI meta

stefanprodan:
Set the created date annotations when pushing OCI artifacts.

### DIFF
--- a/oci/client/list.go
+++ b/oci/client/list.go
@@ -60,6 +60,7 @@ func (c *Client) List(ctx context.Context, url string) ([]Metadata, error) {
 		if m, err := MetadataFromAnnotations(manifest.Annotations); err == nil {
 			meta.Revision = m.Revision
 			meta.Source = m.Source
+			meta.Created = m.Created
 		}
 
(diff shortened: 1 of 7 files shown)


pjbgf (approved):

oci/client/list.go
```diff
@@ -60,6 +60,7 @@ func (c *Client) List(ctx context.Context, url string) ([]Metadata, error) {
 		if m, err := MetadataFromAnnotations(manifest.Annotations); err == nil {
 			meta.Revision = m.Revision
 			meta.Source = m.Source
+			meta.Created = m.Created
```
darkowlzz, line 63: In order to test this, let's add some created value in https://github.com/fluxcd/pkg/blob/680663a877e4096c49dfe733d6355c9e4d959190/oci/client/list_test.go#L40-L43 as well. Since that test does `crane.Push()` the created metadata is not added automatically to the artifact.

darkowlzz (approved):
LGTM!

PR #308, opened 2022-08-05
oci/client: Add internal/fs package for os.Rename

darkowlzz:
`os.Rename()` fails with LinkError when the source and destination are in
different devices.
The following test fails on machines where /tmp is on a different device:
```
    build_test.go:40: 
        Unexpected error:
            <*os.LinkError | 0xc00050f640>: {
                Op: "rename",
                Old: "./testdata/artifact1987351664",
                New: "/tmp/TestBuild3395247411/001/files.tar.gz",
                Err: <syscall.Errno>0x12,
            }
            rename ./testdata/artifact1987351664 /tmp/TestBuild3395247411/001/files.tar.gz: invalid cross-device link
        occurred
--- FAIL: TestBuild (0.00s)
```

internal/fs is copied from source-controller internal/fs. It has an
implementation of Rename with fallback to handle this failure.

This also fixes the build tests failure which failed on some platforms.

This fixes OCI build failure in the CLI.
Before:
```
$ flux build artifact --path ./docs/ --output /tmp/foo.tgz
► building artifact from ./docs/
✗ bulding artifact failed, error: rename ./docs/298647428 /tmp/foo.tgz: invalid cross-device link
```
With this fix:
```
$ flux build artifact --path ./docs/ --output /tmp/foo.tgz
► building artifact from ./docs/
✔ artifact created at /tmp/foo.tgz
```


oci/client/internal/fs/fs_test.go
```diff
@@ -0,0 +1,590 @@
...
+		}
+	}
+}
+
+func cleanUpDir(dir string) {
+	if runtime.GOOS == "windows" {
+		mu.Lock()
+		exec.Command(`taskkill`, `/F`, `/IM`, `git.exe`).Run()
+		mu.Unlock()
+	}
```
pjbgf, line 586: We can remove this as it adds no value to what we are doing withint he project.

darkowlzz: Thanks. Removed it.

pjbgf (approved):
LGTM

thanks @darkowlzz :bow:

stefanprodan (approved):
LGTM 

Thanks @darkowlzz

PR #316, opened 2022-08-17
[pkg/oci]: Add options to filter tags by semver and regex

somtochiama:
Part of: [flux2/#2983](https://github.com/fluxcd/flux2/issues/2983)


stefanprodan (approved):
LGTM

Thanks @somtochiama 🎖️

PR #318, opened 2022-08-22
pkg/git: split into pkg/git/gogit and pkg/git/libgit2

aryan9600:
Update `pkg/git/gogit` and `pkg/git/libgit2` to be self contained modules. Move `pkg/git/e2e` to `pkg/git/internal/e2e`. 


git/internal/e2e/run.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# This script runs e2e tests for pkg/git.
+
+set -o errexit
+PROJECT_DIR=$(git rev-parse --show-toplevel)
+DIR="$(cd "$(dirname "$0")" && pwd)"
+
+cd "$PROJECT_DIR"/git/libgit2 && make libgit2
```
pjbgf, line 8:
```suggestion
cd "${PROJECT_DIR}/git/libgit2" && make libgit2
```

git/internal/e2e/run.sh
```diff
@@ -0,0 +1,22 @@
...
+
+source "$DIR"/setup_gitlab.sh
+
+LIBGIT2_IMG=ghcr.io/fluxcd/golang-with-libgit2
+LIBGIT2_TAG=v0.2.0
```
pjbgf, line 13: On SC we ensure that libgit2 tags are always aligned across the different files that have it set.
We should do the same on pkg, either on this or on a follow-up PR.

https://github.com/fluxcd/source-controller/blob/09ef651bebcd853df555eb689ba69a2cea33b21c/Makefile#L199-L211

git/internal/e2e/run.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# This script runs e2e tests for pkg/git.
+
+set -o errexit
+PROJECT_DIR=$(git rev-parse --show-toplevel)
+DIR="$(cd "$(dirname "$0")" && pwd)"
+
+cd "$PROJECT_DIR"/git/libgit2 && make libgit2
+
+source "$DIR"/setup_gitlab.sh
```
pjbgf, line 10:
```suggestion
source "${DIR}/setup_gitlab.sh"
```

git/internal/e2e/run.sh
```diff
@@ -0,0 +1,22 @@
...
+
+LIBGIT2_IMG=ghcr.io/fluxcd/golang-with-libgit2
+LIBGIT2_TAG=v0.2.0
+LIBGIT2_PKG_CONFIG_PATH=$PROJECT_DIR/git/libgit2/build/libgit2/$LIBGIT2_TAG/lib/pkgconfig
```
pjbgf, line 14:
```suggestion
LIBGIT2_PKG_CONFIG_PATH="${PROJECT_DIR}/git/libgit2/build/libgit2/${LIBGIT2_TAG}/lib/pkgconfig"
```

git/internal/e2e/run.sh
```diff
@@ -0,0 +1,22 @@
...
+CGO_LDFLAGS=$(PKG_CONFIG_PATH="$LIBGIT2_PKG_CONFIG_PATH" pkg-config --libs --static --cflags libgit2)
```
pjbgf, line 15:
```suggestion
CGO_LDFLAGS=$(PKG_CONFIG_PATH="${LIBGIT2_PKG_CONFIG_PATH}" pkg-config --libs --static --cflags libgit2)
```

git/internal/e2e/run.sh
```diff
@@ -0,0 +1,22 @@
...
+
+cd "$DIR"
+PKG_CONFIG_PATH=$LIBGIT2_PKG_CONFIG_PATH CGO_LDFLAGS=$CGO_LDFLAGS go test -v -tags 'netgo,osusergo,static_build,e2e' -race ./...
```
pjbgf, line 18:
```suggestion
cd "${DIR}"
PKG_CONFIG_PATH="${LIBGIT2_PKG_CONFIG_PATH}" CGO_LDFLAGS="${CGO_LDFLAGS}" go test -v -tags 'netgo,osusergo,static_build,e2e' -race ./...
```

Makefile
```diff
@@ -26,8 +26,8 @@ fmt-%:
 	cd $(subst :,/,$*); go fmt ./...
 
 vet-%:
-	@if [ "$(PKG)" = "git" ]; then \
-		cd $(PKG); make vet ;\
+	@if [ "$(PKG)" = "git:libgit2" ]; then \
```
pjbgf, line 29: Maybe just some comments around this, as some folks (such as myself) may not understand this syntax without first looking it up.

pjbgf (commented):
small nits, apart from that looks good.

pjbgf (approved):
LGTM

nice one @aryan9600! :bow:

git/go.mod
```diff
@@ -2,121 +2,22 @@ module github.com/fluxcd/pkg/git
...
-	github.com/go-git/go-billy/v5 v5.3.1
-	github.com/go-git/go-git/v5 v5.4.2
-	github.com/go-logr/logr v1.2.3
-	github.com/libgit2/git2go/v33 v33.0.9
 	github.com/onsi/gomega v1.19.0
-	golang.org/x/crypto v0.0.0-20220427172511-eb4f295cb31f
-	golang.org/x/net v0.0.0-20220225172249-27dd8689420f
-	sigs.k8s.io/controller-runtime v0.11.2
+	golang.org/x/crypto v0.0.0-20220427172511-eb4f295cb31f // indirect
+	golang.org/x/net v0.0.0-20220225172249-27dd8689420f // indirect
```
darkowlzz, line 14: Something looks weird in the mod file change.
These indirect dependencies are usually added in a separate require section below.
And then previously we didn't had `kr/text` and `gopkg.in/check.v1` dependencies. I can't even find any other dependency here that imports them. Tried deleting them and running tidy, looks like they are not required by anyone.

git/libgit2/go.mod
```diff
@@ -0,0 +1,121 @@
...
+	github.com/Masterminds/semver/v3 v3.1.1
+	github.com/ProtonMail/go-crypto v0.0.0-20220812175011-7fcef0dbe794
+	github.com/fluxcd/gitkit v0.6.0
+	github.com/fluxcd/pkg/git v0.5.0
+	github.com/fluxcd/pkg/gittestserver v0.6.0
+	github.com/fluxcd/pkg/gitutil v0.1.0
+	github.com/fluxcd/pkg/http/transport v0.0.1
+	github.com/fluxcd/pkg/runtime v0.16.2
+	github.com/fluxcd/pkg/ssh v0.5.0
+	github.com/fluxcd/pkg/version v0.1.0
```
darkowlzz, line 38: Just proposing an idea, I need some consensus on this. How about setting the version of all these dependencies that are relative to the package to `v0.0.0`? That way, while reviewing the dependency changes, these can easily be ignored without carefully looking at the import path.
I don't remember where I've seen/read this before, but for some reason I tend to use v0.0.0 by default.

git/internal/e2e/go.mod
```diff
@@ -0,0 +1,124 @@
...
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+	k8s.io/api v0.24.2 // indirect
+	k8s.io/apiextensions-apiserver v0.24.2 // indirect
+	k8s.io/apimachinery v0.24.2 // indirect
+	k8s.io/client-go v0.24.2 // indirect
+	k8s.io/component-base v0.24.2 // indirect
+	k8s.io/klog/v2 v2.60.1 // indirect
+	k8s.io/kube-openapi v0.0.0-20220328201542-3ee0da9b0b42 // indirect
+	k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9 // indirect
+	sigs.k8s.io/controller-runtime v0.12.3 // indirect
```
darkowlzz, line 120: Don't think we are ready for v0.12 yet.

git/internal/e2e/go.mod
```diff
@@ -0,0 +1,124 @@
+module github.com/fluxcd/pkg/git/internal/e2e
```
darkowlzz, line 1: Wondering if we need to make this internal.
Usually the tests are placed in a `tests` directory with more subdirectories for the types of tests.
Internal is used to make packages private. But from what I've seen, the tests directory is not made internal and this package seems to be in a similar situation.

git/internal/e2e/run.sh
```diff
@@ -0,0 +1,22 @@
...
+DIR="$(cd "$(dirname "$0")" && pwd)"
+
+cd "${PROJECT_DIR}/git/libgit2" && make libgit2
+
+source "${DIR}/setup_gitlab.sh"
+
+LIBGIT2_IMG=ghcr.io/fluxcd/golang-with-libgit2
+LIBGIT2_TAG=v0.2.0
+LIBGIT2_PKG_CONFIG_PATH="${PROJECT_DIR}/git/libgit2/build/libgit2/${LIBGIT2_TAG}/lib/pkgconfig"
+CGO_LDFLAGS=$(PKG_CONFIG_PATH="${LIBGIT2_PKG_CONFIG_PATH}" pkg-config --libs --static --cflags libgit2)
```
darkowlzz, line 15: These variables are now being created here and in Makefile of libgit2 package.
Would it be better to put these in some script and source and populate the values in env vars?
That way, we have one source of these values, less scattered, easy to maintain in the long term.

darkowlzz: Maybe this applies to the other variables like the image and tag. Updating in one file, would update both libgit2 package and this package.

git/libgit2/go.mod
stefanprodan: You can't import packages that depend on some non-exiting ones, we need to bump the version every time for our consumers.

git/internal/e2e/go.mod
stefanprodan: Yes let's make the tests internal please.

aryan9600: I decided to make it internal because it has it's own `go.mod`. Otherwise, I would agree with you @darkowlzz

git/internal/e2e/run.sh
hiddeco: I think that the the value of consolidating the variables to improve long term maintenance outweighs the other concerns.

Regarding "bash getting cumbersome", if it's a simple script with the expectation to be sourced. I think it would not contain much more than e.g. `export`, `declare` and _maybe_ `readonly` declarations. Which should be fairly easy to understand (the scope of), and extend.

git/libgit2/env.sh
```diff
@@ -0,0 +1,11 @@
...
+# This script contains setup variables related to libgit2.
+# It's purpose is to reduce boilerplate, and provide a single source
+# of info for all other build scripts.
+
+LIBGIT2_IMG=ghcr.io/fluxcd/golang-with-libgit2
+LIBGIT2_TAG=v0.2.0
+LIBGIT2_PATH=${BUILD_DIR}/libgit2/${LIBGIT2_TAG}
+LIBGIT2_LIB_PATH=${LIBGIT2_PATH}/lib
+LIBGIT2=${LIBGIT2_LIB_PATH}/libgit2.a
+PKG_CONFIG_PATH=${LIBGIT2_LIB_PATH}/pkgconfig
```
pjbgf, line 11: I think we should either have environment variables set in a `.env` file, or export such vars in a `.sh`. I am leaning more towards the latter.

darkowlzz (approved):
LGTM!

hiddeco (approved):
LGTM as well, thanks a lot @aryan9600 :bow:

stefanprodan (approved):
👏

PR #321, opened 2022-08-24
[pkg/oci]: Add Delete function

somtochiama:
Part of: [fluxv2/#2983](https://github.com/fluxcd/flux2/issues/2982)

### DIFF
--- a/oci/client/delete.go
+++ b/oci/client/delete.go
@@ -1,0 +1,35 @@
+/*
+Copyright 2022 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


somtochiama:
Failing git-test is fixed #318

oci/client/internal/fs/fs_test.go
```diff
@@ -41,7 +41,6 @@ func TestMain(m *testing.M) {
 	}
 
 	os.MkdirAll("testdata/symlinks", 0o755)
-	defer os.RemoveAll("testdata/symlinks")
```
stefanprodan, line 44: Are OCI tests broken in main? If so please move this to a different PR as it's not related to client.delete.

somtochiama: They aren't broken per se. The defer function never runs because of os.Exit() so the created directory doesn't get removed.

But you are right. I would move it to a different pr

stefanprodan:
@somtochiama please rebase and mark the PR ready for review.

stefanprodan (approved):
LGTM

Thanks @somtochiama

PR #326, opened 2022-08-25
git/libgit2: remove dep to pkg/runtime

aryan9600:


stefanprodan (commented):
Please remove pkg/runtime from go.mod and run locally tidy and test

stefanprodan (changes requested):
Please remove `github.com/fluxcd/pkg/runtime => ../../runtime`

stefanprodan (approved):
LGTM

Thanks @aryan9600

PR #328, opened 2022-08-25
[pkg/oci] Delete image with digest

somtochiama:
Part of [flux2#2882](https://github.com/fluxcd/flux2/issues/2982).
Follow up to #321 

This pull request uses the image digest to delete the image instead of tags.
ECR/GHCR doesn't support using the docker Delete API (which is what crane uses) to delete images; therefore, it won't work for them.

## Provider Specific Considerations
1. GHCR doesn't support deletion using the DELETE Docker API. It returns a `405 Unsupported` error. Deleting images is allowed using the Github API. For us to support this, it would require getting an authentication token from the user. For more info, see https://github.com/orgs/community/discussions/26267
2. ECR also doesn't support deletion using the DELETE Docker API. It returns `404 Not Found`.
3. GCR requires that all tags referencing a layer be deleted first. It returns `GOOGLE_MANIFEST_DANGLING_TAG: Manifest is still referenced by one or more tags` Currently, we delete the tag first, then delete the image. But if there are other tags still referencing the image, the delete will fail
4. Azure doesn't support tag deletion, Deletion by digest works okay.
Notes:
Deletion by tag means using a tag as reference in the URL -> `ghcr.io/podinfo:v0.0.1`
Deletion by image means using a digest as the reference in the URL `ghcr.io/podinfo@sha256:4ed5....`. This deletes the layer in the container registry

This pull request is currently on hold while we decide if we still want to support deletion and how best to go about it.

### DIFF
--- a/oci/client/delete.go
+++ b/oci/client/delete.go
@@ -19,16 +19,41 @@ package client
 import (
 	"context"
 	"fmt"
+	"github.com/fluxcd/pkg/oci/auth/gcp"
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/google/go-containerregistry/pkg/name"
 )
 
 // Delete deletes a particular image from an OCI repository
-// If the url has no tag, the latest image is deleted
-func (c *Client) Delete(ctx context.Context, url string) error {
-	_, err := name.ParseReference(url)
+// For GCR, GAR, the tag is deleted first, then the image
+// as it doesn't let you delete an image if a tag still references it.
+func (c *Client) Delete(ctx context.Context, url string, tag string) error {
+	ref, err := name.ParseReference(url)
 	if err != nil {
 		return fmt.Errorf("invalid URL: %w", err)
+	}
+
+	if tag != "" {
+		img, err := crane.Pull(fmt.Sprintf("%s:%s", url, tag))
+		if err != nil {
+			return fmt.Errorf("error getting digest: %s", err)
+		}
+
+		hash, err := img.Digest()
+		if err != nil {
+			return err
+		}
+
+		// GCP registry doesn't permit deletion of image if it is still
+		// referenced by a tag.
+		if gcp.ValidHost(ref.Context().RegistryStr()) {
+			err = crane.Delete(fmt.Sprintf("%s:%s", ref.Context(), tag))
+			if err != nil {
+				return err
+			}
+		}
+
+		url = fmt.Sprintf("%s@%s", ref.Context(), hash.String())
 	}
 
 	return crane.Delete(url, c.optionsWithContext(ctx)...)
--- a/oci/client/delete.go
+++ b/oci/client/delete.go
@@ -27,6 +27,8 @@ import (
 // Delete deletes a particular image from an OCI repository
 // For GCR, GAR, the tag is deleted first, then the image
 // as it doesn't let you delete an image if a tag still references it.
+// This doesn't work with ECR/GHCR since they don't support DELETE according
+// to the docker API spec.
 func (c *Client) Delete(ctx context.Context, url string, tag string) error {
 	ref, err := name.ParseReference(url)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


makkes:
Please either add a link to the issue this PR resolves or provide a little more context in the PR description. 🙏🏻

darkowlzz:
The registry server logs are very verbose and distracting. I'd like to discard them by doing

```go
	// Discard the registry server logs.
	config.Log.AccessLog.Disabled = true
	logger := logrus.New()
	logger.SetOutput(ioutil.Discard)
	entry := logrus.NewEntry(logger)
	dcontext.SetDefaultLogger(entry)
```
before creating a new registry in https://github.com/fluxcd/pkg/blob/38b2cb21a166c144ff420ae40fafdf0844349814/oci/client/suite_test.go#L55 . Would like to do the same in SC tests as well.

oci/tests/integration/suite_test.go
```diff
@@ -82,8 +82,11 @@ var (
 	// kubernetes client of the created cluster.
 	testEnv *tftestenv.Environment
 
+	// deleteTag is the tag that will be deleted when testing the delete func
+	deleteTag = "test-delete"
+
 	// testImageTags are the tags used in the test for the generated images.
-	testImageTags = []string{"v0.1.0", "v0.1.2", "v0.1.3", "v0.1.4"}
+	testImageTags = []string{"v0.1.0", "v0.1.2", "v0.1.3", "v0.1.4", deleteTag}
```
darkowlzz, line 89: Since the other existing tests are very primitive, without any knowledge of scanned tags, we can just pick and delete any of the existing tags instead of introducing a new one. `deleteTag` can be set to any of the existing tags above.
Maybe in the future as we find a need for each test to have its own set of tags, we can move the image create and push operation into the test functions.

oci/tests/integration/Makefile
```diff
@@ -6,14 +6,14 @@ TEST_IMG ?= fluxcd/testapp:test
...
+	CGO_ENABLED=0 GOOS=linux go build -v -o app ./testapp
 
 docker-build: app
-	docker build -t $(TEST_IMG) --load .
+	docker build -t $(TEST_IMG) .
 
 test:
 	docker image inspect $(TEST_IMG) >/dev/null
-	TEST_IMG=$(TEST_IMG) go test -timeout $(TEST_TIMEOUT) -v ./ $(GO_TEST_ARGS) $(PROVIDER_ARG) --tags=integration
+	TEST_IMG=$(TEST_IMG) go test -timeout $(TEST_TIMEOUT) -v ./ $(PROVIDER_ARG) $(GO_TEST_ARGS) --tags=integration
```
darkowlzz, line 16: Was this causing any problem?

oci/tests/integration/Makefile
```diff
@@ -6,14 +6,14 @@ TEST_IMG ?= fluxcd/testapp:test
 
 .PHONY: app
 app:
-	CGO_ENABLED=0 go build -v -o app ./testapp
+	CGO_ENABLED=0 GOOS=linux go build -v -o app ./testapp
 
 docker-build: app
-	docker build -t $(TEST_IMG) --load .
+	docker build -t $(TEST_IMG) .
```
darkowlzz, line 12: We need `--load` because I believe most of the developer machines and CI setup use docker buildx which requires this flag to copy the image from the builder to the local image store. Can you move this to a variable, like we do in kustomize-controller and helm-controller, refer https://github.com/fluxcd/helm-controller/blob/1758df2b5b314a9940a47e3c8cbb6b1faa45b16d/Makefile#L22-L23 .

oci/client/delete.go
```diff
@@ -19,17 +19,44 @@ package client
...
 
 // Delete deletes a particular image from an OCI repository
-// If the url has no tag, the latest image is deleted
-func (c *Client) Delete(ctx context.Context, url string) error {
-	_, err := name.ParseReference(url)
+// For GCR, GAR, the tag is deleted first, then the image
+// as it doesn't let you delete an image if a tag still references it.
+// This doesn't work with ECR/GHCR since they don't support DELETE according
+// to the docker API spec.
+func (c *Client) Delete(ctx context.Context, url string, tag string) error {
```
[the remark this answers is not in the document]
somtochiama: You are right about removing the tag argument but I think it might best to not add deleting by using digest as an option, since we always delete using the digest (GCP doesn't permit deletion if the image still has a tag, and deleting the tag only deletes the tag, not the image itself. Azure doesn't allow deletion using tags).

We could still add a delete option but right now I am not sure what should be in it.
@stefanprodan suggested putting this issue on hold while we figure out if we do want to add this feature and how best to go about it.

oci/tests/integration/Makefile
somtochiama: I don't remember any. I will undo the change

PR #332, opened 2022-08-25
git/libgit2: improve http sub transport

aryan9600:
Introduce small changes to the libgit2 HTTP sub transport:
1. Get rid of `sentRequest` in `httpSmartSubtransportStream`.
2. Simplify `sendRequest` usage overall.
3. Return `err` in methods instead of `self.httpError`.

### DIFF
--- a/git/libgit2/transport/http.go
+++ b/git/libgit2/transport/http.go
@@ -238,10 +238,8 @@ func (t *httpSmartSubtransport) Action(transportOptionsURL string, action git2go
 	}
 
 	stream := newManagedHttpStream(t, req, client)
-	if req.Method == http.MethodPost {
-		stream.recvReply.Add(1)
-		stream.sendRequestBackground()
-	}
+	stream.recvReply.Add(1)
+	stream.sendRequestBackground()
 
 	return stream, nil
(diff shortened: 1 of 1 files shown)


git/libgit2/transport/http.go
```diff
@@ -238,10 +238,8 @@ func (t *httpSmartSubtransport) Action(transportOptionsURL string, action git2go
 	}
 
 	stream := newManagedHttpStream(t, req, client)
-	if req.Method == http.MethodPost {
-		stream.recvReply.Add(1)
-		stream.sendRequestBackground()
-	}
+	stream.recvReply.Add(1)
+	stream.sendRequestBackground()
```
pjbgf, line 242: I am not entirely sure why upstream was handling this only on `POST`.
But I probably would wait until `pkg` is fully integrated across the controllers before making this change.

pjbgf:
As discussed off-line, let's **hold** merging these changes until we reflactor and start using this package in SC/IAC.

pjbgf:
Giving the prospect of `libgit2` being discontinued in the coming months, I think we should close this.

PR #337, opened 2022-08-29
git/libgit2: add support for ssh proxy and tests

aryan9600:
This PR adds support for specifying custom proxies while using SSH and tests for both HTTP(S) and SSH proxies. It also adds a `ClientOption` to let users configure a proxy for the client.


git/libgit2/transport/ssh.go
```diff
@@ -295,15 +295,53 @@ func (t *sshSmartSubtransport) Action(transportOptionsURL string, action git2go.
...
+				conn, err = ctxDialer.DialContext(ctx, "tcp", addr)
+			} else {
+				conn, err = dialer.Dial("tcp", addr)
+			}
+		default:
+			conn, err = net.Dial("tcp", addr)
+		}
+	} else {
+		conn, err = net.Dial("tcp", addr)
+	}
```
pjbgf, line 340: `net.Dial` does not use context, so we would need to ensure Timeouts are enforced in a different manner.
Otherwise we will be breaking the behaviour established on: https://github.com/fluxcd/source-controller/pull/740

git/libgit2/transport/ssh_proxy_test.go
```diff
@@ -0,0 +1,176 @@
...
+type TestProxyRule struct{}
+
+func (dr TestProxyRule) Allow(ctx context.Context, req *socks5.Request) (context.Context, bool) {
+	// if req.DestAddr.FQDN == "example.com" {
+	// req.DestAddr.FQDN = ""
+	// req.DestAddr.IP = net.ParseIP(gitServerAddr.Hostname())
+	// port, _ := strconv.Atoi(gitServerAddr.Port())
+	// req.DestAddr.Port = port
+	// fmt.Println(req.DestAddr)
+	// }
```
pjbgf, line 173: Maybe?
```suggestion
```

pjbgf (changes requested):
We may want to first start using pkg/git into SC/IAC and review tests before adding new features or changing existing behaviour.

git/libgit2/transport/ssh.go
aryan9600: I think we should get rid of this altogether. If `proxyOpts.Type == git2go.ProxyTypeSpecified` then we do `dialer, err := proxy.FromURL()` else `dialer, err := proxy.FromEnviornment()`, since that'll just do a regular `net.Dial()` when no env vars are present.

aryan9600: or we could just use `net.DialTimeout()`, if we want to honor all possible values of `proxyOpts` (which is not unreasonable since this is a public libgit2 package)

pjbgf: From a context enforcement that sounds reasonable. 

Overall users would be a bit more in control as to when they want to auto-detect proxy settings enabled, as opposed to the current "always on".

How would this be exposed to the end user? Would they have a `ProxyTypeAuto` set by default, and then have the ability to pick a per object setting? On that point, can we do something similar for `go-git`?

pjbgf:
Now that SC/IAC have the refactoring changes merges, we should be in a good place to merge this.

Please rebase.

git/libgit2/transport/http_proxy_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+				return &git.AuthOptions{
+						Transport: git.HTTP,
+					}, func() {
+					}
+			},
+			wantUsedProxy: false,
+		},
+		{
+			name:          "specfied proxy host",
```
pjbgf, line 149:
```suggestion
			name:          "specified proxy host",
```

git/libgit2/transport/http_proxy_test.go
```diff
@@ -0,0 +1,267 @@
...
+	// The certificate used here is valid for both example.com and localhost.
+	var proxyHandler goproxy.FuncReqHandler = func(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
+		userAgent := req.Header.Get("User-Agent")
+		if strings.Contains(req.Host, "example.com") && strings.Contains(userAgent, "git") {
+			atomic.AddInt32(proxiedRequests, 1)
+			req.Host = u.Host
+			req.URL.Host = req.Host
+			return req, nil
+		}
+		// Reject if it isnt our request.
```
pjbgf, line 256:
```suggestion
		// Reject if it isn't our request.
```

git/libgit2/go.mod
```diff
@@ -31,6 +31,8 @@ replace github.com/go-git/go-billy/v5 => github.com/go-git/go-billy/v5 v5.3.2-0.
 require (
 	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/ProtonMail/go-crypto v0.0.0-20221026131551-cf6655e29de4
+	github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5
+	github.com/elazarl/goproxy v0.0.0-20220529153421-8ea89ba92021
```
pjbgf, line 35: `github.com/elazarl/goproxy` seems unmaintained, but mostly self-contained. It would be good to find an alternative at some point.

git/libgit2/transport/ssh_proxy_test.go
```diff
@@ -0,0 +1,169 @@
...
+			name:          "specified proxy host",
+			proxyType:     git2go.ProxyTypeSpecified,
+			wantUsedProxy: true,
+		},
+		{
+			name:          "http through socks",
+			proxyType:     git2go.ProxyTypeSpecified,
+			wantUsedProxy: true,
+			httpServer:    true,
+		},
```
pjbgf, line 64: Please add an use case in which `wantUsedProxy` is `false`.

pjbgf (changes requested):
Tiny nits, apart from those LGTM.

pjbgf (approved):
LGTM

PR #342, opened 2022-08-31
git: fallback to 'git' for username only for ssh

aryan9600:
Fallback to 'git' for username only when cloning Git repos through SSH,
since using 'git' as the username when cloning Git repos through HTTP(S)
might lead to 401s.

### DIFF
--- a/git/options.go
+++ b/git/options.go
@@ -110,14 +110,11 @@ func (o AuthOptions) Validate() error {
 	return nil
 }
 
-// NewAuthOptions constructs an AuthOptions object from the given map.
+// NewAuthOptions constructs an AuthOptions object from the given map and URL.
 // If the map is empty, it returns a minimal AuthOptions object after
 // validating the result.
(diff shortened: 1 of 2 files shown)


git/options.go
```diff
@@ -146,11 +147,11 @@ func NewAuthOptions(u url.URL, data map[string][]byte) (*AuthOptions, error) {
 // newAuthOptions constructs a minimal AuthOptions object from the
 // given URL and then validates the result. It returns the AuthOptions, or an
 // error.
```
pjbgf, line 149: Maybe?
```suggestion
// given URL and then validates the result. It returns the AuthOptions, or nil.
```

pjbgf (approved):
LGTM

thanks @aryan9600... 🙇

stefanprodan:
@aryan9600 would this fix https://github.com/fluxcd/source-controller/issues/880?

PR #343, opened 2022-08-31
git/libgit2: use go-billy/util.Walk to traverse files when committing

aryan9600:
Use https://github.com/go-git/go-billy/blob/master/util/walk.go#L59 to traverse files when committing files.


git/libgit2/repository_client.go
```diff
@@ -249,21 +249,16 @@ func (l *Client) Commit(info git.Commit, signer *openpgp.Entity) (string, error)
...
 			if err != nil {
 				return err
 			}
-
-			rel, err := filepath.Rel(l.repository.Workdir(), path)
-			if err != nil {
-				return err
-			}
-			if info.IsDir() || strings.HasPrefix(rel, ".git") || rel == "." {
+			if info.IsDir() || strings.HasPrefix(path, ".git") || path == "" {
```
pjbgf, line 257: Previous code checked for `"."` instead of `""`, is this just due to the underlying library only? Is there any impact?

aryan9600: It's because we pass `""` as the `root` path to `util.Walk()`. `util.Walk()` calls the callback recursively on each file presented inside the chrooted filesystem, but on the first iteration, it calls the callback on the `root` path, which is `""` in our case. (ref: https://github.com/go-git/go-billy/blob/master/util/walk.go#L18)

git/libgit2/go.mod
```diff
@@ -34,7 +34,7 @@ require (
 	github.com/fluxcd/pkg/http/transport v0.0.1
 	github.com/fluxcd/pkg/ssh v0.6.0
 	github.com/fluxcd/pkg/version v0.2.0
-	github.com/go-git/go-billy/v5 v5.3.1
+	github.com/go-git/go-billy/v5 v5.3.2-0.20210603175951-e0768be422ff
```
pjbgf, line 37: Please add a comment to explain why we are pinning to the specific commit.

git/libgit2/repository_client.go
```diff
@@ -248,22 +248,18 @@ func (l *Client) Commit(info git.Commit, signer *openpgp.Entity) (string, error)
...
 			}
-
-			rel, err := filepath.Rel(l.repository.Workdir(), path)
-			if err != nil {
-				return err
-			}
-			if info.IsDir() || strings.HasPrefix(rel, ".git") || rel == "." {
+			// skip symlinks and any files that are within ".git/"
+			if !info.Mode().IsRegular() || info.IsDir() ||
+				strings.HasPrefix(path, ".git") || path == "" {
```
pjbgf, line 258: We should be able to drop the `info.IsDir()` due to https://github.com/golang/go/blob/fa0e3bffb494fb9f97c241d6579daa443fbcbb06/src/os/types.go#L38

```suggestion
			if !info.Mode().IsRegular() || 		strings.HasPrefix(path, ".git") || path == "" {
```

pjbgf (approved):
LGTM

thank you @aryan9600 for following up on this. :bow:

PR #344, opened 2022-08-31
ci: use inbuilt cache feature of setup-go

aryan9600:
Use the inbuilt cache feature in `actions/setup-go` which internally uses `actions/cache` itself.

This lets us cache module dependencies and previous builds as well which speeds up the entire workflow considerably from [~24 mins](https://github.com/fluxcd/pkg/runs/8111769892?check_suite_focus=true) to [~8mins](https://github.com/fluxcd/pkg/runs/8114612137?check_suite_focus=true) on Ubuntu.
It also fixes an issue where the cache was not being used on macOS runners: https://github.com/fluxcd/pkg/runs/8111770057?check_suite_focus=true#step:3:8

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -18,17 +18,15 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Restore Go cache
-        uses: actions/cache@v1
-        with:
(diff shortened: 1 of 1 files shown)


pjbgf (approved):
LGTM

great job @aryan9600! :bow:

PR #345, opened 2022-09-01
git: tidy the `RepositoryWriter` interface

aryan9600:
Rename `CheckoutOptions` to a more apt name: `CloneOptions`.
Remove `WriteFile()`, and modify `Commit` to accept an optional argument to be used to write files before committing.


git/options.go
```diff
@@ -27,8 +27,8 @@ const (
 	DefaultPublicKeyAuthUser = "git"
 )
 
-// CheckoutOptions are the options used for a Git checkout.
-type CheckoutOptions struct {
+// CloneOptions are the options used for a Git checkout.
```
pjbgf, line 30: Maybe?
```suggestion
// CloneOptions are the options used for a Git clone.
```

git/gogit/repository_client.go
```diff
@@ -156,7 +156,7 @@ func (g *Client) Init(ctx context.Context, url, branch string) error {
 	return nil
 }
 
-func (g *Client) Clone(ctx context.Context, url string, checkoutOpts git.CheckoutOptions) (*git.Commit, error) {
+func (g *Client) Clone(ctx context.Context, url string, checkoutOpts git.CloneOptions) (*git.Commit, error) {
```
pjbgf, line 159:
```suggestion
func (g *Client) Clone(ctx context.Context, url string, cloneOpts git.CloneOptions) (*git.Commit, error) {
```

git/client.go
```diff
@@ -27,7 +27,7 @@ import (
 // on a Git repository.
 type RepositoryReader interface {
 	// Clone clones a repository from the provided url using the options provided.
-	Clone(ctx context.Context, url string, checkoutOpts CheckoutOptions) (*Commit, error)
+	Clone(ctx context.Context, url string, checkoutOpts CloneOptions) (*Commit, error)
```
pjbgf, line 30: Maybe
```suggestion
	Clone(ctx context.Context, url string, cloneOpts CloneOptions) (*Commit, error)
```

Or
```suggestion
	Clone(ctx context.Context, url string, opts CloneOptions) (*Commit, error)
```

git/client.go
```diff
@@ -48,13 +48,11 @@ type RepositoryWriter interface {
...
 	SwitchBranch(ctx context.Context, branch string) error
-	// WriteFile creates a new file at the given path, reads from the reader and
-	// writes the content to the file. If the file already exists, its overwritten.
-	WriteFile(path string, reader io.Reader) error
-	// Commit commits any changes made to the repository.
+	// Commit writes files (if desired) and then commits any changes made to the repository.
 	// signer is an optional argument which can be provided to sign the commit using
-	// OpenPGP.
-	Commit(info Commit, signer *openpgp.Entity) (string, error)
+	// OpenPGP. files is an optional argument with filenames as its key and the content
```
pjbgf, line 53:
```suggestion
	// OpenPGP. 
	// files is an optional argument with filenames as its key and the content
```

git/options.go
```diff
@@ -27,36 +27,49 @@ const (
...
-	// reconciliation.
-	// It is used to skip clone operations when no changes were detected.
-	LastRevision string
+	// LastObservedCommit holds the last observed commit hash of a
+	// Git repository.
+	// If provided, the clone operation will compare it with the HEAD commit
+	// of the branch or tag (as configured via CheckoutStrategy) in the remote
+	// repository. If they match, cloning will be skipped and a "non-concrete"
+	// commit will be returned, which can be verified using `IsConcreteCommit()`.
+
```
darkowlzz, line 46: This line break doesn't work well with godoc. The actual comment above doesn't show up in the godoc.

git/client.go
```diff
@@ -48,13 +48,13 @@ type RepositoryWriter interface {
...
-	// signer is an optional argument which can be provided to sign the commit using
-	// OpenPGP.
-	Commit(info Commit, signer *openpgp.Entity) (string, error)
+	// Commit writes files (if desired) and then commits any changes made to the
+	// repository.
+	// signer is an optional argument which can be provided to sign the commit
+	// using OpenPGP.
+	// files is an optional argument with file names as its key and the content
+	// of the file as the value. If the file already exists, its overwritten.
+	Commit(info Commit, signer *openpgp.Entity, files map[string]io.Reader) (string, error)
```
darkowlzz, line 57: If the signer and files are optional, how about creating an option type for them? Then there won't be any `nil, nil` being passed in a lot of the code that uses or tests this.

aryan9600: Do you mean creating something like `CommitOptions`? I like the idea of avoiding `nil, nil` being passed in. But, would something like` files` naturally fit into a type named `CommitOptions`?

darkowlzz: yes. `Files` can just be an attribute of `CommitOptions` in the sense these are the files that should be part of the commit.

pjbgf: +1 on using `CommitOptions` instead of `signer` and `files`.

git/client.go
```diff
@@ -48,13 +45,9 @@ type RepositoryWriter interface {
...
-	// WriteFile creates a new file at the given path, reads from the reader and
-	// writes the content to the file. If the file already exists, its overwritten.
-	WriteFile(path string, reader io.Reader) error
-	// Commit commits any changes made to the repository.
-	// signer is an optional argument which can be provided to sign the commit using
-	// OpenPGP.
-	Commit(info Commit, signer *openpgp.Entity) (string, error)
+	// Commit commits any changes made to the repository. commitOpts is an
+	// optional argument which can be provided to configure the commit.
+	Commit(info Commit, commitOpts *CommitOptions) (string, error)
```
darkowlzz, line 50: This is fine but I was thinking more in terms of functional options, for example the API of the patch helper in https://github.com/fluxcd/pkg/blob/10b923813d45de15d052264f7fa345ab21a6c1b2/runtime/conditions/patch.go#L125.
A good description of this pattern is https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis.
This way, you don't even have to pass anything, really optional.

git/options.go
```diff
@@ -39,22 +67,39 @@ type CheckoutOptions struct {
...
-	// not supported by all Implementations.
-	RecurseSubmodules bool
+// CommitOptions provides options to configure a Git commit operation.
+type CommitOptions struct {
+	// Signer can be used to sign a commit using OpenPGP.
+	Signer *openpgp.Entity
+	// Files contains file names mapped to the file's content.
+	// Its used to write files which are then included in the commit.
+	Files map[string]io.Reader
+}
```
darkowlzz, line 82: With the function options, this struct can now be private. Options can be set via their function options.

aryan9600: But how would external packages that implement `Commit` actually access the struct?

darkowlzz: The public API to configure `CommitOption` is `WithSigner(value)` and `WithFiles(value)`. They don't need to directly use the `CommitOptions`. We have this pattern everywhere.

```go
r.Commit(c, WithSigner(s), WithFiles(f))
```

darkowlzz: Oh, this is in the git package and you need it in the other packages.

aryan9600: How would they get access to the underlying signer or files to be used in the actual implementation of `Commit()`? `pkg/git/gogit` can't access `pkg/git.commitOptions`

darkowlzz: Looks like an exception to the pattern because of the dependency of others on it. Don't think we can do anything about it. Let it be.

darkowlzz (approved):
LGTM!

git/libgit2/clone.go
```diff
@@ -158,26 +158,26 @@ func (l *Client) cloneTag(ctx context.Context, url, tag string, opts git.Checkou
...
-				currentAnnotatedRevision := fmt.Sprintf("%s/%s", tag, hash)
-				if currentAnnotatedRevision == opts.LastRevision {
+				remoteAnnotatedHEAD := fmt.Sprintf("%s/%s", tag, hash)
+				if remoteAnnotatedHEAD == opts.LastObservedCommit {
 					same = true
 				}
 			}
 			if same {
-				// Construct a partial commit with the existing information.
+				// Construct a non-conrete commit with the existing information.
```
pjbgf, line 180:
```suggestion
				// Construct a non-concrete commit with the existing information.
```

pjbgf (approved):
LGTM

great stuff @aryan9600! :astronaut:

PR #350, opened 2022-09-06
build: Restore Go cache across actions

pjbgf:
Previous to this change the Go cache would only be restored to subsequent
executions of the same module with unchanged dependencies.

This fallbacks to a generic Go cache when dependencies change for example.
It should be a safe and faster alternative for builds that do not generate
binaries.


stefanprodan (approved):

stefanprodan:
Please add caching to CodeQL, it takes 30m on each run.

hiddeco:
Think this became obsolete with #514.

PR #351, opened 2022-09-06
git/libgit2: use uuid for transport options usage 

aryan9600:
Use UUID for transport options URL, decreasing the chance of a clash and
hence use of the wrong transport options for a Git operation.


pjbgf (approved):
LGTM

pjbgf:
The fuzzing issue seems to be due to some upstream change, investigating that in parallel.

pjbgf (approved):
LGTM

PR #358, opened 2022-09-21
ci: add e2e workflow to path filtering and workflow_dispatch trigger

aryan9600:


stefanprodan (approved):
LGTM

Thanks @aryan9600 🏅

PR #362, opened 2022-09-26
Add regex to support ECR as OCI Helm registry

bnsfrt:
This change adds a new regex to parse AWS ECR OCI URLs when to allow URLs that point to the root of an AWS account's ECR registry, instead of forcing all repositories to contain a "/" character.

When accepted this will resolve https://github.com/fluxcd/source-controller/issues/905.

I've verified this change resolves the issue by rebuilding the source-controller image and deploying it to a Kubernetes cluster and verifying it can properly authenticate and pull from AWS ECR when given a URL like `oci://<account>.dkr.ecr.<region>.amazonaws.com` in a HelmRepository object.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203048932854531

### DIFF
--- a/oci/auth/aws/auth.go
+++ b/oci/auth/aws/auth.go
@@ -33,13 +33,13 @@ import (
 	"github.com/fluxcd/pkg/oci"
 )
 
-var registryPartRe = regexp.MustCompile(`([0-9+]*).dkr.ecr.([^/.]*)\.(amazonaws\.com[.cn]*)/([^:]+):?(.*)`)
+var registryPartRe = regexp.MustCompile(`([0-9+]*).dkr.ecr.([^/.]*)\.(amazonaws\.com[.cn]*)`)
 
-// ParseImage returns the AWS account ID and region and `true` if
-// the image repository is hosted in AWS's Elastic Container Registry,
+// ParseRegistry returns the AWS account ID and region and `true` if
+// the image registry/repository is hosted in AWS's Elastic Container Registry,
 // otherwise empty strings and `false`.
-func ParseImage(image string) (accountId, awsEcrRegion string, ok bool) {
-	registryParts := registryPartRe.FindAllStringSubmatch(image, -1)
+func ParseRegistry(registry string) (accountId, awsEcrRegion string, ok bool) {
+	registryParts := registryPartRe.FindAllStringSubmatch(registry, -1)
 	if len(registryParts) < 1 || len(registryParts[0]) < 3 {
 		return "", "", false
 	}
@@ -108,11 +108,11 @@ func (c *Client) getLoginAuth(accountId, awsEcrRegion string) (authn.AuthConfig,
 
 // Login attempts to get the authentication material for ECR. It extracts
 // the account and region information from the image URI. The caller can ensure
-// that the passed image is a valid ECR image using ParseImage().
+// that the passed image is a valid ECR image using ParseRegistry().
 func (c *Client) Login(ctx context.Context, autoLogin bool, image string) (authn.Authenticator, error) {
 	if autoLogin {
 		ctrl.LoggerFrom(ctx).Info("logging in to AWS ECR for " + image)
-		accountId, awsEcrRegion, ok := ParseImage(image)
+		accountId, awsEcrRegion, ok := ParseRegistry(image)
 		if !ok {
 			return nil, errors.New("failed to parse AWS ECR image, invalid ECR image")
 		}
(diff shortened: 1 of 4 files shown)


oci/auth/aws/auth.go
```diff
@@ -34,14 +34,18 @@ import (
 )
 
 var registryPartRe = regexp.MustCompile(`([0-9+]*).dkr.ecr.([^/.]*)\.(amazonaws\.com[.cn]*)/([^:]+):?(.*)`)
+var registryPartReNoRepo = regexp.MustCompile(`([0-9+]*).dkr.ecr.([^/.]*)\.(amazonaws\.com[.cn]*)`)
```
souleb, line 37:
```suggestion
var fullImagePathRe = regexp.MustCompile(`([0-9+]*).dkr.ecr.([^/.]*)\.(amazonaws\.com[.cn]*)/([^:]+):?(.*)`)
var registryPartRe = regexp.MustCompile(`([0-9+]*).dkr.ecr.([^/.]*)\.(amazonaws\.com[.cn]*)`)
```

bnsfrt: I like that change

oci/auth/aws/auth.go
```diff
@@ -33,15 +33,19 @@ import (
...
 // otherwise empty strings and `false`.
 func ParseImage(image string) (accountId, awsEcrRegion string, ok bool) {
-	registryParts := registryPartRe.FindAllStringSubmatch(image, -1)
+	registryParts := fullImagePathRe.FindAllStringSubmatch(image, -1)
 	if len(registryParts) < 1 || len(registryParts[0]) < 3 {
-		return "", "", false
+		registryParts = registryPartRe.FindAllStringSubmatch(image, -1)
+		if len(registryParts) < 1 || len(registryParts[0]) < 3 {
+			return "", "", false
+		}
```
darkowlzz, line 48: I went on to find some background on why this function parses image and not registry like it's done for GCP and Azure.
The image parsing and the regex was introduced in https://github.com/fluxcd/image-reflector-controller/pull/174 , which later was refactored into a package and moved to this repository.
Based on the above PR, I don't see any specific reason to have it different from the other providers. For GCP and Azure, only the registry host is checked to be valid. We should be able to do something similar for AWS.
I think the reason for the name to be specific to image is that it was added in image-reflector-controller, where the input is expected to be an image. Now that this is becoming more generic, we should be able to relax the parsing and make it about registry only. The information that's extracted after parsing are the account ID and region, which are still present in a registry URL.
Based on the above reasoning, I think we should rename the function to `ParseRegistry()`, then replace the old regex with the new ECR registry pattern, update login.go to use `aws.ParseRegistry()` and anywhere else where it's used, and update the old tests accordingly.
I think this change would be backwards compatible because we are relaxing the parser, shouldn't break any existing behavior.

bnsfrt: Thanks for digging up that background, @darkowlzz . I went ahead and tried refactoring `ParseImage` into `ParseRegistry` and relaxed the parsing. I believe I've gotten all the pieces you've mentioned and the updated tests are passing locally. Would you give it another look?

darkowlzz: Thanks for updating. The changes look good to me. I'll run the integration tests with cloud providers to make sure it still works.

darkowlzz: I ran the integration tests in this repo and in image-reflector-controller against AWS with these changes and they continue to work.

darkowlzz: We are about to do a release today. So, I'll just rebase and squash the commits myself. Hope that's okay. The new source-controller in this release will have this fix.

bnsfrt: Thanks!

austinorth:
This fix did not work on our end, for the record, after https://github.com/bnsfrt/fluxcd-pkg/commit/f7c66eb06aa7810e03f106131306574d192314c3. Still unable to pull Helm charts from root ECR repo.

darkowlzz:
@austinorth can you provide some more information? Error string, registry address, etc. Because this change also updates the test case for the code that was causing the failure for the initially reported issue.
A new issue with all the necessary details about what you tried, what you observed, version of flux controllers, etc would also be very helpful.

cep21:
@austinorth @darkowlzz 
We are in the same boat: Cannot pull from private ECR repositories for helm charts.

darkowlzz:
@cep21 you seem to have a different issue in https://github.com/fluxcd/source-controller/issues/951 . This fix was for invalid chart reference, which you did had initially in your issue. But now, you've authentication issue, which is not related to this PR.
You may have to follow the registry login IRSA setup instructions for using service account https://fluxcd.io/flux/components/source/ocirepositories/#aws . And also restart source-controller after that to ensure it's using the new token.

romitzz1:
@darkowlzz I added opened #429 to add context that @austinorth  is pointing out.

PR #365, opened 2022-09-29
disable cosign verification in CI

darkowlzz:
Refer the build issues in https://github.com/fluxcd/source-controller/issues/899

PR #366, opened 2022-09-29
build: disable cosign verification for CI

pjbgf:
The libgit2 libraries are downloaded and verified before some of the make targets are executed. This assures the provenance of such files before using them and is very important specially for end users running such tests on their machines.


pjbgf:
Closing in favour of https://github.com/fluxcd/pkg/pull/365.

PR #368, opened 2022-10-05
build: Bump golang-with-libgit2 to v0.4.0

pjbgf:
The new version uses libgit2 1.5.0 and requires git2go/v34.


stefanprodan (approved):

pjbgf:
This has been tested for two days now (based on this change) on my setup with no regression and no new issues, so merging this PR.

The upgrade to libgit `1.5.0` was tested for over 10 days by @aryan9600 as well.

PR #369, opened 2022-10-06
git/libgit2: log context closed message as info instead of err

aryan9600:


stefanprodan (approved):

stefanprodan:
@aryan9600 see the CI failures:
```
cannot use ctx.Err() (value of type error) as type string in argument to t.logger.V(traceLevel).Info
```

pjbgf (approved):
LGTM

PR #370, opened 2022-10-07
Add Kubernetes account impersonation helpers

stefanprodan:
Move the KubeConfig API field to `apis/meta` and the [impersonation](https://github.com/fluxcd/kustomize-controller/blob/v0.29.0/controllers/kustomization_impersonation.go) helpers from kustomize-controller to `runtime/client`.


runtime/client/impersonator.go
```diff
@@ -0,0 +1,206 @@
...
+limitations under the License.
+*/
+
+package client
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/fluxcd/pkg/apis/meta"
```
hiddeco, line 23: We tend to put this in a separate import group at the end.

hiddeco (approved):
As I know where this bit of code comes from, I am happy to have this included as-is. But I think an issue should be back filled to add proper test coverage, as I am not at ease with `runtime/client` growing any further without.

PR #373, opened 2022-10-07
feat: accept files in oci build

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

_https://github.com/fluxcd/flux2/discussions/3176_

/cc @stefanprodan

### DIFF
--- a/oci/client/build.go
+++ b/oci/client/build.go
@@ -33,7 +33,7 @@ import (
 // Build archives the given directory as a tarball to the given local path.
 // While archiving, any environment specific data (for example, the user and group name) is stripped from file headers.
 func (c *Client) Build(artifactPath, sourceDir string, ignorePaths []string) (err error) {
-	if f, err := os.Stat(sourceDir); os.IsNotExist(err) || !f.IsDir() {
+	if _, err := os.Stat(sourceDir); os.IsNotExist(err) {
 		return fmt.Errorf("invalid source dir path: %s", sourceDir)
 	}
 
--- a/oci/client/build_test.go
+++ b/oci/client/build_test.go
@@ -92,6 +92,41 @@ func TestBuild(t *testing.T) {
 	}
 }
 
+// test only one file exists
+func TestBuildOneFile(t *testing.T) {
+	c := NewLocalClient()
+	g := NewWithT(t)
+
+	tmpDir := t.TempDir()
+	artifactPath := filepath.Join(tmpDir, "files.tar.gz")
+
+	sourceDir := "testdata/artifact"
+	sourceFile := filepath.Join(sourceDir, "/deployment.yaml")
+
+	err := c.Build(artifactPath, sourceFile, []string{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = os.Stat(artifactPath)
+	g.Expect(err).ToNot(HaveOccurred())
+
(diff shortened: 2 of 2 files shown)


oci/client/build.go
```diff
@@ -32,12 +31,30 @@ import (
...
+		if err != nil {
+			return err
+		}
+
+		err = os.WriteFile(filepath.Join(tmp, filepath.Base(source)), f, 0644)
+		if err != nil {
+			return err
+		}
+
+		source = tmp
```
stefanprodan, line 54: Wouldn't this make the tar have the whole tmp dir inside? I don't think users expect to have `tmp/xxxx/xxxxx/oci_copy/some.yaml` in their artifacts.

developer-guy: I've changed the code, is that what you want?

developer-guy: I've used "." as a value in "source" parameter.

oci/client/build_test.go
```diff
@@ -61,6 +61,11 @@ func TestBuild(t *testing.T) {
 			ignorePath: []string{"/*", "!/internal"},
 			checkPaths: []string{"/testdata", "!internal/", "build.go", "meta.go"},
 		},
+		{
+			name:       "existing file",
+			path:       "testdata/artifact/deployment.yaml",
+			ignorePath: []string{"ignore.txt", "ignore-dir/", "!/deploy", "somedir/git"},
```
stefanprodan, line 67:
```suggestion
			ignorePath: []string{"ignore.txt", "ignore-dir/", "!/deploy", "somedir/git"},
			checkPaths: []string{"deployment.yaml"},
```

stefanprodan (commented):
To actually test that the file is included the tarball

oci/client/build.go
stefanprodan: I guess now the file is ignored and tarball will contain the current dir `.`. IMO we need to cd into tmp.

oci/client/build.go
```diff
@@ -32,12 +31,34 @@ import (
...
+		if err != nil {
+			return err
+		}
+
+		err = os.WriteFile(filepath.Join(tmp, filepath.Base(source)), f, 0644)
+		if err != nil {
+			return err
+		}
+
+		if err := os.Rename(filepath.Join(tmp, filepath.Base(source)), filepath.Join("./", source)); err != nil {
```
stefanprodan, line 54: Are we copying the file to temp and then we move it from tmp to current directory? Instead of doing this let's just feed the file stream to the tar one please.

oci/client/build_test.go
```diff
@@ -61,6 +61,12 @@ func TestBuild(t *testing.T) {
 			ignorePath: []string{"/*", "!/internal"},
 			checkPaths: []string{"/testdata", "!internal/", "build.go", "meta.go"},
 		},
+		{
+			name:       "existing file",
+			path:       "testdata/artifact",
```
stefanprodan, line 66: This doesn’t look like it tests the new feature. We need to pass a file and check that only that one is in the tarball

developer-guy: According to that, I've updated the `checkPathExists` function to be able to work with files in addition to directories

stefanprodan (approved):
LGTM

Thanks @developer-guy 🎖️

PR #377, opened 2022-10-13
tar: Optimise memory usage

pjbgf:
The previous untar logic would take into memory each tar file entry,
causing large files to increase the likelihood of concurrent
reconciliations to cause OOM.

This change forces the untar process to use a buffer when processing
tar file entries, causing the memory profile per operation to be
reduced. On benchmarks below, the memory consumption per operation
was reduced from 10mb to 0.24mb:

```
Benchmark_Untar-16        6  188662186 ns/op  10962682 B/op  19836 allocs/op
Benchmark_Untar_Limit-16  8  131385484 ns/op   1964667 B/op  19274 allocs/op
```
The biggest file within the tar file was 95mb in size.

Testing the changes on `kustomize-controller` by tracking memory consumption, the instances with the new changes (blue and green) peak at 29mb whilst the existing code (yellow) peaks at 75mb:
image

This PR also decommissions the `untar` package, so that the code does not need to be duplicated.

Should decrease the likelihood of: https://github.com/fluxcd/kustomize-controller/issues/725

### DIFF
--- a/oci/client/build_test.go
+++ b/oci/client/build_test.go
@@ -23,9 +23,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/fluxcd/pkg/tar"
 	. "github.com/onsi/gomega"
-
-	"github.com/fluxcd/pkg/untar"
 )
 
 func TestBuild(t *testing.T) {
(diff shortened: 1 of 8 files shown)


stefanprodan:
We have the same issue in https://github.com/fluxcd/pkg/blob/main/http/fetch/archive_fetcher.go for large tarballs it runs into OOM during checksum before untar is called.

stefanprodan (approved):
LGTM

Thanks @pjbgf

hiddeco (approved):
Great work @pjbgf :bow:

PR #379, opened 2022-10-14
runtime/patch: Add SerialPatcher

darkowlzz:
SerialPatcher is a wrapper around the patch helper to help perform consecutive patching of object in a transparent manner without the need to keep track of the object state after patching.

The SerialPatcher persists a before state of the object and a kubernetes client. The patch operation configurations aren't persisted and are accepted as argument to SerialPatcher.Patch(). This makes the patcher reusable in different scenarios during its lifetime.

### DIFF
--- a/runtime/patch/serial.go
+++ b/runtime/patch/serial.go
@@ -1,0 +1,60 @@
+/*
+Copyright 2022 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

Thanks @darkowlzz 🏅

hiddeco:
This looks good to me as well, thanks for taking my suggestions into account around API simplification and caller responsibilities. 🙇

hiddeco (approved):

PR #383, opened 2022-10-19
git: Bump go-git to enable `multi_ack`

pjbgf:
The upstream `go-git` project made API changes around handling SSH HostKey callbacks, this change accommodates for such changes.

This has been tested as part of https://github.com/fluxcd/go-git/pull/4.


git/gogit/go.mod
```diff
@@ -8,6 +8,7 @@ replace (
 	github.com/fluxcd/pkg/gitutil => ../../gitutil
 	github.com/fluxcd/pkg/ssh => ../../ssh
 	github.com/fluxcd/pkg/version => ../../version
+	github.com/go-git/go-git/v5 => github.com/fluxcd/go-git/v5 v5.0.0-20221019082644-b33dae782906
```
makkes, line 11: why do we need a replace here?

pjbgf: I removed it from the internal replace block and made a standalone replace instead.

In terms of why we need it here, we need to get the latest gogit here so we can accommodate the new API changes and have them tested at pkg level.

stefanprodan:
@pjbgf does this means go-git can now be used with Azure DevOps and AWS CodeCommit? Or just CodeCommit?

pjbgf:
@stefanprodan this works with all providers I tested, which includes the 3 aforementioned plus a few others. For more info refer to:
https://github.com/fluxcd/go-git/pull/4

stefanprodan (approved):
LGTM

PR #384, opened 2022-10-20
oci: Use upstream layer digest for artifact diff

developer-guy:
**_Contributor details_**:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>
Co-authored-by: Furkan Türkal <furkan.turkal@trendyol.com>

**_Features that are included in the PR_**:
- [x] use digest extracted directly from an OCI artifact instead of recalculating it through its contents
- [x] add size control to check whether the size between the OCI artifact layer and the local file or directory is equal
- [x] check whether the error message is the expected message.

/cc @stefanprodan

### DIFF
--- a/oci/client/diff.go
+++ b/oci/client/diff.go
@@ -17,9 +17,9 @@ limitations under the License.
 package client
 
 import (
-	"bytes"
 	"context"
 	"crypto/sha256"
+	"encoding/hex"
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

Thanks @developer-guy 🏅

PR #386, opened 2022-10-24
gogit: Implement osfs to better support git clones

pjbgf:
The upstream implementation has some limitations which causes Git worktrees to be dirty soon after a clone, as they change the contents of symlink and filemodes. Here's a list of behaviour changes:
- Chroot is not supported and paths are not changed from the underlying OS fs.
- Relative paths are forced to descend from the working dir.
- Symlinks don't have its targets modified, and therefore can point to locations
outside the working dir or to non-existent paths.
- OpenFile honours the FileMode passed as argument.
- ReadLink and Lstat does not follow symlinks as most other funcs do.
However, it ensures that:
a) The filename is located within the current dir.
b) The dir in which filename is based, is located within the current dir.

Supersedes https://github.com/fluxcd/pkg/pull/385.


stefanprodan (commented):
Please move the `fs` inside the git package

stefanprodan (approved):
LGTM

Thanks @pjbgf

PR #389, opened 2022-11-02
gogit: Optimise force push

pjbgf:
When force push is enabled, we always override the push branch therefore references for the target branch are redundant.
By short-circuting the process, we avoid a fetch operation against the server, which in this case has no impact on the end result.

For servers that require `multi_ack`, this operation was leading to a silenced error.

### DIFF
--- a/git/gogit/repository_client.go
+++ b/git/gogit/repository_client.go
@@ -311,14 +311,23 @@ func (g *Client) SwitchBranch(ctx context.Context, branchName string) error {
 		return fmt.Errorf("could not checkout to branch '%s': %w", branchName, err)
 	}
 
-	g.repository.FetchContext(ctx, &extgogit.FetchOptions{
+	// When force push is enabled, we always override the push branch.
+	// No need to fetch additional refs from that branch.
+	if g.forcePush {
+		return nil
+	}
+
+	err = g.repository.FetchContext(ctx, &extgogit.FetchOptions{
 		RemoteName: extgogit.DefaultRemoteName,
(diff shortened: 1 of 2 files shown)


git/gogit/repository_client.go
```diff
@@ -311,14 +311,23 @@ func (g *Client) SwitchBranch(ctx context.Context, branchName string) error {
...
+
+	err = g.repository.FetchContext(ctx, &extgogit.FetchOptions{
 		RemoteName: extgogit.DefaultRemoteName,
 		RefSpecs: []config.RefSpec{
 			config.RefSpec(fmt.Sprintf("+refs/heads/%s:refs/remotes/%s/%[1]s", branchName, extgogit.DefaultRemoteName)),
 		},
 		Auth: authMethod,
 	})
-	ref, err := g.repository.Reference(plumbing.ReferenceName(fmt.Sprintf("/refs/remotes/origin/%s", branchName)), true)
+	if err != nil && err != extgogit.NoErrAlreadyUpToDate && !errors.Is(err, extgogit.NoMatchingRefSpecError{}) {
```
hiddeco, line 327:
```suggestion
	if err != nil && !errors.Is(err, extgogit.NoErrAlreadyUpToDate) && !errors.Is(err, extgogit.NoMatchingRefSpecError{}) {
```

is preferable over simple checks, as documented in https://pkg.go.dev/errors#pkg-overview

hiddeco (approved):
Looks good to me, thanks @pjbgf :bow:

PR #390, opened 2022-11-07
Introduce the Event v1beta1 API schema

stefanprodan:
The `fluxcd/pkg/apis/event` package is for formalising how eventing works in Flux. Once this package is released, the `fluxcd/pkg/runtime/events` and all controllers should switch to it. The Event API comes with helper functions to be used in notification-controller for filtering metadata.

To be used in: https://github.com/fluxcd/notification-controller/pull/435


apis/event/v1beta1/event.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// ID of the controller instance, e.g. `source-controller-xyzf`.
+	// +optional
+	ReportingInstance string `json:"reportingInstance,omitempty"`
+}
+
+// IsProgressing returns true if the event reason is Progressing.
+func (in *Event) IsProgressing() bool {
+	return in.Reason == "Progressing"
+}
```
hiddeco, line 85:
```suggestion
// HasReason returns true if the Reason equals the given value.
func (in *Event) HasReason(reason string) bool {
	return in.Reason == reason
}
```

would make the API more versatile and less constrained to implementation details.

apis/event/v1beta1/event.go
```diff
@@ -0,0 +1,141 @@
...
+	// EventSeverityInfo represents an informational event, usually
+	// informing about changes.
+	EventSeverityInfo string = "info"
+	// EventSeverityError represent an error event, usually a warning
+	// that something goes wrong.
+	EventSeverityError string = "error"
+)
+
+// EventTypeTrace represents a trace event.
+const EventTypeTrace string = "Trace"
```
hiddeco, line 41: What's the difference between a "trace" event type and a "trace" severity?

apis/event/v1beta1/event.go
```diff
@@ -0,0 +1,141 @@
...
+// IsProgressing returns true if the event reason is Progressing.
+func (in *Event) IsProgressing() bool {
+	return in.Reason == "Progressing"
+}
+
+// StripMetadata removes metadata entries which are not user-facing
+// and trims the API group from the metadata keys.
+func (in *Event) StripMetadata() {
+	group := in.InvolvedObject.GetObjectKind().GroupVersionKind().Group
+	excludeList := []string{fmt.Sprintf("%s/checksum", group)}
```
hiddeco, line 91: Would expect `[%s/]checksum` to be exposed as part of the contract. In addition, with RFC-0005 landing soon, we would probably need a `%s/digest` as well.

In extend to this, I am wondering if we can settle on all relevant annotations being prefixed with the GVK domain. As this would be more inline with Kubernetes annotations in general.

apis/event/v1beta1/event.go
```diff
@@ -0,0 +1,141 @@
...
+		// Filter other meta based on group prefix, while filtering out excludes
+		for key, val := range in.Metadata {
+			if strings.HasPrefix(key, group) && !inList(excludeList, key) {
+				newKey := strings.TrimPrefix(key, fmt.Sprintf("%s/", group))
+				meta[newKey] = val
+			}
+		}
+	}
+
+	in.Metadata = meta
```
hiddeco, line 105: What would be the intend of the caller / why are we directly mutating the Event?

apis/event/v1beta1/event.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// IsCommitStatusUpdate returns true if the event is meant for commit status updates.
+func (in *Event) IsCommitStatusUpdate() bool {
+	if val, ok := in.Metadata["commit_status"]; ok && val == "update" {
```
hiddeco, line 110: Instead of creating a dedicated method for this, and pulling in implementation specific details. I would rather expose the `commit_status` and offer a generic key/value checker.

hiddeco: In addition to this, I would rename `commit_status` to `commitStatus` (leaving backwards compatibility details as an exercise), which is much more inline with both Kubernetes standards and JSON camelCase which the world has settled on.

apis/event/v1beta1/event.go
```diff
@@ -0,0 +1,141 @@
...
+		return true
+	}
+	return false
+}
+
+// GetCommitID extracts the Git commit value from the revision string.
+func (in *Event) GetCommitID() (string, error) {
```
hiddeco, line 117: Like with `IsCommitStatusUpdate`, I would make a generic method to retrieve a value from the map and leave the parsing as an exercise for something else. Current logic also conflicts with RFC-0005.

hiddeco (commented):
By exposing the annotation keys in combination with more generic methods, I think the API would become more versatile and resistant to implementation changes within the domains of the producer and/or consumer.

apis/event/v1beta1/event.go
stefanprodan: This PR doesn't try to address the RFC. This PR is intended to be first step towards Event consolidation while keeping things as they are. After RFC-0005 is merged, we could work on v1beta2 and introduce there all these changes.

apis/event/v1beta1/event.go
stefanprodan: I'm not for introducing any breaking changes in v1beta1. This PR is intended to be first step towards Event consolidation while keeping things as they are

apis/event/v1beta1/event.go
hiddeco: Following my suggestion would make this much easier?

apis/event/v1beta1/event.go
stefanprodan: The prefix borks the alerts, as the keys end up as Stack/Discord/etc labels, that's why we need this strip function.

apis/event/v1beta1/event.go
stefanprodan: It's used here for mapping https://github.com/fluxcd/pkg/blob/main/runtime/events/recorder.go#L229

hiddeco:
Talked more about this with @stefanprodan. The goal of this PR for now is to at least better document some of the indirect contracts that have been established in the notification-controller around e.g. expected "revision" and "checksum" annotation keys, in combination with generic API utilities to retrieve the values of said annotations.

Not within the scope, but discussed to be followed up at a later time are at least the following things:

1. Settle on GVK prefixing metadata keys (and how they are filtered out by e.g. notification-controller).
2. Addressing the "Trace" event type. This type has no value to any arbitrary event recorder implementations, while it indirectly does equal a "Normal" event type (once it goes past our custom recorder implementation). Making the "Flux event" an imperfect and/or awkward first class citizen.

   One idea would be to make the filtering happen based on an annotation on the K8s event (e.g. `group.version.domain/trace: true`). Which would both result in reflection of the filter itself on the Kubernetes Event, as it allowing to co-exist with arbitrary recorder implementations.
3. Dealing with RFC-0005 implications, which will change the revision format and introduce a digest.
4. Further breaking changes which may result in a better future API.

hiddeco (approved):
Happy with the scoped outcome of this, thanks @stefanprodan :bow:

PR #392, opened 2022-11-07
Clean-up Kustomize module and enable dry-run for varsub

souleb:
The goals of this PR are:
1. make sure we use the same `envsubst` version as in `kustomize-controller`
2. introduce `securefs`
3. enable `dry-run` when performing variable substitution. This means that `in-line` vars are substituted, but we do not allow connection to the cluster to retrieve resources.

1 & 2 will enable using this module from both `flux cli` and `kustomize controller`. This will make it easier to maintain and avoid drift i.e. `flux build kustomize` and `kustomize-controller` should have same behavior.

3 Will enable `flux build kustomize` to run without having to connect to the cluster.


.gitignore
```diff
@@ -17,3 +17,5 @@
 build/
 bin/
 testbin/
+
+**/.DS_Store
```
stefanprodan, line 21: Please undo this, and use your machine .gitignore

souleb:
I have not moved `SubstituteReference` to `pkg/apis/meta` because  there are kubebuilder markers used for validation on it in `kustomize-controller`.

stefanprodan (approved):
LGTM

Thanks @souleb 🏅

PR #393, opened 2022-11-08
git: add support for cloning empty repos

aryan9600:
Add support for cloning empty repos using `go-git` and `libgit2`.
`go-git` returns an error when cloning an empty repository directly, which is handled by using the advice provided here: https://github.com/go-git/go-git/issues/118#issuecomment-773488577. `libgit2` doesn't return an error and thus the change is much simpler.
The `go-git` package gets test coverage for this scenario but the `libgit2` package does not because of the following weird behavior:
```go
_, err := git2go.InitRepository(filepath.Join(server.Root(), emptyRepoPath), true)

err = l.remote.Fetch([]string{branch},
	&git2go.FetchOptions{
		DownloadTags:    git2go.DownloadTagsNone,
		RemoteCallbacks: remoteCallBacks,
	},
	"")
if err != nil {
	return nil, fmt.Errorf("unable to fetch remote '%s': %w", url, gitutil.LibGit2Error(err))
}

isEmpty, err := l.repository.IsEmpty()
fmt.Println(isEmpty) // prints false
```
To make sure the changes to the `libgit2` package have the intended effect, it was tested (and confirmed to have the correct behavior) using empty repos on GitHub and GitLab.(same goes for the `go-git` package as well).

### DIFF
--- a/git/libgit2/clone.go
+++ b/git/libgit2/clone.go
@@ -79,6 +79,14 @@ func (l *Client) cloneBranch(ctx context.Context, url, branch string, opts git.C
 		return nil, fmt.Errorf("unable to fetch remote '%s': %w", url, gitutil.LibGit2Error(err))
 	}
 
+	isEmpty, err := l.repository.IsEmpty()
+	if err != nil {
+		return nil, fmt.Errorf("unable to check if cloned repo '%s' is empty: %w", url, err)
+	}
+	if isEmpty {
+		return nil, nil
+	}
+
 	branchRef, err := l.repository.References.Lookup(fmt.Sprintf("refs/remotes/origin/%s", branch))
 	if err != nil {
 		return nil, fmt.Errorf("unable to lookup branch '%s' for '%s': %w", branch, url, gitutil.LibGit2Error(err))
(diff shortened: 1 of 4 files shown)


git/libgit2/clone.go
```diff
@@ -79,6 +79,14 @@ func (l *Client) cloneBranch(ctx context.Context, url, branch string, opts git.C
 		return nil, fmt.Errorf("unable to fetch remote '%s': %w", url, gitutil.LibGit2Error(err))
 	}
 
+	isEmpty, err := l.repository.IsEmpty()
```
pjbgf, line 82: Can we add a test case for this as well?

aryan9600: kindly see the PR description regarding this

pjbgf: We could still have a test case for it, which has a different expectation from GoGit. The key point here was to have something to identify regression in case changes are made here and empty clones would not break again.

As we may decomission the entire libgit2 implementation by Bootstrap GA, we can park this.

pjbgf (approved):
LGTM

thanks @aryan9600! please rebase.

stefanprodan (approved):

PR #394, opened 2022-11-08
git/gogit: configure hostkey callback only when known_hosts is provided

aryan9600:
Some applications like CLIs might prefer to use the known_hosts of the machine which then leads to an empty known_hosts value being used in `git.AuthOptions` resulting in an invalid callback.


pjbgf (approved):
LGTM

Thanks @aryan9600! :bow:

stefanprodan (approved):

PR #396, opened 2022-11-08
git: Add new WithInsecureCredentialsOverHTTP

pjbgf:
Blocks the transmission of credentials over HTTP by default but add a new `WithInsecureCredentialsOverHTTP` which allows consumers to re-enable it.

Passing credentials in plain-text over HTTP is not recommended for production environments.

### DIFF
--- a/git/gogit/clone_test.go
+++ b/git/gogit/clone_test.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


stefanprodan (approved):

git/gogit/clone_test.go
```diff
@@ -896,6 +898,189 @@ func Test_getRemoteHEAD(t *testing.T) {
...
 }
 
+func TestClone_CredentialsOverHttp(t *testing.T) {
+	tests := []struct {
+		name                     string
+		username                 string
+		password                 string
+		allowCredentialsOverHttp bool
+		transformURL             func(string) string
+		expectNewErr             string
```
aryan9600, line 908: This field is not specified in the any of the test cases below.

git/gogit/clone_test.go
```diff
@@ -896,6 +898,189 @@ func Test_getRemoteHEAD(t *testing.T) {
...
+				Password:  tt.password,
+			}, opts...)
+
+			if tt.expectNewErr != "" {
+				g.Expect(err).To(HaveOccurred())
+				g.Expect(err.Error()).To(ContainSubstring(tt.expectNewErr))
+				return
+			} else {
+				g.Expect(err).ToNot(HaveOccurred())
+			}
```
aryan9600, line 1059:
```suggestion
		    g.Expect(err).ToNot(HaveOccurred())
```

aryan9600 (changes requested):

git/gogit/repository_client.go
```diff
@@ -183,6 +196,29 @@ func (g *Client) Clone(ctx context.Context, url string, cloneOpts git.CloneOptio
...
+	if g.credentialsOverHTTP {
+		return nil
+	}
+
+	ru, err := url.Parse(u)
+	if err != nil {
+		return fmt.Errorf("cannot parse url: %w", err)
+	}
+
+	httpOrEmpty := ru.Scheme == string(git.HTTP) || ru.Scheme == ""
```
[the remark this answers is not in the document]
pjbgf: Testing in libgit2, empty scheme yield unsupported protocol. In go-git, their `NewEndpoint` logic would assume `scheme` as `file`, for an input such as `github.com/user/repository.git`.

From a Flux perspective, the scheme is enforced via the CRD validation, so we should never get this empty. Either way, I would rather fail safe and error.

git/gogit/clone_test.go
pjbgf: That was due to change where the validation was occurring, this is now fixed in both libgit2 and gogit.

aryan9600 (approved):
lgtm! 🚀

PR #404, opened 2022-11-14
git: change Commit#String format

hiddeco:
This changes the string format of `Commit` to use an `@` separator instead of `/`, and drops the usage of "HEAD" as a virtual named reference for commits without a named pointer (e.g. a Git branch and/or tag). Matching the (revision) format proposed in RFC-0005.


pjbgf (approved):
LGTM

Great job @hiddeco! :bow:

hiddeco:
Given we are also optimizing some other bits of the Git package at present to facilitate the removal of libgit2 and address other issues. I suggest we keep this on hold until I have my proposed changes incorporating RFC-0005 for source-controller ready.

stefanprodan (approved):
LGTM

PR #410, opened 2022-11-16
ci: fix git e2e tests workflow

aryan9600:


pjbgf:
The other 2 e2e tests are still being executed when not done via fork PR, as intended as per: https://github.com/fluxcd/pkg/actions/runs/3480017008

PR #412, opened 2022-11-18
Add support for Kustomize components

klausenbusk:
Ref: fluxcd/kustomize-controller#753

Signed-off-by: Kristian Klausen <kristian@klausen.dk>

### DIFF
--- a/kustomize/kustomize_generator.go
+++ b/kustomize/kustomize_generator.go
@@ -45,6 +45,7 @@ const (
 	specField                 = "spec"
 	targetNSField             = "targetNamespace"
 	patchesField              = "patches"
+	componentsField           = "components"
 	patchesSMField            = "patchesStrategicMerge"
 	patchesJson6902Field      = "patchesJson6902"
 	imagesField               = "images"
@@ -154,6 +155,19 @@ func (g *Generator) WriteFile(dirPath string, opts ...SavingOptions) (Action, er
 			Patch:  p.Patch,
 			Target: adaptSelector(&p.Target),
 		})
+	}
+
+	components, _, err := g.getNestedStringSlice(specField, componentsField)
+	if err != nil {
+		errf := CleanDirectory(dirPath, action)
+		return action, fmt.Errorf("unable to get components: %w", fmt.Errorf("%v %v", err, errf))
+	}
+
+	for _, component := range components {
+		if !IsLocalRelativePath(component) {
+			return "", fmt.Errorf("component path '%s' must be local and relative", component)
+		}
+		kus.Components = append(kus.Components, component)
 	}
 
 	patchesSM, err := g.getPatchesStrategicMerge()
@@ -356,6 +370,15 @@ func (g *Generator) getNestedString(fields ...string) (string, bool, error) {
 	val, ok, err := unstructured.NestedString(g.kustomization.Object, fields...)
 	if err != nil {
 		return "", ok, err
+	}
+
+	return val, ok, nil
+}
+
+func (g *Generator) getNestedStringSlice(fields ...string) ([]string, bool, error) {
+	val, ok, err := unstructured.NestedStringSlice(g.kustomization.Object, fields...)
+	if err != nil {
+		return []string{}, ok, err
 	}
 
 	return val, ok, nil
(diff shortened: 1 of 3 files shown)


stefanprodan:
Hey @klausenbusk can you please add here the tests you wrote.

klausenbusk:
> Hey @klausenbusk can you please add here the tests you wrote.

I have added the test back. I'm not sure if we want to use `interface{}`? If we want to use `kustypes.Kustomization` we need to do some PR juggling.

stefanprodan:
We can't use the API types in pkg, that would get us to cyclic dependency between the controller and pkg which that would break go.mod.

klausenbusk:
I'm not sure the fuzz failure is related to this change:

```
+ cd addimport
/src/build.sh: line 38: cd: addimport: No such file or directory
+ cleanup
```

stefanprodan:
@klausenbusk we can ignore the fuzzers, these were broken upstream and we're going to fix them in a different PR.

kustomize/kustomize_generator_test.go
```diff
@@ -117,3 +120,53 @@ func Test_SecureBuild_rel_basedir(t *testing.T) {
...
+			fluxComponents:     []string{"componentB", "componentC"},
+			expectedComponents: []interface{}{"componentA", "componentB", "componentC"},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g := NewWithT(t)
+			tmpDir := t.TempDir()
+			g.Expect(copy.Copy("./testdata/components", tmpDir)).To(Succeed())
```
souleb, line 155: My understanding here is that you never copy ""./testdata/zero-components", and the `kustomization.yaml` for the second test-case is always generated. If that's correct, then do you need the test sub-repo?

kustomize/kustomize_generator_test.go
```diff
@@ -117,3 +120,53 @@ func Test_SecureBuild_rel_basedir(t *testing.T) {
...
+		name               string
+		dir                string
+		fluxComponents     []string
+		expectedComponents []interface{}
+	}{
+		{
+			name:               "test kustomization.yaml with components and Flux Kustomization without components",
+			dir:                "components",
+			fluxComponents:     []string{},
+			expectedComponents: []interface{}{"componentA"},
```
souleb, line 135: since go 1.18 you can replace `interface{}` with `any`.

kustomize/kustomize_generator.go
```diff
@@ -591,3 +614,21 @@ func copyFile(src, dst string) (err error) {
 
 	return
 }
+
+func isLocalRelativePath(path string) bool {
```
souleb, line 618: It would be nice to have a test for this function.

stefanprodan:
Fuzzing has been fixed in main, please rebase.

kustomize/kustomize_generator.go
klausenbusk: Added. the function is now public because it can't be accessed from the test package otherwise.

kustomize/kustomize_generator_test.go
klausenbusk: `./testdata/components` is copied and it has two directories: `components` and `zero-components`.

I have refactored the test so `zero-components` isn't needed, now it just uses the empty `tmpDir` as root if `dir` is a empty string.

klausenbusk:
Any idea why the test is failing on macos-12 https://github.com/fluxcd/pkg/actions/runs/3539040439/jobs/5940762510?

stefanprodan:
@klausenbusk use:
```go
func testTempDir(t *testing.T) (string, error) {
	tmpDir := t.TempDir()

	tmpDir, err := filepath.EvalSymlinks(tmpDir)
	if err != nil {
		return "", fmt.Errorf("error evaluating symlink: '%w'", err)
	}

	return tmpDir, err
}
```

instead of `t.TempDir()`

souleb (approved):
LGTM. Thanks @klausenbusk for the PR 🙇‍♂️

klausenbusk:
Do you need more from me, before this can be merged? :)

pjbgf:
@klausenbusk please rebase so we can go ahead with the merge.

stefanprodan (approved):

klausenbusk:
Rebased :)

pjbgf (approved):
LGTM

Thanks @klausenbusk! :bow:

PR #413, opened 2022-11-18
oci: Bump github.com/aws/aws-sdk-go to v2

pjbgf:
Migrating into v2 we mitigate upstream CVEs that have not been back ported into the previously used version: 
- CVE-2020-8911
- CVE-2020-8912
- GHSA-7f33-f4f5-xwgw
- GHSA-f5pg-7wfw-84q9


stefanprodan:
If these CVE fixes will not be back-ported in AWS SDK v1, then we really need to get rid of v1 across all controllers. I know that kustomize-controller uses both SDKs to test backwards compatibly with SOPS, but given this maybe it's time to drop the tests and v1 from KC? cc @hiddeco

hiddeco:
Tests are not included in builds, so this should not be an issue. Lets just wait until SOPS moves and do it properly.

pjbgf:
Shall we merge these changes?

hiddeco (approved):

PR #415, opened 2022-11-23
build: Fix fuzz tests

pjbgf:
Temporarily ignores `recorder_fuzzer_test.go`.


makkes (approved):
Do we have an issue tracking re-enabling this test? Might be worth creating one if we're planning on having this disabled for longer than a couple of days.

pjbgf:
@makkes I am reviewing the entirety of our fuzz tests as part of the current issues. I will create an issue to track it all, including this.

PR #419, opened 2022-11-28
git/gogit: modify chroot to return a new os filesystem

aryan9600:
Modify `Chroot()` into returning a new OS filesystem. The filesystem returned is not actually chrooted, but only enforces all files to be under the root working directory. This is required to avoid breaking cloning repos with submodules. Ref: https://github.com/fluxcd/go-git/blob/d979ddd67ae30aebcdd285c0b0133594e7194980/storage/filesystem/dotgit/dotgit.go#L1100-L1103

Related to: https://github.com/fluxcd/source-controller/issues/970

### DIFF
--- a/git/gogit/fs/osfs_os.go
+++ b/git/gogit/fs/osfs_os.go
@@ -38,7 +38,7 @@ const (
 // OS is a fs implementation based on the OS filesystem which has some
 // changes in behaviour when compared to the upstream go-git/go-billy/v5/osfs:
 //
-// - Chroot is not supported and paths are not changed from the underlying OS fs.
+// - Chroot doesn't return a chrooted filesystem but returns a new OS filesystem.
 // - Relative paths are forced to descend from the working dir.
 // - Symlinks don't have its targets modified, and therefore can point to locations
 // outside the working dir or to non-existent paths.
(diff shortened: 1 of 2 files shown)


git/gogit/fs/osfs_os.go
```diff
@@ -38,7 +38,7 @@ const (
 // OS is a fs implementation based on the OS filesystem which has some
 // changes in behaviour when compared to the upstream go-git/go-billy/v5/osfs:
 //
-// - Chroot is not supported and paths are not changed from the underlying OS fs.
+// - Chroot doesn't return a chrooted filesystem but returns a new OS filesystem.
```
pjbgf, line 41: Maybe?
```suggestion
// - Chroot doesn't return a chrooted filesystem, instead it returns a new OS filesystem that has read and write operations confined to descending locations from the working directory.
```

git/gogit/fs/osfs_os.go
```diff
@@ -213,8 +213,9 @@ func (fs *OS) Readlink(link string) (string, error) {
 	return os.Readlink(link)
 }
 
+// We don't actually return a chrooted filesystem, just a regular OS filesystem.
```
pjbgf, line 216:
```suggestion
// Chroot returns a new OS filesystem, with working directory set to path.
```

pjbgf (approved):
LGTM

PR #421, opened 2022-12-02
git/gogit: fallback to machine known_hosts

aryan9600:
Add a new `AuthMethod` which uses the default `known_hosts`. Fallback to this when the provided auth options has an empty private key and known hosts.

Related to: https://github.com/fluxcd/flux2/issues/3358

### DIFF
--- a/git/gogit/transport.go
+++ b/git/gogit/transport.go
@@ -30,7 +30,7 @@ import (
 
 // transportAuth constructs the transport.AuthMethod for the git.Transport of
 // the given git.AuthOptions. It returns the result, or an error.
-func transportAuth(opts *git.AuthOptions) (transport.AuthMethod, error) {
+func transportAuth(opts *git.AuthOptions, fallbackToDefaultKnownHosts bool) (transport.AuthMethod, error) {
 	if opts == nil {
 		return nil, nil
 	}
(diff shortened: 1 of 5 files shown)


git/gogit/transport.go
```diff
@@ -46,6 +46,23 @@ func transportAuth(opts *git.AuthOptions) (transport.AuthMethod, error) {
...
+			}
+			pkCallback, ok := authMethod.(*ssh.PublicKeysCallback)
+			if ok {
+				return &DefaultAuth{
+					pkCallack: pkCallback,
+				}, nil
+			} else {
+				// we won't reach this branch, but just in case
+				return nil, nil
+			}
```
pjbgf, line 64:
```suggestion			
			if pkCallback, ok := authMethod.(*ssh.PublicKeysCallback); ok {
				return &DefaultAuth{
					pkCallack: pkCallback,
				}, nil
			}
			return nil, nil
```

pjbgf (changes requested):
I think we want to have a toggle to enable/disable this behaviour, as this should not take place in the controllers.

git/gogit/transport.go
```diff
@@ -46,6 +46,20 @@ func transportAuth(opts *git.AuthOptions) (transport.AuthMethod, error) {
...
+			authMethod, err := ssh.DefaultAuthBuilder(opts.Username)
+			if err != nil {
+				return nil, err
+			}
+			pkCallback, ok := authMethod.(*ssh.PublicKeysCallback)
+			if ok {
+				return &DefaultAuth{
+					pkCallack: pkCallback,
+				}, nil
+			}
```
pjbgf, line 61: We may want to highlight the abnormal flow in which `authMethod` is not convertible to `ssh.PublicKeysCallback` by returning early with an error. Falling through to the previous code seems like an invalid execution path.

aryan9600: i agree, thats why i had an `else` branch here which returned `nil, nil` which would lead to a host key error further down the stack, but i removed it at your suggestion during the previous review.

pjbgf: The suggestion removed the redundant `else`, but kept the `return` statement.

pjbgf (approved):

pjbgf:
Please rebase so we can go ahead with the merge.

stefanprodan:
Is there an issue that should be linked here?

pjbgf:
Merging after the update linking the issue in the description.

Issue #380, opened 2022-10-17
runtime/conditions/check: Remove ObservedGeneration check


stefanprodan:
When using `SerialPatcher` the `Reconciling` condition will have `ObservedGeneration=Generation` so we need to remove the following check:

```
[Check-FAIL]: The root ObservedGeneration must be less than the Reconciling condition ObservedGeneration when Reconciling condition is True
```

PR #422, opened 2022-12-05
Progressive status helper and in-progress status checker

darkowlzz:
This change introduces a helper and a checker for progressive status of objects that are undergoing reconciliation. Progressive status are the status conditions that appear on an object in the middle of a reconciliation, also referred as mid-reconciliation status in this change. `Reconciling` and `Ready` are primarily involved in progressive status to indicate that an object is undergoing reconciliation.

- When an object is reconciled for the first time, progressive status of `Reconciling=True` and `Ready=Unknown` are expected on the object.
- On subsequent reconciliation with `Ready=True` set by a previous reconciliation, the `Reconciling=True` with `Ready=True` indicates that the reconciler is reconciling an object but hasn't detected any drift to act upon.
- `Ready=Unknown` indicates that there may be some drift for which the reconciler is performing some action.

## ProgressiveStatus helper

For working with the above scenarios of progressive status, a helper  `ProgressiveStatus(drift, obj, reason, msg, msgArgs...)` is introduced. It helps mark an object `Reconciling=True` and also `Ready=Unknown` with the same reason and message if a drift is detected. If no drift is detected, only `Reconciling=True` is set.

## In-progress status checker

The progressive status conditions of an object are observed to be orthogonal to the final status of the object after a full reconciliation that's computed by various status summarization techniques. During a reconciliation, there's not enough information to perform full summarization of the status conditions. The partial information is sufficient to only report the current status of reconciliation in the form of progressive status. This results in differences in the considerations of status conditions check on an object. Due to this, the WARN and FAIL checks of the status checker have been updated to indicate if they apply to mid-reconciliation status or not. And a new check is added that's specific to mid-reconciliation status.
Due to the different sets of checks, a new in-progress status checker is introduced, which can be created with `NewInProgressChecker()`. With this, we have two separate status checkers which can be used to check the progressive status and fully reconciled status of an object.

Fixes https://github.com/fluxcd/pkg/issues/380

### DIFF
--- a/runtime/conditions/check/checker.go
+++ b/runtime/conditions/check/checker.go
@@ -34,6 +34,10 @@ type checkFunc func(ctx context.Context, obj conditions.Getter, condns *Conditio
 // Checker performs all the status checks. It is configured to provide context
 // of the target controller.
 type Checker struct {
+	// requireConditions is used to indicate that the checker requires
+	// conditions context to operate. It is used to perform validation of the
+	// checker instance.
+	requireConditions bool
 	// K8s client, to fetch the latest version of an object.
 	client.Client
 	// conditions is the conditions context of the target controller.
@@ -55,7 +59,8 @@ type Checker struct {
 	// ExcludeChecks map[string]bool
 }
 
-// NewChecker constructs and returns a new Checker for a controller.
+// NewChecker constructs and returns a new reconciled status Checker for a
+// controller.
 func NewChecker(cli client.Client, condns *Conditions) *Checker {
 	warnChecks := []checkFunc{
 		check_WARN0001,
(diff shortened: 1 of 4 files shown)


runtime/reconcile/result_test.go
```diff
@@ -677,3 +677,103 @@ func TestDetermineSuccessType(t *testing.T) {
...
+		name             string
+		drift            bool
+		reason           string
+		msg              string
+		msgArgs          []interface{}
+		beforeFunc       func(obj *testdata.Fake)
+		assertConditions []metav1.Condition
+	}{
+		{
+			name:   "Unset Reconciling and Ready, make Reconcilg=True and Ready=Unknown",
```
pjbgf, line 692:
```suggestion
			name:   "Unset Reconciling and Ready, make Reconcile=True and Ready=Unknown",
```

pjbgf (approved):
LGTM

Great stuff @darkowlzz! :bow:

runtime/conditions/check/checker.go
```diff
@@ -74,11 +79,38 @@ func NewChecker(cli client.Client, condns *Conditions) *Checker {
...
+		conditions:        condns,
+		warnChecks:        warnChecks,
+		failChecks:        failChecks,
+		Stdout:            os.Stdout,
+		Stderr:            os.Stderr,
+	}
+}
+
+// NewInProgressChecker constructs and returns a new in-progress status Checker
+// for a controller. This exists separate from NewChecker because the status
```
hiddeco, line 95:
```suggestion
// for a controller. This exists separately from NewChecker because the status
```

runtime/reconcile/result.go
```diff
@@ -182,6 +182,21 @@ func (rs ResultFinalizer) Finalize(obj conditions.Setter, res ctrl.Result, recEr
 	return recErr
 }
 
+// ProgressiveStatus helps report the progressive status of an object based on
+// the given status value and drift information.
+// It always sets Reconciling=True with the given status values.
+// Ready condition is set only when:
+// - there's no existing Ready condition in the status
+// - Ready=Unknown
+// - a drift is detected
```
hiddeco, line 191:
```suggestion
// - drift is detected
```

hiddeco (approved):
Other than the two minor commented nits, this looks good to me.

PR #430, opened 2022-12-12
git/gogit: fix chroot filesystem and add submodule test

aryan9600:
Use `filepath-securejoin` to join root working dir of the base filesystem and the desired chroot root and return a filesystem using the resulting path as the root working dir. Add a test which clones a repo with a submodule to make sure this doesn't get broken again.

Related to: https://github.com/fluxcd/source-controller/issues/970


pjbgf (approved):
LGTM

Thanks @aryan9600! :bow:

PR #432, opened 2022-12-13
oci/auth: Fix aws login with aws-sdk-go-v2

darkowlzz:
Follow-up of https://github.com/fluxcd/pkg/commit/750df674c5b58fcfb5a20a144e8f733daefa2643 , upgrade to aws-sdk-go-v2.

Without this change, the login fails with error:
> `operation error ECR: GetAuthorizationToken, https response error StatusCode: 400, RequestID: 49826b0a-de68-411d-acb1-3fd1554c64b4, api error MissingAuthenticationTokenException: Missing Authentication Token`

As per [aws-sdk-go-v2 migration docs](https://aws.github.io/aws-sdk-go-v2/docs/migrating/#migrating-from-newsession-with-awsconfig-options), new config for creating a session should use `config.LoadDefaultConfig()` and [account ID is not needed for ECR authorization token](https://github.com/aws/aws-sdk-go-v2/blob/service/ecr/v1.17.22/service/ecr/api_op_GetAuthorizationToken.go#L44).

In order to maintain the auth API compatibility, which is shared with other provider, return an empty client for `NewClient()`. The login implementation loads the default config if the client config is nil. A test can stub the client config before calling login.

Tested this against EKS using the integration tests with the EKS terraform module fix from https://github.com/fluxcd/test-infra/pull/2.


stefanprodan (approved):
LGTM 

Thanks @darkowlzz

pjbgf (approved):
LGTM

Thanks for the thread safety improvements @darkowlzz! :bow:

Issue #429, opened 2022-12-09
Unable to Pull Helm Charts from ECR Repositories


romitzz1:
This is in response to fluxcd/pkg/pull/362.  That have helm charts in the root of ECR repos, and flux is unable to pull them. 

The error code we now get:  `invalid chart reference: failed to get chart version for remote reference: could not get tags for "SERVICENAME": could not fetch tags for "oci://ACCOUNTNUMBER.dkr.ecr.REGION.amazonaws.com/SERVICENAME": GET `"https://ACCOUNTNUMBER.dkr.ecr.REGION.amazonaws.com/v2/SERVICENAME/tags/list":` credential required for basic auth`

source-controller version: `ghcr.io/fluxcd/source-controller:v0.32.1`

This is the URL of our `HelmRepository` : ` url: `oci://ACCOUNTNUMBER.dkr.ecr.REGION.amazonaws.com``

[...]

PR #434, opened 2022-12-14
oci/auth: Fix ImageRegistryProvider for repo root

darkowlzz:
When a repository root is provided, `ImageRegistryProvider()` fails to identify the registry host. The provided `name.Reference` for repository root is a `Tag` reference.
`name.ParseReference()`, assumes that that [given string is the name](https://github.com/google/go-containerregistry/blob/v0.12.1/pkg/name/repository.go#L78-L93) and [registry host being empty defaults to index.docker.io](https://github.com/google/go-containerregistry/blob/v0.12.1/pkg/name/registry.go#L118-L120).

`ImageRegistryProvider()` now takes the full address, in addition to the `name.Reference` and uses the full address for analysis if there's no "/" in the address, which indicates that it's a repository root.

This fixes login with repository root for all the providers, not just AWS.

Fixes https://github.com/fluxcd/source-controller/issues/951, https://github.com/fluxcd/pkg/issues/429 .

Tested it against EKS with OCI Helm Chart, which was failing before as reported in the above issues.

### DIFF
--- a/oci/auth/login/login.go
+++ b/oci/auth/login/login.go
@@ -18,6 +18,7 @@ package login
 
 import (
 	"context"
+	"strings"
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

Thanks @darkowlzz 🏅

souleb (approved):
LGTM.

pjbgf (approved):
LGTM

hiddeco (approved):

PR #444, opened 2023-01-04
kustomize: Handle malformed YAML in manifest scan

darkowlzz:
While scanning YAML manifests for generating kustomization, certain invalid YAML files can cause panic. Add panic recovery while scanning manifests.

Extracts the manifest scan code into a separate function and adds white-box tests for it for different scenarios.

The recovery message tries to provide some helpful hint about the bad manifest to make it easier to find and fix the issue
```
recovered from panic while parsing YAML file alert.yaml: runtime error: invalid memory address or nil pointer dereference
```

Refer https://github.com/fluxcd/flux2/issues/3458

### DIFF
--- a/kustomize/kustomize_generator.go
+++ b/kustomize/kustomize_generator.go
@@ -415,54 +415,12 @@ func (g *Generator) generateKustomization(dirPath string) (Action, string, error
 		}
 	}
 
-	scan := func(base string) ([]string, error) {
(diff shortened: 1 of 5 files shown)


souleb (approved):
LGTM! Thanks @darkowlzz

hiddeco (approved):

PR #472, opened 2023-02-15
sourceignore/oci: update dependencies

hiddeco:


hiddeco:
CI fails because the `go.mod` is still set to `go 1.18` while we optimize dependencies for compatibility with `1.19`. This will likely cause other PRs I created to fail as well, so need to do a round of adjusting.

stefanprodan (approved):

PR #490, opened 2023-02-20
oci: Allow consumers to change the media types

stefanprodan:


oci/globals.go
```diff
@@ -0,0 +1,30 @@
...
+package oci
+
+import "github.com/google/go-containerregistry/pkg/v1/types"
+
+var (
+	// ConfigMediaType is the OCI media type for the config layer.
+	ConfigMediaType types.MediaType = "application/vnd.cncf.flux.config.v1+json"
+
+	// ContentMediaType is the OCI media type for the content layer.
+	ContentMediaType types.MediaType = "application/vnd.cncf.flux.content.v1.tar+gzip"
```
hiddeco, line 26:
```suggestion
	// CanonicalConfigMediaType is the OCI media type for the config layer.
	CanonicalConfigMediaType types.MediaType = "application/vnd.cncf.flux.config.v1+json"

	// CanonicalContentMediaType is the OCI media type for the content layer.
	CanonicalContentMediaType types.MediaType = "application/vnd.cncf.flux.content.v1.tar+gzip"
```

maybe? Not sure if applicable to `UserAgent` as well.

hiddeco:
Think `/s/ssa/oci/` on top of commented.

hiddeco (approved):
To widen the API, it may be good to introduce in a future release an API which also allows the canonical variables to be configured by one-off configurations (with the canonicals passed in as defaults). However, for the current proposed scope of usage, I think this suffices.

PR #491, opened 2023-02-20
ssa: make `Action` true to type

hiddeco:
This makes the Action type a first-class citizen, over casting it to a string. Resolving a whole lot of annoyances while working with the change set when determining if something equals to (the string representation of).


stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #494, opened 2023-02-21
Change `all` target to run only `tidy|generate|fmt|vet`

hiddeco:
This is much cheaper than running all `test-%` variants, which can now be triggered using `make test`.

It wil also help speeding up the extremely slow CodeQL workflow, which at present runs the whole test suite without it being strictly required. Due to the behavior of the Autobuild step described in https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-the-codeql-workflow-for-compiled-languages#go

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -27,7 +27,7 @@ jobs:
       - name: Run tests
         env:
           SKIP_COSIGN_VERIFICATION: true
-        run: make all
+        run: make test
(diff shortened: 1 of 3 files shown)


stefanprodan:
Please change the build workflow to do `make tidy fmt vet test` as now, the build no longer runs any tests.

stefanprodan (approved):

Issue #476, opened 2023-02-16
Include ref name in the commit when using `RefName` to checkout


aryan9600:
Currently, we don't use the `RefName` in the commit object thats returned by `cloneRefName()`, leading to the following behavior:
```go
commit, err := ggc.Clone(context.TODO(), "https://github.com/fluxcd/pkg", "refs/heads/main", cloneOpts)
fmt.Println(commit.String()) //prints: `sha1:ca1dce0513a89b59c46d2369ad08599d97ea3fd7`
```
we should change this behavior to include the `RefName` in the commit object, so that the string representation of the commit includes the reference:
```go
commit, err := ggc.Clone(context.TODO(), "https://github.com/fluxcd/pkg", "refs/heads/main", cloneOpts)
fmt.Println(commit.String()) //should print: `refs/heads/main@sha1:ca1dce0513a89b59c46d2369ad08599d97ea3fd7`
```

PR #497, opened 2023-02-21
git: add support for showing full references

aryan9600:
Add a new method `AbsoluteReference()` to show the full reference along
with the hash of a commit object. Refactor logic related to
`LastObservedCommit` in order for it to work with full references when
required.

Fixes: #476

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -90,12 +90,23 @@ type Commit struct {
 }
 
 // String returns a string representation of the Commit, composed
-// out the last part of the Reference element, and/or Hash.
-// For example: 'tag-1/a0c14dc8580a23f79bc654faa79c4f62b46c2c22',
-// for a "tag-1" tag.
+// out of the last part of the Reference element (if not empty) and Hash.
+// For example: 'tag-1@sha1:a0c14dc8580a23f79bc654faa79c4f62b46c2c22',
+// for a "refs/tags/tag-1" Reference.
 func (c *Commit) String() string {
 	if short := strings.SplitAfterN(c.Reference, "/", 3); len(short) == 3 {
 		return fmt.Sprintf("%s@%s", short[2], c.Hash.Digest())
+	}
+	return c.Hash.Digest()
+}
+
+// AbsoluteReference returns a string representation of the Commit, composed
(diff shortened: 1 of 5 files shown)


git/git.go
```diff
@@ -87,13 +87,20 @@ type Commit struct {
...
 
-// String returns a string representation of the Commit, composed
-// out the last part of the Reference element, and/or Hash.
+// String returns a string representation of the Commit.
+// By default its composed out the last part of the Reference element
+// and/or Hash. If c.ShowFullRef=true, the entire Reference is used.
 // For example: 'tag-1/a0c14dc8580a23f79bc654faa79c4f62b46c2c22',
 // for a "tag-1" tag.
 func (c *Commit) String() string {
+	if c.ShowFullRef {
```
hiddeco, line 101: Instead of making this an option on the commit, I think it should be a separate method on the commit itself. Otherwise the `Commit` all of sudden bears other data than just commit data.

aryan9600: I considered that as well, but went ahead with this approach as I didn't want consumers of the package having to decide whether to call `.String()` or `.FullString()` (for example)

hiddeco: I think that's absolutely up to the consumer to decide, and dictating it from the perspective of the library is wrong.

git/git.go
```diff
@@ -90,16 +90,27 @@ type Commit struct {
...
 		return fmt.Sprintf("%s@%s", short[2], c.Hash.Digest())
 	}
 	return c.Hash.Digest()
 }
 
+// AbsoulteReference returns a string representation of the Commit, composed
+// out of the Reference element (if not empty) and Hash.
+// For example: 'refs/tags/tag-1@sha1:a0c14dc8580a23f79bc654faa79c4f62b46c2c22'
+// for a "refs/tags/tag-1" Reference.
+func (c *Commit) AbsoulteReference() string {
```
stefanprodan, line 107:
```suggestion
func (c *Commit) AbsoluteReference() string {
```

hiddeco (approved):
LGTM, thank you for following through with the requested changes @aryan9600 :1st_place_medal:

PR #498, opened 2023-02-21
ssa: report Skipped for diff exclusion

hiddeco:
Follow up on #495


stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #500, opened 2023-03-02
runtime/events: change log level to trace

hiddeco:
This commit changes the log level of the event recorder to `trace`, reduing the noise one may receive while running a controller with log level set to `debug` (to get a more extensive insight into what a controller reports, but not as in-depth as `trace`).

This seems a more appropriate level, given the controller often emits a variation of the information pushed in an event by itself on log level `info`, and the knowledge about emitted events likely only has value when deeply looking into the behavior of the application.


stefanprodan (approved):
LGTM

PR #501, opened 2023-03-02
Update API and runtime dependencies

hiddeco:
Excludes `apis/kustomize` as this needs to be updated together with `kustomize` package itself.


stefanprodan:
I would bump the Kubernetes package in every go.mod not only runtime

hiddeco:
I am planning on doing the others in a separate PR, as some things are tangled due to cross-dependencies. It's easier to see if no accidental downgrades happen as was the case with `apis/kustomize`.

stefanprodan (approved):

PR #502, opened 2023-03-02
[pkg/ssa] Do not force apply if resource doesn't exist on the cluster

somtochiama:
Closes https://github.com/fluxcd/flux2/issues/3627


stefanprodan (approved):
LGTM

Thanks @somtochiama please squash.

PR #508, opened 2023-03-07
runtime/logger: add `SetLogger` helper

hiddeco:
This adds a new helper which simultaneously configures controller-runtime and klog to make use of the same logger instance.

### DIFF
--- a/runtime/logger/logger.go
+++ b/runtime/logger/logger.go
@@ -20,6 +20,8 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/spf13/pflag"
 	"go.uber.org/zap/zapcore"
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
@@ -70,7 +72,7 @@ var stackLevelStrings = map[string]zapcore.Level{
 //		flag.Parse()
 //
 //		// Use the values during the initialisation of the logger
-//		ctrl.SetLogger(logger.NewLogger(logOptions))
+//		logger.SetLogger(logger.NewLogger(loggerOptions))
 //	}
 type Options struct {
 	LogEncoding string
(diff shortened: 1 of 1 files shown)


runtime/logger/logger.go
```diff
@@ -112,3 +114,10 @@ func NewLogger(opts Options) logr.Logger {
 
 	return zap.New(zap.UseFlagOptions(&zapOpts))
 }
+
+// SetLogger sets the logger for the controller-runtime and klog packages to the given logger.
+// It is not thread-safe, and should be called as early as possible in the program's execution.
+func SetLogger(logger logr.Logger) {
+	ctrl.SetLogger(logger)
+	klog.SetLoggerWithOptions(logger, klog.ContextualLogger(true))
```
hiddeco, line 122: Should we set a name on this logger?

stefanprodan (approved):

runtime/logger/logger.go
hiddeco: That's not related to anything this is doing? The `setup` log is used in `main`, this sets however the general logging function being utilized by `klog` calls, which at present would result in it logging something like:

```json
{"level":"info","ts":"2023-03-07T12:22:54.290Z","msg":"attempting to acquire leader lease flux-system/helm-controller-leader-election...\n"}
{"level":"info","ts":"2023-03-07T12:22:54.297Z","msg":"successfully acquired lease flux-system/helm-controller-leader-election\n"}
```

Or, for underlying (discovery) client calls:

```json
{"level":"error","ts":"2023-03-07T12:23:21.825Z","msg":"couldn't get resource list for metrics.k8s.io/v1beta1: the server is currently unable to handle the request\n","stacktrace":"k8s.io/client-go/discovery.fetchGroupVersionResources.func1\n\tk8s.io/client-go@v0.26.2/discovery/discovery_client.go:520"}
```

This could be sugared with e.g. `"logger":"runtime"` (or `"logger":"klog", or whatnot) by doing:

```go
klog.SetLoggerWithOptions(logger.WithName("runtime"), klog.ContextualLogger(true))
```

stefanprodan: Ok, runtime feels like a good name

PR #513, opened 2023-03-21
ssa: Set common labels and annotations

stefanprodan:
This PR adds a helper function to set common labels and annotations to Kubernetes objects. This function is needed to avoid dealing with Kustomize commonLabels/Annotations transformers which break deployments due to mutating selectors. 

Ref: https://github.com/fluxcd/kustomize-controller/issues/508


ssa/utils.go
```diff
@@ -40,6 +40,38 @@ import (
 
 const fmtSeparator = "/"
 
+// SetCommonMetadata adds the specified labels and annotations to all objects.
+// Existing keys will have their value overridden.
```
hiddeco, line 44:
```suggestion
// SetCommonMetadata adds the specified labels and annotations to all objects.
// An existing key will have its value overridden.
```

maybe?

hiddeco (approved):

PR #516, opened 2023-03-22
actions: update versions and ease maintenance burden

hiddeco:


stefanprodan (approved):
LGTM

Thanks @hiddeco 🏅

PR #525, opened 2023-03-28
runtime: Add controller sharding capability based on label selector

stefanprodan:
Add `--watch-label-selector` flag to controller runtime. When specified the controller's reconcilers will only watch for changes of those resources with matching labels.

To restrict the controller watch, the selector exposed by this package must be wired in each controller's cache manager.

This enables horizontal scaling of Flux controllers, where each controller can be deployed multiple times with a unique label selector which is used as the sharding key.

For example, to shard the reconciliation of HelmReleases in a cluster, user will deploy two extra helm-controller instances:
- Deployment/helm-controller with `--watch-label-selector=sharding.fluxcd.io/shard notin (shard1, shard2)` 
- Deployment/helm-controller-1 with `--watch-label-selector=sharding.fluxcd.io/shard=shard1` 
- Deployment/helm-controller-2 with `--watch-label-selector=sharding.fluxcd.io/shard=shard2` 

By labeling HelmReleases with `sharding.fluxcd.io/shard`, users can spread the load between `helm-controller-1` and `helm-controller-2` by setting the label value to `shard1` or `shard2`. HelmReleases which don't have the `sharding.fluxcd.io/shard` label, will be reconcile by the `helm-controller` default instance.

Sharding source-controller works the same but users will have to create a matching Kubernetes Service for each SC deployment, then set the Service name in storage address flag. Example:

- Deployment+Service source-controller-1 with: 
```
--storage-adv-addr=source-controller-1.$(RUNTIME_NAMESPACE).svc.cluster.local.
--watch-label-selector=sharding.fluxcd.io/shard=shard1
```

- Deployment+Service source-controller-2 with:
```
--storage-adv-addr=source-controller-2.$(RUNTIME_NAMESPACE).svc.cluster.local.
--watch-label-selector=sharding.fluxcd.io/shard=shard12
```

### DIFF
--- a/runtime/controller/watch.go
+++ b/runtime/controller/watch.go
@@ -1,0 +1,74 @@
+/*
+Copyright 2023 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"github.com/spf13/pflag"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+const (
+	flagWatchWatchAllNamespaces = "watch-all-namespaces"
+	flagWatchLabelSelector      = "watch-label-selector"
+)
+
+// WatchOptions defines the configurable options for reconciler resources watcher.
+type WatchOptions struct {
+	// AllNamespaces defines the watch filter at namespace level.
+	// If set to false, the reconciler will only watch the runtime namespace for resource changes.
+	AllNamespaces bool
+
+	// LabelSelector defines the watch filter based on matching label expressions.
+	// When set, the reconciler will only watch for changes of those resources with matching labels.
+	// Docs: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#list-and-watch-filtering.
+	LabelSelector string
+}
+
+// BindFlags will parse the given pflag.FlagSet for the controller and
+// set the WatchOptions accordingly.
+func (o *WatchOptions) BindFlags(fs *pflag.FlagSet) {
+	fs.BoolVar(&o.AllNamespaces, flagWatchWatchAllNamespaces, true,
+		"Watch for resources in all namespaces, if set to false it will only watch the runtime namespace.")
+	fs.StringVar(&o.LabelSelector, flagWatchLabelSelector, "",
+		"Watch for resources with matching labels e.g. 'sharding.fluxcd.io/shard=shard1'.")
+}
+
+// GetWatchLabelSelector parses the label selector option from WatchOptions
+// and returns an error if the expression is invalid.
+func GetWatchLabelSelector(opts WatchOptions) (*metav1.LabelSelector, error) {
+	if opts.LabelSelector == "" {
+		return nil, nil
+	}
+
+	return metav1.ParseToLabelSelector(opts.LabelSelector)
+}
+
+// GetWatchSelector parses the label selector option from WatchOptions and returns the label selector.
+// If the WatchOptions contain no selectors, then a match everything is returned.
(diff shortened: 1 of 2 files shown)


runtime/controller/watch.go
```diff
@@ -0,0 +1,69 @@
...
+	return metav1.ParseToLabelSelector(opts.LabelSelector)
+}
+
+// Matches returns true if the specified object labels match the given selector.
+func Matches(object client.Object, selector *metav1.LabelSelector) bool {
+	if selector == nil {
+		return true
+	}
+
+	s, _ := metav1.LabelSelectorAsSelector(selector)
```
hiddeco, line 67: This is unsafe as on error `s` will be `nil` and the call below will panic.

hiddeco: Maybe use `klog` here and log the error while returning `false`?

hiddeco: Or simply use `ctrl.Log` which should be setup once this gets called.

runtime/controller/watch.go
```diff
@@ -0,0 +1,76 @@
...
+// BindFlags will parse the given pflag.FlagSet for the controller and
+// set the WatchOptions accordingly.
+func (o *WatchOptions) BindFlags(fs *pflag.FlagSet) {
+	fs.BoolVar(&o.AllNamespaces, flagWatchWatchAllNamespaces, true,
+		"Watch for resources in all namespaces, if set to false it will only watch the runtime namespace.")
+	fs.StringVar(&o.LabelSelector, flagWatchLabelSelector, "",
+		"Watch for resources with matching labels e.g. 'sharding.fluxcd.io/shard=shard1'.")
+}
+
+// GetWatchLabelSelector parses the '--watch-label-selector' flag and returns the label selector.
```
makkes, line 53:
```suggestion
// GetWatchLabelSelector parses the label selector option from opts and returns the label selector.
```

stefanprodan:
I've tested this with source-controller and passing the label selector to the cache manager works flawlessly 🎉

runtime/predicates/match_labels.go
```diff
@@ -0,0 +1,48 @@
+/*
```
hiddeco, line 1: This can now be removed after confirmation, I think.

runtime/controller/watch.go
```diff
@@ -0,0 +1,76 @@
...
+func GetWatchLabelSelector(opts WatchOptions) (*metav1.LabelSelector, error) {
+	if opts.LabelSelector == "" {
+		return nil, nil
+	}
+
+	return metav1.ParseToLabelSelector(opts.LabelSelector)
+}
+
+// Matches returns true if the specified object labels match the given selector.
+func Matches(object client.Object, selector *metav1.LabelSelector) bool {
```
hiddeco, line 63: This can also be removed now.

runtime/controller/watch.go
```diff
@@ -0,0 +1,76 @@
...
+		"Watch for resources with matching labels e.g. 'sharding.fluxcd.io/shard=shard1'.")
+}
+
+// GetWatchLabelSelector parses the label selector option from WatchOptions and returns the label selector.
+func GetWatchLabelSelector(opts WatchOptions) (*metav1.LabelSelector, error) {
+	if opts.LabelSelector == "" {
+		return nil, nil
+	}
+
+	return metav1.ParseToLabelSelector(opts.LabelSelector)
```
hiddeco, line 59: Making this return the result of `metav1.LabelSelectorAsSelector` instead would save us from doing another error catch in every controller.

hiddeco: Or alternative: add another `func GetWatchSelector(opts WatchOptions) (labels.Selector, error)` which wraps this.

runtime/controller/watch_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+import (
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"testing"
+
+	"github.com/spf13/pflag"
```
hiddeco, line 27:
```suggestion
	"testing"

	. "github.com/onsi/gomega"
	corev1 "k8s.io/api/core/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/apimachinery/pkg/labels"
	"sigs.k8s.io/controller-runtime/pkg/client"
	"github.com/spf13/pflag"
```

hiddeco (approved):
Awesome, thanks for following through with the nits :bow:

PR #526, opened 2023-03-29
runtime/leaderelection: add `GenerateID` and update dependencies

hiddeco:


runtime/controller/watch.go
```diff
@@ -72,3 +75,20 @@ func GetWatchSelector(opts WatchOptions) (labels.Selector, error) {
...
 	return metav1.LabelSelectorAsSelector(ls)
 }
+
+// GenerateLeaderElectionID generates a unique leader election ID based on the
+// given name and the watch options. When the watch options contain a label
+// selector, the ID will be suffixed with a short hash of the selector.
+// This is to avoid conflicts when multiple controllers are watching the same
+// resources with different label selectors. The ID will be empty when the
+// given name is empty.
+func GenerateLeaderElectionID(name string, opts WatchOptions) string {
```
hiddeco, line 85: Placement of this is a bit weird as we also have a `runtime/leaderelection`. One solution I can think of is moving this there, while changing the signature to `GenerateLeaderElectionID(name string, salt ...string)`, where all of `salt` is looped into the hash.

stefanprodan (approved):
LGTM

Thanks @hiddeco

runtime/controller/watch.go
hiddeco: Wasn't too happy with it either so updated it now.

PR #528, opened 2023-03-29
Add support for .sourceignore and ignore string

souleb:
If implemented the generator will be able to take into account a `.sourceignore` in the base directory  and a provided `ignore` string while generating/updating the kustomization file before build.

This is needed in order to address https://github.com/fluxcd/flux2/issues/3695


kustomize/go.mod
```diff
@@ -6,7 +6,9 @@ replace github.com/fluxcd/pkg/apis/kustomize => ../apis/kustomize
 
 require (
 	github.com/drone/envsubst v1.0.3
+	github.com/fluxcd/go-git/v5 v5.0.0-20221219190809-2e5c9d01cfc4
 	github.com/fluxcd/pkg/apis/kustomize v1.0.0
+	github.com/fluxcd/pkg/sourceignore v0.3.3
```
stefanprodan, line 11: Please replace this above, like we do for `apis/kustomize`

stefanprodan (approved):
LGTM

Thanks @souleb 🏅

PR #530, opened 2023-04-06
fix edge case where slice has 1 elem

souleb:
there was a bug in how we were slicing our input to remove filtered elements. We now keep track of our position and keep only non-filtered elements. Tested in `flux2` with `.sourceignore` and `ignore-paths`


stefanprodan (approved):
LGTM

Thanks @souleb

PR #539, opened 2023-04-19
git: Reimplement cloneRefName using cloneTag's implementation to support annotated tags

bburky:
`cloneRefName()` failed to support annotated tags and would attempt to clone the annotated tag object's id as if it were a commit object. This would cause the following error (with the tag's hash, not the commit):
```
unable to resolve commit object for 'da39a3ee5e6b4b0d3255bfef95601890afd80709': object not found
```

`cloneTag()` already had support for annotated tags, and could actually be modified to support cloning any ref with no modification. The implementation was moved to `cloneRefName()` and now `cloneTag()` is a wrapper for `cloneRefName()`.

Sharing an implementation between `cloneRefName()` and `cloneTag()` means error messages are slightly worse: they always say "ref" now instead of "tag" when appropriate.

Due to a bug(?) in go-git a ref like `refs/pull/1/head` can't be cloned used with SingleBranch == True. Go-git's `Repository.cloneRefSpec()` will try to build a refspec using `.Short()` on the ref, which doesn't behave the same for refs named like this (that don't [match any RefRevParseRules](https://github.com/src-d/go-git/blob/v4.13.1/plumbing/reference.go#L21)).
https://github.com/src-d/go-git/blob/v4.13.1/repository.go#L826-L829

A test for `cloneRefName()` using an annotated tag was added. Arguably the tests could be merged somewhat now that they share an implementation. There is one change in the cloneTag tests for expectConcreteCommit that I don't quite understand.

There is a TODO in this code about SingleBranch. Other that, I think this approach works. I've tested using the unit tests and they pass. If you have a different approach you'd like to use to fix this, please feel free to close this PR if you want an alternate solution.

### DIFF
--- a/git/gogit/clone.go
+++ b/git/gogit/clone.go
@@ -120,73 +120,9 @@ func (g *Client) cloneBranch(ctx context.Context, url, branch string, opts repos
 }
 
 func (g *Client) cloneTag(ctx context.Context, url, tag string, opts repository.CloneOptions) (*git.Commit, error) {
-	if g.authOpts == nil {
-		return nil, fmt.Errorf("unable to checkout repo with an empty set of auth options")
-	}
-
-	authMethod, err := transportAuth(g.authOpts, g.useDefaultKnownHosts)
-	if err != nil {
-		return nil, fmt.Errorf("unable to construct auth method with options: %w", err)
-	}
-
 	ref := plumbing.NewTagReferenceName(tag)
-	// check if previous revision has changed before attempting to clone
-	if lastObserved := git.TransformRevision(opts.LastObservedCommit); lastObserved != "" {
-		head, err := getRemoteHEAD(ctx, url, ref, g.authOpts, authMethod)
-		if err != nil {
-			return nil, err
-		}
-		hash := git.ExtractHashFromRevision(head)
-		shortRef := fmt.Sprintf("%s@%s", tag, hash.Digest())
-		if head != "" && shortRef == lastObserved {
-			// Construct a non-concrete commit with the existing information.
-			c := &git.Commit{
-				Hash:      hash,
-				Reference: ref.String(),
-			}
-			return c, nil
-		}
-	}
 
-	var depth int
-	if opts.ShallowClone {
-		depth = 1
-	}
-	cloneOpts := &extgogit.CloneOptions{
-		URL:               url,
-		Auth:              authMethod,
-		RemoteName:        git.DefaultRemote,
-		ReferenceName:     plumbing.NewTagReferenceName(tag),
-		SingleBranch:      g.singleBranch,
-		NoCheckout:        false,
-		Depth:             depth,
-		RecurseSubmodules: recurseSubmodules(opts.RecurseSubmodules),
-		Progress:          nil,
-		Tags:              extgogit.NoTags,
-		CABundle:          caBundle(g.authOpts),
-	}
-
-	repo, err := extgogit.CloneContext(ctx, g.storer, g.worktreeFS, cloneOpts)
-	if err != nil {
-		if err == transport.ErrEmptyRemoteRepository || err == transport.ErrRepositoryNotFound || isRemoteBranchNotFoundErr(err, ref.String()) {
-			return nil, git.ErrRepositoryNotFound{
-				Message: fmt.Sprintf("unable to clone: %s", err),
-				URL:     url,
-			}
-		}
-		return nil, fmt.Errorf("unable to clone '%s': %w", url, goGitError(err))
-	}
-
-	head, err := repo.Head()
-	if err != nil {
-		return nil, fmt.Errorf("unable to resolve HEAD of tag '%s': %w", tag, err)
-	}
-	cc, err := repo.CommitObject(head.Hash())
-	if err != nil {
-		return nil, fmt.Errorf("unable to resolve commit object for HEAD '%s': %w", head.Hash(), err)
-	}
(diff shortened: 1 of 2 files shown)


git/gogit/clone_test.go
```diff
@@ -225,7 +225,7 @@ func TestClone_cloneTag(t *testing.T) {
 			tagsInRepo:           []testTag{{"tag-1", false}},
 			checkoutTag:          "tag-1",
 			lastRevTag:           "tag-1",
-			expectConcreteCommit: false,
+			expectConcreteCommit: true,
```
bburky, line 228: I don't really know what caused this to change.

git/gogit/clone.go
```diff
@@ -363,36 +299,77 @@ func (g *Client) cloneSemVer(ctx context.Context, url, semverTag string, opts re
...
+	}
+	cloneOpts := &extgogit.CloneOptions{
+		URL:           url,
+		Auth:          authMethod,
+		RemoteName:    git.DefaultRemote,
+		ReferenceName: ref,
+		// TODO: go-git cloneRefSpec() doesn't handle SingleBranch == True with a ref like refs/pull/1/head because it calls .Short on the ref, and this cannot be converted to a short ref (no matching RefRevParseRules)
+		// I'm not quite sure how this works with SingleBranch == False either, because DefaultFetchRefSpec is "+refs/heads/*:refs/remotes/%s/*" which doesn't refs/pull/1/head either
+		// https://github.com/src-d/go-git/blob/v4.13.1/repository.go#L815-L835
+		// SingleBranch:      g.singleBranch,
```
bburky, line 343: As mentioned in the PR message, go-git CloneOptions.SingleBranch does not support refs/pull/1/head because it builds the refpec to clone wrong.

I'm not entirely sure, but this may be a bug in go-git.

Ignoring singleBranch would be a regression for branches and other non-tag refs. Even with singleBranch=false go-git actually will detect tags automatically (it checks `ReferenceName.IsTag()` and only fetches one ref). I don't have a great fix to work around this. If it is a go-git bug it could be fixed upstream.

aryan9600:
Hi @bburky, thanks for discovering this issue and opening this PR! I did manage to repro this issue on my local.

These changes seem to have a lot of blast radius, bringing forward other issues in go-git and also seems to be breaking existing tests. An easier and straightforward solution might be to instead detect if the ref name is pointing to a tag and then just do `g.cloneTag()`. We could do the same for branches too.
One thing we'd need to make sure about how `Reference` is set on the resultant `git.Commit` object. When using a branch or a tag directly, the `Reference` starts with the short version of the ref, such as `main`, `v0.0.1`, etc. So the logic to quit early by checking `opts.LastObservedCommit` needs to account for this.

asking @hiddeco and @darkowlzz to weigh in here also.

bburky:
> An easier and straightforward solution might be to instead detect if the ref name is pointing to a tag and then just do g.cloneTag()

If you have a good way to do this with a simper change, it may be for the best. This could definitely introduce unexpected issues, I don't know why the expectConcreteCommit test changed for example (but I didn't debug it too closely).

I suppose you could do what you suggest based on a ref name matching `refs/tags/`. That may be a reliable enough heuristic (technically ref names are by convention only and you can put a git tag object under a nonstandard ref). The approach I tried to take was doing `g.cloneTag()` for all refs (which required refactoring done in this PR, because it previously only handled tag names not any ref).

aryan9600:
Now that go-git has support for peeled refs (see: https://github.com/go-git/go-git/pull/750), solving this issue becomes a lot easier. We can use the `AppendPeeled` option when we call `remote.ListContext()` to include the commit ID that all annotated tags point to with the `^{}` syntax. The way I see it we have two options:
1. Instruct users in docs to always use the `^{}` syntax when specifying an annotated tag under a ref name.
2. If we could find the ref && ref is a tag && the ref doesn't end with `^{}`, then we try and see if the ref list contains an entry for that ref appended with `^{}`. If such an entry exists, then we use that instead and if it doesn't we use the original entry.

hiddeco:
My vote would to go to option 2 to reduce the burden on users, as not everyone actually knows what an annotated tag is.

My suggestion would be to use `AppendPeeled` from https://github.com/go-git/go-git/pull/750, and look if there is an exact match or one suffixed with `^{}` to reduce the burden on users.

bburky:
Option 2 sounds good to me from the UX perspective too. We accidentally stumbled across this when a new developer made the tag and they used an annotated tag not expecting there to be any difference at all for Flux. We had to a little bit of debugging to figure out it an annotated tag that caused the issue. I think most users would expect any ref to work with `names:`.

I'm not familiar enough with go-git to implement a solution using peeled refs. I can close this PR if someone else can pick this up. Let me know if you want me to create an issue to track this, but someone else may want to with more info about go-git than I know.

I'd summarize the original issue I reported as:

--- 

cloneRefName() failed to support annotated tags and would attempt to clone the annotated tag object's id as if it were a commit object. This would cause the following error (with the tag's hash, not the commit):

```
unable to resolve commit object for 'da39a3ee5e6b4b0d3255bfef95601890afd80709': object not found
```

Currently Flux's `cloneRefName()` passes the ref to `getRemoteHEAD()` to resolve the ref to an object id, and then passes the id to `cloneCommit()` which only supports commit objects, not annotated tags.

aryan9600:
yes, it'd be great if you could open an issue for this. since you won't be picking this up, i'm closing this PR. thanks again for finding this issue 🙇

PR #540, opened 2023-04-20
git: add support for passphrases for commit gpg signing

aryan9600:
Add a new `CommitOption` for specifying the passphrase to use for decrypting the private key of the signer which will be used for signing the commit.

### DIFF
--- a/git/gogit/client.go
+++ b/git/gogit/client.go
@@ -341,6 +341,12 @@ func (g *Client) Commit(info git.Commit, commitOpts ...repository.CommitOption) 
 	}
 
 	if options.Signer != nil {
+		if options.SignerPassphrase != "" {
+			err = options.Signer.PrivateKey.Decrypt([]byte(options.SignerPassphrase))
+			if err != nil {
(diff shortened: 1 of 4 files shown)


git/gogit/client.go
```diff
@@ -341,6 +341,12 @@ func (g *Client) Commit(info git.Commit, commitOpts ...repository.CommitOption)
 	}
 
 	if options.Signer != nil {
+		if options.SignerPassphrase != "" {
```
hiddeco, line 344: Instead of doing this, and thereby potentially keeping the passphrase around in memory for an extensive period of time. Wouldn't it simply be possible to run the `Decrypt` operation on the private key before providing it as a signer to options?

aryan9600: Yeah but I fail to see how that's an issue? The passphrases are not going to be more than a few bytes, so it's affect on memory consumption should be negligible. Also, a primary reason for this is to make it easier for the consumers of this package. Both flux2 and IAC need this logic, so this can be used in multiple places by us too.

aryan9600: okay closing this PR for now then

PR #548, opened 2023-05-10
http/fetch: improve error messages

hiddeco:
Would have helped with  https://github.com/fluxcd/flux2/issues/3861.


makkes (approved):
💯 Thanks!

http/fetch/archive_fetcher.go
```diff
@@ -150,13 +150,17 @@ func (r *ArchiveFetcher) Fetch(archiveURL, digest, dir string) error {
 // verifyDigest verifies the digest of the reader, and returns an error if it
 // doesn't match.
```
makkes, line 151:
```suggestion
// doesn't match or if the given digest is empty.
```

PR #550, opened 2023-05-10
git: add `PushConfig` to specify push config and add support for refspecs

aryan9600:
Add `PushConfig` for configuring a push operation. Users can use `PushConfig.Refspecs` to specify the refspecs when using `Push()`.

Furthermore, fix a bug related to `Push()` where all refs were pushed to origin, since we did not specify a refspec and the default refspec used by gogit is `refs/heads/*:/refs/heads/*`.

Add `PushConfig.Force` for force pushing and remove `gogit.Client.forcePush` in favor of the former.

Remove some stale test cases from `TestSwitchBranch` related to force push since we don't check if force pushing is enabled while switching branches anymore. Ref: https://github.com/fluxcd/pkg/pull/433

Part of: https://github.com/fluxcd/image-automation-controller/issues/509

### DIFF
--- a/git/gogit/client.go
+++ b/git/gogit/client.go
@@ -52,7 +52,6 @@ type Client struct {
 	authOpts             *git.AuthOptions
 	storer               storage.Storer
 	worktreeFS           billy.Filesystem
-	forcePush            bool
 	credentialsOverHTTP  bool
 	useDefaultKnownHosts bool
 	singleBranch         bool
@@ -144,16 +143,6 @@ func WithMemoryStorage() ClientOption {
 	return func(c *Client) error {
 		c.storer = memory.NewStorage()
 		c.worktreeFS = memfs.New()
-		return nil
-	}
-}
-
-// WithForcePush enables the use of force push for all push operations
-// back to the Git repository.
-// By default this is disabled.
-func WithForcePush() ClientOption {
-	return func(c *Client) error {
-		c.forcePush = true
 		return nil
 	}
 }
@@ -381,7 +370,7 @@ func (g *Client) Push(ctx context.Context, cfg repository.PushConfig) error {
 
 	return g.repository.PushContext(ctx, &extgogit.PushOptions{
 		RefSpecs:   refspecs,
-		Force:      g.forcePush,
+		Force:      cfg.Force,
 		RemoteName: extgogit.DefaultRemoteName,
 		Auth:       authMethod,
 		Progress:   nil,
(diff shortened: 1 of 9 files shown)


hiddeco:
Think this approach will become difficult when we add support for push options, as there are now multiple methods which would require the additional option as an argument.

aryan9600:
Main reason why I added a new function is that I didn't want to break the API. But your point regarding push options is certainly valid. How about we add a new type `PushConfig` which would store refspecs, options (in the future), etc. and change `Push(ctx) error` to `Push(ctx, pushConfig) error`?

hiddeco:
The second option would be to introduce an option pattern, which wouldn't break the existing API from a consumer perspective except for breaking the interface.

stefanprodan:
The git package is not v1 so I would not worry about breaking changes, let's chose the best UX even if it means changing the functions args.

aryan9600:
If we don't have any problems breaking API, I don't see the harm in just passing the config object directly as an argument. It also follows the pattern for `Clone()` where we pass a `CloneOptions` object directly. One thing to note is that I suggested `PushConfig` on purpose as calling it `PushOptions` could (and probably would) lead to confusion given the existence of Git push options.

hiddeco:
If we go that route, I would maybe even go as far as renaming `CloneOptions` to `CloneConfig` to be annoying once instead of annoying forever in terms of naming differences. But otherwise, this sounds good to me.

aryan9600:
cool sounds good to me as well

git/gogit/client.go
```diff
@@ -351,7 +351,7 @@ func (g *Client) Commit(info git.Commit, commitOpts ...repository.CommitOption)
 	return commit.String(), nil
 }
 
-func (g *Client) Push(ctx context.Context) error {
+func (g *Client) Push(ctx context.Context, pushConfig repository.PushConfig) error {
```
hiddeco, line 354:
```suggestion
func (g *Client) Push(ctx context.Context, config repository.PushConfig) error {
```

git/gogit/client.go
```diff
@@ -361,7 +361,26 @@ func (g *Client) Push(ctx context.Context) error {
...
+			return err
+		}
+
+		headRefspec := config.RefSpec(fmt.Sprintf("%s:%[1]s", head.Name()))
+		refspecs = append(refspecs, headRefspec)
+	}
+
 	return g.repository.PushContext(ctx, &extgogit.PushOptions{
+		RefSpecs:   refspecs,
 		Force:      g.forcePush,
```
hiddeco, line 384: Should this now be part of `PushConfig`?

hiddeco (approved):
LGTM, thank you @aryan9600 :bow:

aryan9600:
talked to @pjbgf offline and he's okay with the removal of the obsolete tests

PR #551, opened 2023-05-11
workflows: Update GCP & AZ secret names

darkowlzz:
Update the GCP and Azure secret names to be specific to the test they are for. This will help identify and manage the secrets and the permissions needed for particular tests separately.

Also, update the resource tags to have the name of the repository where they originate from. This will help identify which resources were created from which repository.

NOTE: Had a discussion with @hiddeco  before deciding on the naming format.


stefanprodan (approved):
Great improvements! Thanks @darkowlzz

hiddeco (approved):

PR #564, opened 2023-05-22
Update dependencies across packages

hiddeco:


stefanprodan (approved):
LGTM

aryan9600 (approved):

stefanprodan:
@aryan9600 looks like BitBucket changed in go-git-providers and the tests here now panic. Can you look into this?

aryan9600:
its panicking because we go-git-providers uses the returned response without checking if its nil: https://github.com/fluxcd/go-git-providers/blob/v0.15.3/stash/repositories.go#L226
i don't know why is there an error in the first place, since there have been no changes to the relevant e2e code. @souleb any ideas?

hiddeco:
Created a fix instead in https://github.com/fluxcd/go-git-providers/pull/223

hiddeco:
Based on the CI failure there the certificate of the testing instance has expired. Given this, and the fact that it's in a testing package, going to merge and ignore the failure for now.

Issue #546, opened 2023-05-04
cloneRefName() does not support annotated git tags


bburky:
cloneRefName() does not support annotated git tags and attempts to clone the annotated tag object's id as if it were a commit object. This causes the following error (with the tag's hash, not the commit):

```
unable to resolve commit object for 'da39a3ee5e6b4b0d3255bfef95601890afd80709': object not found
```

Currently Flux's `cloneRefName()` passes the ref to `getRemoteHEAD()` to resolve the ref to an object id, and then passes the id to `cloneCommit()` which only supports commit objects, not annotated tags.

A configuration like the following will trigger the error:

```
apiVersion: source.toolkit.fluxcd.io/v1
kind: GitRepository
metadata:
  name: repository-name
spec:
  ref:
    name: refs/tags/an-annotated-git-tag
```

We accidentally stumbled across this when a new developer made the tag and they used an annotated tag not expecting there to be any difference at all for Flux. We had to a little bit of debugging to figure out that an annotated tag caused the issue. I think most users would expect any ref to work with `name:`.

* * * 

See #539 for discussion and https://github.com/fluxcd/pkg/pull/539#issuecomment-1534250857 for a proposed solution

PR #568, opened 2023-05-24
git/gogit: add support for using annotated tags with `RefName`

aryan9600:
Add support for using annotated tags with `CheckoutStrategy.Refname`. Earlier, users were forced to specify the tag dereference suffix for the clone operation to be successful, as the resolved ref pointed to a tag object and not its child commit object.

Fixes: #546


git/gogit/clone.go
```diff
@@ -427,10 +430,45 @@ func getRemoteHEAD(ctx context.Context, url string, ref plumbing.ReferenceName,
...
+			// Check if the ref is the same as currentRef with the tag dereference suffix.
+			if ref.Name().String() == fmt.Sprintf("%s%s", currentRef.String(), tagDereferenceSuffix) {
+				return fmt.Sprintf("%s@%s", currentRef.String(), git.Hash(ref.Hash().String()).Digest())
+				// If we've reached the end of the list and stil haven't found it, then fallback
+				// to the ref we found earlier.
+			} else if i == len(refs)-1 {
+				return fmt.Sprintf("%s@%s", currentRef.String(), git.Hash(fallbackRef.Hash().String()).Digest())
+			} else {
+				continue
+			}
```
hiddeco, line 471: Think this does the same?

```go
	currentRefStr := currentRef.String()

	for _, ref := range refs {
		if ref.Name().String() == currentRefStr {
			if !currentRef.IsTag() || strings.HasSuffix(currentRefStr, tagDereferenceSuffix) {
				return fmt.Sprintf("%s@%s", currentRefStr, git.Hash(ref.Hash().String()).Digest())
			}
		}

		if ref.Name().String() == currentRefStr+tagDereferenceSuffix {
			return fmt.Sprintf("%s@%s", currentRefStr, git.Hash(ref.Hash().String()).Digest())
		}
	}

	if len(refs) > 0 {
		fallbackRef := refs[0]
		return fmt.Sprintf("%s@%s", currentRefStr, git.Hash(fallbackRef.Hash().String()).Digest())
	}

	return ""
```

hiddeco (approved):
Thank you @aryan9600 :bow:

PR #570, opened 2023-05-24
Add GH Actions for Flux manifests validation

stefanprodan:
These actions are for replacing https://github.com/fluxcd/flux2-kustomize-helm-example/tree/main/.github/actions/tools


actions/kubeconform/action.yml
```diff
@@ -0,0 +1,88 @@
+name: Setup kubeconform CLI
+description: A GitHub Action for installing the kubeconform CLI
+author: Flux project
+branding:
+  color: blue
+  icon: command
+inputs:
+  version:
+    description: Strict SemVer of the Helm CLI to install. Defaults to the latest release.
```
aryan9600, line 9: description needs to be changed for kubeconform accordingly.

hiddeco (approved):
LGTM, thank you @stefanprodan :1st_place_medal:

PR #571, opened 2023-05-25
git/gogit: set `HostKeyCallback` for parent `PublicKeys` object

aryan9600:
Set `HostKeyCallback` for the parent `PublicKeys` object to avoid setting the callback to one that uses the system's known_hosts.

Fixes errors seen in: https://github.com/fluxcd/source-controller/actions/runs/5072863613/jobs/9111130473#step:4:164


hiddeco (approved):

hiddeco (commented):
While not blocking to get this quickly patched, I think we should add a test to ensure regression does not occur.

hiddeco (approved):
Thanks for adding a test :1st_place_medal:

PR #591, opened 2023-06-22
oci: Allow consumers to set the OCI artifact created date

stefanprodan:
For reproducible builds we need to allow consumers to set their own created date instead of enforcing time.now.


makkes (commented):
Can we have a test for this in push_pull_test.go? 🙏🏻

hiddeco (approved):

makkes (approved):

PR #594, opened 2023-06-26
runtime/client: add util func to check conn options compatibility with env

aryan9600:
Add `CheckEnvironmentCompatibility()` to check whether the configured connection options are compatible with the environment, by checking env vars like `HTTP_PROXY` and `HTTPS_PROXY`.

### DIFF
--- a/runtime/controller/connection.go
+++ b/runtime/controller/connection.go
@@ -16,12 +16,21 @@ limitations under the License.
 
 package controller
 
-import "github.com/spf13/pflag"
+import (
+	"fmt"
+	"net/url"
 
-const flagInsecureAllowHTTP = "insecure-allow-http"
+	"github.com/spf13/pflag"
+	"golang.org/x/net/http/httpproxy"
+)
+
+const (
+	flagInsecureAllowHTTP = "insecure-allow-http"
+)
 
 // ConnectionOptions defines the configurable options for outbound connections
-// opened by reconcilers.
+// opened by reconcilers. Consumers are expected to check for compatibility via
+// `CheckEnvironmentCompatibility()` before using its values.
 type ConnectionOptions struct {
 	// AllowHTTP, if set to true allows the controller to make plain HTTP
 	// connections to external services.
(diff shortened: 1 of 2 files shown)


runtime/controller/connection.go
```diff
@@ -16,9 +16,19 @@ limitations under the License.
...
+	"os"
 
-const flagInsecureAllowHTTP = "insecure-allow-http"
+	"github.com/spf13/pflag"
+)
+
+const (
+	flagInsecureAllowHTTP = "insecure-allow-http"
+	HttpProxyEnvVar       = "HTTP_PROXY"
+	HttpsProxyEnvVar      = "HTTPS_PROXY"
```
hiddeco, line 30:
```suggestion
	EnvVarHttpProxy       = "HTTP_PROXY"
	EnvVarHttpsProxy      = "HTTPS_PROXY"
```

runtime/controller/connection.go
```diff
@@ -34,3 +44,24 @@ func (o *ConnectionOptions) BindFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&o.AllowHTTP, flagInsecureAllowHTTP, true,
 		"Allow the controller to make HTTP requests to external services like insecure Git servers, container registries, etc.")
 }
+
+// CheckEnvironmentCompatibility checks if the enviornment is compatible with
+// the configured connection options.
+func (o *ConnectionOptions) CheckEnvironmentCompatibility() error {
```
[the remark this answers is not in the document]
aryan9600: i prefer the current name over `Validate()` as imo, the latter means that its checking the configured options are valid which is a bit misleading.

runtime/controller/connection.go
```diff
@@ -34,3 +44,24 @@ func (o *ConnectionOptions) BindFlags(fs *pflag.FlagSet) {
...
+	if !o.AllowHTTP {
```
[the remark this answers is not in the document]
aryan9600: i prefer to keep it the current way since it keeps all the logic related to the `o.AllowHTTP` field in a single block making it relatively easy for this function to be expanded for when we add more options.

runtime/controller/connection.go
```diff
@@ -16,9 +16,19 @@ limitations under the License.
...
+	"os"
 
-const flagInsecureAllowHTTP = "insecure-allow-http"
+	"github.com/spf13/pflag"
+)
+
+const (
+	flagInsecureAllowHTTP = "insecure-allow-http"
+	EnvVarHTTPProxy       = "HTTP_PROXY"
+	EnvVarHTTPSProxy      = "HTTPS_PROXY"
```
makkes, line 30: I think it's safer to use [httpproxy.FromEnvironment](https://pkg.go.dev/golang.org/x/net/http/httpproxy#FromEnvironment) here as that's the source of truth for Go's runtime.

stefanprodan (approved):
LGTM

runtime/controller/connection.go
```diff
@@ -34,3 +43,26 @@ func (o *ConnectionOptions) BindFlags(fs *pflag.FlagSet) {
...
 		"Allow the controller to make HTTP requests to external services like insecure Git servers, container registries, etc.")
 }
+
+// CheckEnvironmentCompatibility checks if the enviornment is compatible with
+// the configured connection options.
+func (o *ConnectionOptions) CheckEnvironmentCompatibility() error {
+	if !o.AllowHTTP {
+		config := httpproxy.FromEnvironment()
+		if config.HTTPProxy != "" {
+			return fmt.Errorf("usage of HTTP requests is blocked but found a HTTP proxy set in enviornment.")
```
hiddeco, line 53:
```suggestion
			return fmt.Errorf("usage of HTTP requests is blocked but found a HTTP proxy set in enviornment")
```

runtime/controller/connection.go
```diff
@@ -34,3 +43,26 @@ func (o *ConnectionOptions) BindFlags(fs *pflag.FlagSet) {
...
+		}
+
+		if config.HTTPSProxy != "" {
+			proxy, err := url.Parse(config.HTTPSProxy)
+			if err != nil {
+				return fmt.Errorf("unable to parse address specified in the HTTPS proxy enviornment setting: %w", err)
+			}
+
+			if proxy.Scheme != "https" {
+				return fmt.Errorf("usage of HTTP requests is blocked but found a non-https address in the HTTPS proxy enviornment setting.")
```
hiddeco, line 63:
```suggestion
				return fmt.Errorf("usage of HTTP requests is blocked but found a non-https address in the HTTPS proxy enviornment setting")
```

hiddeco (approved):

PR #595, opened 2023-06-26
Align `go.mod` version with Kubernetes (Go 1.20)

stefanprodan:
Part of: https://github.com/fluxcd/flux2/issues/4014


makkes (commented):
What about [.github/actions/run-tests/Dockerfile](https://github.com/fluxcd/pkg/blob/main/.github/actions/run-tests/Dockerfile#L2)?

makkes (approved):

PR #597, opened 2023-06-30
git: Add support for adding tag info to commit

aryan9600:
Add a new struct `Tag` that represents a Git annotated tag and add it as a new field `Commit.ReferencingTag`, allowing for a commit to contain information about its referencing tag. Add support for verifying the referencing tag as well.
Update the cloning logic to include the tag object in the returned commit object when checking out via a specific tag, a semver range or if the provided refname points to a tag.

Related to: https://github.com/fluxcd/source-controller/issues/1133
Example usage: https://github.com/fluxcd/source-controller/pull/1187

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -87,6 +87,8 @@ type Commit struct {
 	Encoded []byte
 	// Message is the commit message, containing arbitrary text.
 	Message string
+	// ReferencingTag is the parent tag, that points to this commit.
+	ReferencingTag *AnnotatedTag
 }
 
 // String returns a string representation of the Commit, composed
@@ -113,24 +115,15 @@ func (c *Commit) AbsoluteReference() string {
 
 // Verify the Signature of the commit with the given key rings.
 // It returns the fingerprint of the key the signature was verified
-// with, or an error.
-func (c *Commit) Verify(keyRing ...string) (string, error) {
-	if c.Signature == "" {
-		return "", fmt.Errorf("commit does not have a PGP signature")
+// with, or an error. It does not verify the signature of the parent
+// tag (if present). Users are expected to explicitly verify the parent
+// tag's signature using `c.ReferencingTag.Verify()`
+func (c *Commit) Verify(keyRings ...string) (string, error) {
+	fingerprint, err := verifySignature(c.Signature, c.Encoded, keyRings...)
+	if err != nil {
+		return "", fmt.Errorf("unable to verify Git commit: %w", err)
 	}
-
-	for _, r := range keyRing {
-		reader := strings.NewReader(r)
-		keyring, err := openpgp.ReadArmoredKeyRing(reader)
-		if err != nil {
-			return "", fmt.Errorf("unable to read armored key ring: %w", err)
-		}
-		signer, err := openpgp.CheckArmoredDetachedSignature(keyring, bytes.NewBuffer(c.Encoded), bytes.NewBufferString(c.Signature), nil)
-		if err == nil {
-			return signer.PrimaryKey.KeyIdString(), nil
-		}
-	}
-	return "", fmt.Errorf("unable to verify commit with any of the given key rings")
+	return fingerprint, nil
 }
 
 // ShortMessage returns the first 50 characters of a commit subject.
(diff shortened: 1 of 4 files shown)


stefanprodan:
Is this part of https://github.com/fluxcd/source-controller/issues/1133? if so please add it to the PR description.

git/git.go
```diff
@@ -87,6 +84,8 @@ type Commit struct {
 	Encoded []byte
 	// Message is the commit message, containing arbitrary text.
 	Message string
+	// Parent is the parent tag, that points to this commit.
+	Parent *AnnotatedTag
```
hiddeco, line 88:
```suggestion
	ReferencingTag *AnnotatedTag
```

maybe?

git/utils.go
```diff
@@ -133,3 +136,22 @@ func ExtractHashFromRevision(rev string) Hash {
 	}
 	return Hash(rev)
 }
+
+func verifySignature(sig string, payload []byte, keyRings ...string) (string, error) {
```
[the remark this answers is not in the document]
aryan9600: I don't think I agree. in its current form, its a generic function that will verify any payload which is signed with PGP and has nothing to do with Git.
also, creating a new file just for a single function seems unnecessary, when we already have a file for such kind of helper functions.

git/gogit/clone.go
```diff
@@ -189,8 +190,39 @@ func (g *Client) cloneTag(ctx context.Context, url, tag string, opts repository.
...
+		}
+	}
+	// The commit is considered to have a parent only if the tag is annotated,
+	// as lightweight tags are essentially named pointers.
+	if annotated {
+		annotatedTag, err := buildTag(tagObj)
+		if err != nil {
+			return nil, err
+		}
+		commit.Parent = annotatedTag
```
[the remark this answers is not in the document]
aryan9600: we'd need to pass the repo object to this wrapper since we need to resolve the tag and fetch the tag object.
we also resolve the commit in the clone function and then pass the commit object to `buildCommitWithRef()` to build our repr of the commit, so whats the issue with resolving the tag in the function responsible for cloning tags and then building our repr of a tag using `buildTag()`?

git/utils.go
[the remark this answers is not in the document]
hiddeco: It's private and not used within the same file, which is indeed weird and uneasy when reasoning about the logic in `git.go`.

hiddeco (approved):
LGTM! Thank you @aryan9600 :bow:

aryan9600:
wondering whether we should also add support for verifying tags if `.spec.ref.name` points to a tag?

hiddeco:
If we can automagically resolve this, it would probably be good for feature completeness.

git/gogit/clone.go
```diff
@@ -254,11 +261,29 @@ func (g *Client) cloneCommit(ctx context.Context, url, commit string, opts repos
 	if err != nil {
 		return nil, fmt.Errorf("unable to checkout commit '%s': %w", commit, err)
 	}
-	g.repository = repo
+
+	var tagObj *object.Tag
 	if opts.RefName != "" {
 		cloneOpts.ReferenceName = plumbing.ReferenceName(opts.RefName)
+		// If the refname points to a tag then try to resolve the tag object and include
+		// in in the commit being returned. Refname that point to a tag but have the dereference
```
hiddeco, line 269:
```suggestion
		// in the commit being returned. Refnames that point to a tag but have the dereference
```

hiddeco (approved):

git/git.go
```diff
@@ -164,11 +164,25 @@ func (t *AnnotatedTag) Verify(keyRings ...string) (string, error) {
...
+// the tag name, i.e. "1.0.0".
+func (t *Tag) String() string {
+	if len(t.Hash) == 0 {
+		return t.Name
+	}
 	return fmt.Sprintf("%s@%s", t.Name, t.Hash.String())
 }
 
+// IsAnnotated returns true of if the tag is annotated.
+func (t *Tag) IsAnnotated() bool {
```
hiddeco, line 178: Any reason why we prefer methods here over functions (versus `IsConcreteCommit`)?

aryan9600: i prefer methods, as they make more semantic sense. since you're computing a property of that particular object, i feel a method is more apt than a function.

hiddeco: I feel the same, but this now creates a mixed usage pattern within the same library. Which doesn't make the UX intuitive.

stefanprodan (approved):
LGTM

Thanks @aryan9600 🏅

aryan9600:
[Failing BitBucket Server test](https://github.com/fluxcd/pkg/actions/runs/5928696230/job/16075307811?pr=597) is due to TLS issues on the upstream server and not related to these changes.

PR #604, opened 2023-07-10
oci: Improve handling of cosgin artifacts

errordeveloper:
Cosign uses "special" tags that have the following naming format:

```    
<digestAgorithm>-<digestHex>.<artifactType(.att, .sbom, .sig)>
```
    
These are used to attach attestations, SBOMs and signatures to a  subject image reference by the first portion of the tag.
    
Until now Flux would ignore `.sig` tags, as its primary goal is to filter through app image versions. This change extends the list of omitted cosing artifacts to cover `.att` and `.sbom`.
    
In some cases it's desirable to observe cosgin artifacts, e.g. in order to determine wheather an image had been signed or not. Hence this change introduces a new field in `ListOptions`.

### DIFF
--- a/oci/client/list.go
+++ b/oci/client/list.go
@@ -37,6 +37,9 @@ type ListOptions struct {
 	SemverFilter string
 	// RegexFilter contains a regex that tags will be filtered by.
 	RegexFilter string
+	// IncludeCosignArtifacts can be used to include cosign attestation,
+	// signature and SBOM tags in the list, as these are excluded by default.
+	IncludeCosignArtifacts bool
 }
 
 // List fetches the tags and their manifests for a given OCI repository.
@@ -66,8 +69,8 @@ func (c *Client) List(ctx context.Context, url string, opts ListOptions) ([]Meta
 	}
 
 	for _, tag := range tags {
-		// exclude cosign signatures
-		if strings.HasSuffix(tag, ".sig") {
+		// ignore cosign artifacts by default
+		if !opts.IncludeCosignArtifacts && IsCosignArtifact(tag) {
 			continue
 		}
 
@@ -117,4 +120,16 @@ func (c *Client) List(ctx context.Context, url string, opts ListOptions) ([]Meta
 	}
 
 	return metas, nil
+}
+
+// IsCosignArtifact will return true if the tag has one of the following suffices:
+// ".att", ".sbom", or ".sig". These are the suffices used by cosign to store the
+// attestations, SBOMs, and signatures respectively.
+func IsCosignArtifact(tag string) bool {
+	for _, suffix := range []string{".att", ".sbom", ".sig"} {
+		if strings.HasSuffix(tag, suffix) {
+			return true
+		}
+	}
+	return false
 }
--- a/oci/client/list_test.go
+++ b/oci/client/list_test.go
@@ -35,7 +35,18 @@ func Test_List(t *testing.T) {
 	ctx := context.Background()
 	c := NewClient(DefaultOptions())
 	repo := "test-list" + randStringRunes(5)
-	tags := []string{"v0.0.1", "v0.0.2", "v0.0.3", "v6.0.0", "v6.0.1", "v6.0.2", "v6.0.2-rc.1", "v6.0.2-alpha", "staging-fb3355b"}
+	appTags := []string{
+		"v0.0.1", "v0.0.2", "v0.0.3", "v6.0.0", "v6.0.1", "v6.0.2", "v6.0.2-rc.1", "v6.0.2-alpha",
+		"staging-fb3355b",
+	}
+	cosignTags := []string{
+		"sha256-e2688bb75ee43df49c9bfb2aa30dd98173649db53955e87c347024ba71bc1c80.sig",
+		"sha256-e2688bb75ee43df49c9bfb2aa30dd98173649db53955e87c347024ba71bc1c80.sbom",
+		"sha256-e2688bb75ee43df49c9bfb2aa30dd98173649db53955e87c347024ba71bc1c80.att",
+	}
+	allTags := append([]string{}, cosignTags...)
+	allTags = append(allTags, appTags...)
+
 	source := "github.com/fluxcd/fluxv2"
 	rev := "rev"
 	ct := time.Now()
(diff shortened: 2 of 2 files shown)


oci/client/list.go
```diff
@@ -37,6 +37,9 @@ type ListOptions struct {
 	SemverFilter string
 	// RegexFilter contains a regex that tags will be filtered by.
 	RegexFilter string
+	// IncludeCosignSignatures can be used to include cosign signature tags
+	// (i.e. ".sig" suffix) in the list, as these are excluded by default.
+	IncludeCosignSignatures bool
```
stefanprodan, line 42: I would rename this to `IncludeCosignArtifacts` and in the exclusion list please add `.att`.

As seen here, there are 2 artifacts for each container image, `sig` and `att`: https://github.com/fluxcd/kustomize-controller/pkgs/container/kustomize-controller

errordeveloper: Sure! They also use '.sbom', should I add it also?

stefanprodan: Yes please

oci/client/list.go
```diff
@@ -118,3 +121,12 @@ func (c *Client) List(ctx context.Context, url string, opts ListOptions) ([]Meta
 
 	return metas, nil
 }
+
+func isCosigneArtifact(tag string) bool {
```
errordeveloper, line 125:
```
func (o ListOptions) isCosigneArtifact(tag string) bool
```
I'm temped to make it more like this, but maybe it'd be pointless - wdyt?

oci/client/list.go
```diff
@@ -66,8 +69,8 @@ func (c *Client) List(ctx context.Context, url string, opts ListOptions) ([]Meta
 	}
 
 	for _, tag := range tags {
-		// exclude cosign signatures
-		if strings.HasSuffix(tag, ".sig") {
+		// ignore cosign artifacts by default
+		if isCosignArtifact(tag) && !opts.IncludeCosignArtifacts {
```
errordeveloper, line 73: I'm temped to make it a method on `ListOptions`, but maybe it'd be pointless - wdyt?
```
func (o ListOptions) ignore(tag string) bool {
      for _, suffix := range []string{".att", ".sbom", ".sig"} {
 		if strings.HasSuffix(tag, suffix) {
 			return !o.IncludeCosignArtifacts
 	        }
      }
      return false
 }
```
So it's be called like `if opts.ignore(tag)`...

oci/client/list.go
```diff
@@ -118,3 +121,12 @@ func (c *Client) List(ctx context.Context, url string, opts ListOptions) ([]Meta
 
 	return metas, nil
 }
+
+func isCosignArtifact(tag string) bool {
```
stefanprodan, line 125: Let’s make this public as we need it in image-reflector-controller, also please add a comment to it.

oci/client/list.go
stefanprodan: We'll need the `IsCosignArtifact` func in other places of Flux where `List` is not used. So I would keep it as stand alone public function.

stefanprodan (approved):
LGTM

Thanks @errordeveloper 🏅

oci/client/list.go
```diff
@@ -66,8 +69,8 @@ func (c *Client) List(ctx context.Context, url string, opts ListOptions) ([]Meta
 	}
 
 	for _, tag := range tags {
-		// exclude cosign signatures
-		if strings.HasSuffix(tag, ".sig") {
+		// ignore cosign artifacts by default
+		if IsCosignArtifact(tag) && !opts.IncludeCosignArtifacts {
```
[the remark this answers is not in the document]
errordeveloper: Thanks, I amended the commit with this!

errordeveloper:
I updated the tests, but couldn't run these locally as they seem to push images to the registry, will wait for CI.

stefanprodan:
> I updated the tests, but couldn't run these locally as they seem to push images to the registry

You should be able to do that, we spin un an in-memory registry that gets wiped when tests are done.

oci/client/list_test.go
```diff
@@ -80,14 +97,39 @@ func Test_List(t *testing.T) {
 			regexFilter:  "alpha",
 			expectedTags: []string{"v6.0.2-alpha"},
 		},
+		{
+			name:         "filter for  6.0.x",
+			semverFilter: ">=6.0.0-rc",
+			regexFilter:  "alpha",
+			expectedTags: []string{"v6.0.2-alpha"},
+		},
```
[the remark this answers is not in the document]
errordeveloper: Thanks, just pasted it with intention to change, but forgot :)

oci/client/list_test.go
```diff
@@ -35,7 +35,18 @@ func Test_List(t *testing.T) {
...
+		"v0.0.1", "v0.0.2", "v0.0.3", "v6.0.0", "v6.0.1", "v6.0.2", "v6.0.2-rc.1", "v6.0.2-alpha",
+		"staging-fb3355b",
+	}
+	cosignTags := []string{
+		"sha256-e2688bb75ee43df49c9bfb2aa30dd98173649db53955e87c347024ba71bc1c80.sig",
+		"sha256-e2688bb75ee43df49c9bfb2aa30dd98173649db53955e87c347024ba71bc1c80.sbom",
+		"sha256-e2688bb75ee43df49c9bfb2aa30dd98173649db53955e87c347024ba71bc1c80.att",
+	}
+	allTags := append([]string{}, cosignTags...)
+	allTags = append(allTags, appTags...)
```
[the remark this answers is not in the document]
errordeveloper: Well, yes, but it actually adds to the underlying slice of `allTags`, which has issues, so I prefer to be explicitly safe.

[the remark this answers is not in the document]
errordeveloper: I assume you meant `allTags := append(appTags, cosignTags...)`... 
Yes, it largely does the same thing, but it will use the underlying location of `appTags`, and it's possible to end up with a combined slice in the wrong place.  I prefer to allocate a new slice explicitly.

stefanprodan:
@errordeveloper I've released the package with your changes in `oci/v0.30.0`.

PR #609, opened 2023-07-17
Add failfast option to ssa.Wait

somtochiama:
This pull request adds a FailFast option to the `WaitOptions` in the ssa pkg. When this option is set, the Wait function returns early if there are stalled resources.

Ref: https://github.com/fluxcd/flux2/discussions/3980
Ref: https://github.com/fluxcd/kustomize-controller/issues/701

### DIFF
--- a/ssa/manager_wait.go
+++ b/ssa/manager_wait.go
@@ -43,6 +43,9 @@ type WaitOptions struct {
 	// Timeout defines after which interval should the engine give up on waiting for resources
 	// to become ready.
 	Timeout time.Duration
+
+	// FailFast makes the Wait function return an error as soon as a resource reaches the failed state.
+	FailFast bool
 }
 
 // DefaultWaitOptions returns the default wait options where the poll interval is set to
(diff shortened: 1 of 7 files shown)


ssa/manager_wait.go
```diff
@@ -129,7 +145,7 @@ func (m *ResourceManager) WaitForSet(set object.ObjMetadataSet, opts WaitOptions
 				errors = append(errors, builder.String())
 			}
 		}
-		return fmt.Errorf("timeout waiting for: [%s]", strings.Join(errors, ", "))
+		return fmt.Errorf("%s: [%s]", msg, strings.Join(errors, ", "))
```
stefanprodan, line 148: Does the error contains the objects that have failed? what does the message look like, can you add a test for it please.

ssa/manager_wait.go
```diff
@@ -43,6 +43,10 @@ type WaitOptions struct {
 	// Timeout defines after which interval should the engine give up on waiting for resources
 	// to become ready.
 	Timeout time.Duration
+
+	// FailFast option allows the Wait function to return early if one or more resources are stalled
+	// and the rest are ready.
```
stefanprodan, line 48:
```suggestion
	// FailFast makes the Wait function return an error as soon as a resource reaches the failed state.
```

ssa/manager_wait.go
```diff
@@ -91,12 +96,17 @@ func (m *ResourceManager) WaitForSet(set object.ObjMetadataSet, opts WaitOptions
...
+					countFailed++
+				}
 				rss = append(rss, rs)
 			}
+			failedResources = countFailed
 
 			desired := status.CurrentStatus
 			aggStatus := aggregator.AggregateStatus(rss, desired)
-			if aggStatus == desired {
+			if aggStatus == desired || opts.FailFast && countFailed > 0 {
```
stefanprodan, line 109:
```suggestion
			if aggStatus == desired || (opts.FailFast && countFailed > 0) {
```

stefanprodan (approved):
LGTM

Thanks @somtochiama 🏅

PR #613, opened 2023-08-01
runtime: add `jitter` utils and options

hiddeco:
This adds a `jitter` package to runtime to allow the configuration of a random +/-% jitter to the interval at which resources are requeued.

Applying jitter to the interval duration can be useful to mitigate spikes in memory and CPU usage caused by many resources being configured with the same interval.

### DIFF
--- a/runtime/jitter/interval.go
+++ b/runtime/jitter/interval.go
@@ -104,18 +104,15 @@ type IntervalOptions struct {
 // BindFlags will parse the given pflag.FlagSet and load the interval jitter
 // with the default value of 10%.
 func (o *IntervalOptions) BindFlags(fs *pflag.FlagSet) {
-	o.BindFlagsWithDefault(fs, -1)
+	o.BindFlagsWithDefault(fs, defaultIntervalJitterPercentage)
 }
 
 // BindFlagsWithDefault will parse the given pflag.FlagSet and load the interval
 // jitter. The defaultPercentage is used to set the default value for the
 // interval jitter percentage. If the defaultPercentage is negative, then the
 // default value (of 10%) will be used.
-func (o *IntervalOptions) BindFlagsWithDefault(fs *pflag.FlagSet, defaultPercentage int) {
-	if defaultPercentage < 0 {
-		defaultPercentage = defaultIntervalJitterPercentage
-	}
-	fs.Uint8Var(&o.Percentage, flagIntervalJitter, uint8(defaultPercentage),
+func (o *IntervalOptions) BindFlagsWithDefault(fs *pflag.FlagSet, defaultPercentage uint8) {
+	fs.Uint8Var(&o.Percentage, flagIntervalJitter, defaultPercentage,
 		"Percentage of jitter to apply to interval durations. A value of 10 "+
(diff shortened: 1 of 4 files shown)


stefanprodan (approved):
LGTM

Thanks @hiddeco 🏅

runtime/jitter/jitter.go
```diff
@@ -0,0 +1,61 @@
...
+// When p <= 0 or p >= 1, duration is returned without a modification.
+// If r is nil, a new rand.Rand will be created using the current time as the
+// seed.
+func Percent(p float64, r *rand.Rand) Duration {
+	r = defaultOrRand(r)
+	if p <= 0 || p >= 1 {
+		return NoJitter
+	}
+	return func(d time.Duration) time.Duration {
+		randomP := p * (2*r.Float64() - 1)
```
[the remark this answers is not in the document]
hiddeco: No common name, just a formula with an equal chance of being +/-.

Good resource for (formulas around) jitter in general (albeit partly based on error rates) is https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/.

runtime/jitter/jitter.go
```diff
@@ -0,0 +1,61 @@
...
+	"math/rand"
+	"time"
+)
+
+// Duration is a function that takes a duration and returns a modified duration
+// with jitter added.
+type Duration func(time.Duration) time.Duration
+
+// NoJitter returns a Duration function that will return the given duration
+// without modification.
```
aryan9600, line 29: this seems to suggest that `NoJitter` returns a func, but it just returns the provided `time.Duration` unmodified

aryan9600: could be better to do this, to bring it in alignment with `Percent`, but what's here currently is also fine, as long as the comments reflect it properly.
```go
func NoJitter(d time.Duration) Duration {
	return func(d time.Duration) time.Duration {
	   return d
	}
}
```

runtime/jitter/interval.go
```diff
@@ -0,0 +1,134 @@
...
+	// Percentage of jitter to apply to interval durations. A value of 10
+	// will apply a jitter of +/-10% to the interval duration. It can not be negative,
+	// and must be less than 100.
+	Percentage uint8
+}
+
+// BindFlags will parse the given pflag.FlagSet and load the interval jitter
+// with the default value of 10%.
+func (o *IntervalOptions) BindFlags(fs *pflag.FlagSet) {
+	o.BindFlagsWithDefault(fs, -1)
```
aryan9600, line 107: just curious, why not just pass 10 directly? i know it'll be set to 10 because its negative, but this way someone else reading the function would be instantly be sure that the default is 10%

runtime/jitter/jitter.go
hiddeco: Should have been "is a ...", thanks :bow:

runtime/jitter/interval.go
hiddeco: Last minute modifications and not spending sufficient time to think them through, really :-)

aryan9600 (approved):
lgtm! 💯

PR #625, opened 2023-08-08
ssa: Add `IfNotPresent` apply strategy

stefanprodan:
Add `IfNotPresentSelector` to the apply options, the resources that match this selector will be applied on the cluster only if they don't exist.

Ref: https://github.com/fluxcd/kustomize-controller/issues/936

### DIFF
--- a/ssa/utils.go
+++ b/ssa/utils.go
@@ -307,8 +307,11 @@ func IsImmutableError(err error) bool {
 // AnyInMetadata searches for the specified key-value pairs in labels and annotations,
 // returns true if at least one key-value pair matches.
 func AnyInMetadata(object *unstructured.Unstructured, metadata map[string]string) bool {
+	labels := object.GetLabels()
+	annotations := object.GetAnnotations()
 	for key, val := range metadata {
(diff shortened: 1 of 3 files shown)


ssa/utils.go
```diff
@@ -308,7 +308,8 @@ func IsImmutableError(err error) bool {
 // returns true if at least one key-value pair matches.
 func AnyInMetadata(object *unstructured.Unstructured, metadata map[string]string) bool {
 	for key, val := range metadata {
-		if object.GetLabels()[key] == val || object.GetAnnotations()[key] == val {
+		if strings.ToLower(object.GetLabels()[key]) == strings.ToLower(val) ||
```
hiddeco, line 311:
```suggestion
	labels := object.GetLabels()
	annotations := object.GetAnnotations()

	for key, val := range metadata {
		if (labels[key] != "" && strings.EqualFold(labels[key], val)) ||
		   (annotations[key] != "" && strings.EqualFold(annotations[key], val)) {
			return true
		}
	}

	return false
```

hiddeco (approved):

PR #627, opened 2023-08-14
feat(client): kube-api timeout requests

genofire:


genofire:
@stefanprodan any feedback?

genofire:
where is this timeout set?

stefanprodan:
The context comes from controller runtime, e.g. https://github.com/fluxcd/kustomize-controller/blob/385df6785d602eed39ffd7310be021c1ef2d518b/internal/controller/kustomization_controller.go#L165

genofire:
sorry, i mean - where is the context created? (for my problem inside of source-controller for helmrepos).

stefanprodan:
@hiddeco @darkowlzz any idea how SC could get stuck forever when reconciling Helm repos?

patsevanton:
Are there any updates?

PR #634, opened 2023-08-21
Remove special error handling for secrets

somtochiama:
This pull request removes the special error handling for secrets that was used to prevent leaking secret data in the error message. With the move to server-side apply, the original patch (which contained the secret data) is no longer returned in the error.

Fix: https://github.com/fluxcd/kustomize-controller/issues/946
Fix: https://github.com/fluxcd/flux2/issues/4122


stefanprodan (approved):
LGTM

Thanks @somtochiama please rebase.

PR #637, opened 2023-08-22
runtime/tls: enforce minimum version

hiddeco:


stefanprodan (approved):

hiddeco:
Not required, is actually the default since a couple Go versions back.

Issue #645, opened 2023-09-06
bug: (pkg/oci/auth) Importing the auth package causes "`kubeconfig` flag redefined" error


vishal-chdhry:
## Background

We at Kyverno were facing issues with the ACR keychain where there was a 30-second timeout when using an image from Azure ACR in a non-AKS cluster. To work around that problem we decided to use the auth pkg in this repo to create custom keychains. The implementation works fine, but there is a problem.

PR: https://github.com/kyverno/kyverno/pull/7908
## Description

After importing `github.com/fluxcd/pkg/oci v0.30.0` in kyverno, we started getting an error when running the pod saying [`panic: /ko-app/kyverno flag redefined: kubeconfig` ](https://github.com/kyverno/kyverno/pull/7908#issuecomment-1697423977). `fluxcd/pkg/oci` is the only new direct import in the PR. So I believe that is whats causing the issue here.

Any suggestion on how to fix it?

kingdonb:
We just had an interesting discussion about this at Bug Scrub, It looks like the controller-runtime [are aware](https://github.com/kubernetes-sigs/controller-runtime/blob/2a553d6f910dfc4b0796142ba1e059639a21c54a/pkg/client/config/config.go#L39-L45) of this issue and I might have found this by clicking through @makkes links.

The working theory is that by importing `fluxcd/pkg` to get the oci package, we've changed the dependency order and moved controller-runtime up the import stack, and so this flag gets defined by controller-runtime in the updated version. Whereas it was getting defined by kyverno before and controller-runtime was observing the flag already defined, and avoiding redefining it. That explains it wasn't a problem in the prior order, because controller-runtime checks for existence of the flag, so kyverno just needs to do the same now based on the updated import paths with flux. @vishal-chdhry Thanks for the tour of this issue 👍 

The redefinition is between controller-runtime and kyverno, it's a fun coincidence that fluxcd/pkg causes this, but the indirect dependency explanation appears to be the correct one.

PR #646, opened 2023-09-06
oci: Use controller-runtime pkg/log specifically

darkowlzz:
This helps avoid importing the controller-runtime pkg/client/config package which has a [flag initialization](https://github.com/kubernetes-sigs/controller-runtime/blob/2a553d6f910dfc4b0796142ba1e059639a21c54a/pkg/client/config/config.go#L44) for "kubeconfig". This results in all the users of the oci package to also have the flag initialized in their applications. The root level controller-runtime package has type aliases which we use a lot for ease of use, but the package also [imports the client config package](https://github.com/kubernetes-sigs/controller-runtime/blob/2a553d6f910dfc4b0796142ba1e059639a21c54a/alias.go#L23).
The usage of controller-runtime in oci package is just for logging. Importing the pkg/log package specifically helps avoid importing the client config which sets the flag.

Fixes https://github.com/fluxcd/pkg/issues/645

Verified by building kyverno locally and testing.

GCP integration test run: https://github.com/fluxcd/pkg/actions/runs/6098520241/job/16548522220

### DIFF
--- a/oci/auth/aws/auth.go
+++ b/oci/auth/aws/auth.go
@@ -29,7 +29,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/ecr"
 	"github.com/google/go-containerregistry/pkg/authn"
-	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/fluxcd/pkg/oci"
 )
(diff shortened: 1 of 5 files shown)


makkes (approved):
Perfect, thanks. I was able to verify that it works as expected when imported.

makkes:
@darkowlzz we should add a test case that prevents a regression here.

makkes:
> But this is a very general problem for all the packages that use controller-runtime

We can start small by adding a tests for this one. No need to make this just another huge project.

realshuting:
Is there an ETA when this fix will be merged? We (Kyverno) encountered this issue while importing some of the fluxcd packages, would love to get this in!

stefanprodan:
@realshuting you can now use `github.com/fluxcd/pkg/oci v0.31.1`

realshuting:
Thank you @stefanprodan!

PR #648, opened 2023-09-11
Add support for loading UTF16 encoded files

souleb:
Add support for loading UTF16 encoded files if they have a correct BOM (Byte-Order-Mark https://en.wikipedia.org/wiki/Byte_order_mark) at the beginning of the file. Falls back on UTF8 encoding, if no understandable BOM is present.

This is dependent on go-yaml support, utf8, utf16LE and utf16BE. see: https://github.com/kubernetes-sigs/kustomize/blob/7c36ed21b3587e7124326743a888db5374e86d93/kyaml/internal/forked/github.com/go-yaml/yaml/readerc.go#L181C4-L181C4


kustomize/kustomize_generator_whitebox_test.go
```diff
@@ -49,6 +49,14 @@ func TestScanManifests(t *testing.T) {
 			base:    "./testdata/nokustomization/panic",
 			wantErr: true,
 		},
+		{
+			name: "utf-16LE encoding",
```
makkes, line 53: Can we give this test a more descriptive name, please? What does it actually test? Something like "Files with UTF-16LE encoding should be included".

makkes: Plus: This tests passes just fine without the changes to kustomize_generator.go.

souleb:
I have tested this with an `utf16-LE` encoded yaml without a prepended `BOM`. Windows programs sometimes do that...

Well it fails with `failed to decode Kubernetes YAML from testdata/nokustomization/utf16le/secret.yaml: MalformedYAMLError: yaml: control characters are not allowed`. This is to be expected as we prepend `utf8` `BOM` when it is missing.

This is not perfect but it is still better than the actual behaviour where it silently drop those files, because it provides an actionable error.

Note that it is a breaking change because running workload  containing `utf16-LE without BOM` will start erroring.

makkes:
> I have tested this with an `utf16-LE` encoded yaml without a prepended `BOM`. Windows programs sometimes do that...
> 
> Well it fails with `failed to decode Kubernetes YAML from testdata/nokustomization/utf16le/secret.yaml: MalformedYAMLError: yaml: control characters are not allowed`. This is to be expected as we prepend `utf8` `BOM` when it is missing.

I manually reverted your changes to `kustomize_generator.go` and ran the new test case and it passes. What am I missing?

souleb:
> > I have tested this with an `utf16-LE` encoded yaml without a prepended `BOM`. Windows programs sometimes do that...
> > Well it fails with `failed to decode Kubernetes YAML from testdata/nokustomization/utf16le/secret.yaml: MalformedYAMLError: yaml: control characters are not allowed`. This is to be expected as we prepend `utf8` `BOM` when it is missing.
> 
> I manually reverted your changes to `kustomize_generator.go` and ran the new test case and it passed. What am I missing?

Yes you are right, but in practice we have seen that it drops silently those files. I have to dig deeper. Let me put this on draft.

PR #653, opened 2023-09-22
oci/client: pull any artifact

errordeveloper:


oci/client/client.go
```diff
@@ -44,13 +43,7 @@ func NewClient(opts []crane.Option) *Client {
 // DefaultOptions returns an array containing crane.WithPlatform
 // to set the platform to flux.
```
stefanprodan, line 44: Please update the description

stefanprodan (commented):
Can we have a test that proves "pull any artifact"? Maybe push something with crane in tests, no annotations from empty base and test pull.

errordeveloper:
@stefanprodan I added a simple test, PTAL :)

oci/client/pull_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	dst := fmt.Sprintf("%s/%s:%s", dockerReg, repo, tag)
+
+	artifact := filepath.Join(t.TempDir(), "artifact.tgz")
+	g.Expect(c.Build(artifact, testDir, nil)).To(Succeed())
+
+	img := mutate.MediaType(empty.Image, types.OCIManifestSchema1)
+	img = mutate.ConfigMediaType(img, oci.CanonicalConfigMediaType)
+
+	layer, err := tarball.LayerFromFile(artifact, tarball.WithMediaType(oci.CanonicalContentMediaType))
```
stefanprodan, line 51: Lets use different media type e.g. `application/vnd.acme.some.content.layer.v1.tar+gzip`

hiddeco (approved):

stefanprodan (approved):
LGTM

Thanks @errordeveloper 🏅

Issue #538, opened 2023-04-19
Untar does not support symbolic links


rrileyca:
## Summary
When creating an `OCIRepository` to monitor the following docker repo: https://artifacthub.io/packages/helm/kiwigrid/spring-cloud-config-server, I get an error saying `failed to extract layer contents from artifact: tar file entry bin/bzcmp contained unsupported file type Lrwxrwxrwx`.

The errors at `loglevel=debug` are:
```
[...]

PR #655, opened 2023-09-29
Skip symlinks during OCI layer decompression

stefanprodan:
This is a followup of https://github.com/fluxcd/pkg/pull/653 as pulling any OCI artifact currently fails if it contains symlinks.

Changes:
- Add `WithSkipSymlinks` option to `tar.Untar`
- Add test to prove that tarballs with symlinks no longer error out during decompression if skip is enabled
- Use `WithSkipSymlinks` in `oci/client.Pull` so that the Flux CLI will not error out if it finds symlinks in the tarball, and instead it will skip them
- Use `WithSkipSymlinks` in `http/ArchiveFetcher.Fetch` to avoid failing to extract artifacts which are not recompiled by source-controller

Ref: https://github.com/fluxcd/pkg/issues/538

### DIFF
--- a/http/fetch/archive_fetcher.go
+++ b/http/fetch/archive_fetcher.go
@@ -140,7 +140,7 @@ func (r *ArchiveFetcher) Fetch(archiveURL, digest, dir string) error {
 	}
 
 	// Extracts the tar file.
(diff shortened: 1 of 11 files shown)


hiddeco (approved):

rashedkvm (commented):
lgtm

PR #657, opened 2023-10-03
ssa: introduce `jsondiff` package

hiddeco:
TODO:

- [x] More test coverage around selectors, and potentially other things
- [x] Back-porting of e.g. HPA fixes which do exist in the diff logic of the `ResourceManager`: https://github.com/fluxcd/pkg/pull/658
- [x] Add configuration option for `jsondiff.Rationalize()` (as this may not be desired by all consumers)
- [x] Documentation (within commit message)

### DIFF
--- a/ssa/jsondiff/diff.go
+++ b/ssa/jsondiff/diff.go
@@ -22,29 +22,29 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
-// ChangeType is the type of change detected by the server-side apply diff
+// DiffType is the type of change detected by the server-side apply diff
 // operation.
-type ChangeType string
+type DiffType string
 
 const (
-	// ChangeTypeCreate indicates that the resource does not exist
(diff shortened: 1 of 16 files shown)


ssa/jsondiff/diff.go
```diff
@@ -0,0 +1,74 @@
...
+		Type:             t,
+		GroupVersionKind: obj.GetObjectKind().GroupVersionKind(),
+		Namespace:        obj.GetNamespace(),
+		Name:             obj.GetName(),
+		Patch:            p,
+	}
+}
+
+// ChangeSet is a list of changes.
+type ChangeSet []*Change
```
stefanprodan, line 74: Let's rename `ChangeSet` to `DiffSet` and `Change` to `Diff` to avoid naming collision with `ssa.ChangeSet`.

stefanprodan (approved):
LGTM

Awesome work @hiddeco 🥇

PR #658, opened 2023-10-04
ssa: object normalization improvements

hiddeco:


ssa/normalize.go
```diff
@@ -0,0 +1,161 @@
...
+	// Ensure the object has an empty creation timestamp, to avoid
+	// issues with the Kubernetes API server rejecting the object
+	// or causing any spurious diffs.
+	unstructured.SetNestedField(object.Object, nil, "metadata", "creationTimestamp")
+
+	// To ensure kstatus continues to work with CRDs, we need to keep the
+	// status field for CRDs.
+	if !IsCRD(object) {
+		unstructured.RemoveNestedField(object.Object, "status")
+	}
```
hiddeco, line 124: I have my doubts about this actually being required at this point, as we at least seem to do (part of) this again in `manager_diff.go`. If the goal here is to prepare the resource for an apply, wouldn't it be better for this to be done as a preparation step _before_ the apply?

hiddeco: Removing it also does not appear to make any of the tests fail, except for the `NormalizeUnstructured` tests I added myself.

ssa/utils.go
```diff
@@ -318,248 +322,12 @@ func AnyInMetadata(object *unstructured.Unstructured, metadata map[string]string
...
 }
 
-// SetNativeKindsDefaults implements workarounds for server-side apply upstream bugs affecting Kubernetes < 1.22
-// ContainerPort missing default TCP proto: https://github.com/kubernetes-sigs/structured-merge-diff/issues/130
-// ServicePort missing default TCP proto: https://github.com/kubernetes/kubernetes/pull/98576
-// PodSpec resources missing int to string conversion for e.g. 'cpu: 2'
-// secret.stringData key replacement add an extra key in the resulting data map: https://github.com/kubernetes/kubernetes/issues/108008
+// SetNativeKindsDefaults sets default values for native Kubernetes objects,
+// working around various upstream Kubernetes API bugs.
 func SetNativeKindsDefaults(objects []*unstructured.Unstructured) error {
```
hiddeco, line 327: Theoretically, we could introduce a `NormalizeUnstructuredList(objects []*unstructured.Unstructured)` and deprecate this one.

stefanprodan: With all the changes to follow I think we can go with a major version bump for SSA, so I would rename the function.

ssa/normalize.go
stefanprodan: If we remove the status for all including CRD, then how are we going to add it back before apply?

hiddeco: My point is that we would not remove (or modify) anything here at all, we would just normalize the object as-is, and if required, prepare it right before an apply and/or diff.

Due to a lack of test cases covering the scenario in which the existing modification would be required, I can however not be a good judge here.

stefanprodan: We have such a test, the that reconciles CRDs and CRs

hiddeco: It continues to work if I remove L112-121.

hiddeco: As this needs much more testing, I will (eventually) do the remainder of this in a separate PR.

stefanprodan (approved):
LGTM

Great refactoring, thanks @hiddeco 🥇

PR #661, opened 2023-10-06
ssa: improve data masking

hiddeco:
Instead of masking the data with `****` or `*****`, provide an easier to distinguish `*** (before)` and `*** (after)` for values which have changed. Matching the output of `kubectl diff`.

In addition, make the function public as it can be of use outside of the `Diff` of the resource manager, and allow passing only one instance of an object (for scenarios where a resource is either created or deleted).


stefanprodan (approved):
LGTM

Thanks @hiddeco 🎖️

PR #663, opened 2023-10-10
Address various nits throughout packages

hiddeco:


oci/client/login.go
```diff
@@ -81,7 +81,7 @@ func (c *Client) LoginWithProvider(ctx context.Context, url string, provider oci
 	case oci.ProviderAzure:
 		authenticator, err = azure.NewClient().Login(ctx, true, url, ref)
 	default:
-		return errors.New(fmt.Sprintf("unsupported provider"))
+		return errors.New("unsupported provider")
```
aryan9600, line 84:
```suggestion
		return fmt.Errorf("unsupported provider: '%s',  provider)
```

makkes (approved):

oci/client/login.go
hiddeco: It is an `int` so this wouldn't work.

PR #664, opened 2023-10-13
Add new package `auth` for various authentication scenarios

### LINKED ISSUE #642: Cache registry credentials for cloud providers
We support using workload identity to enable passwordless logins for ACR, ECR and GCR. This is done by fetching an authentication token from the cloud provider. But, this token is not cached, which means we fetch a new token during every reconciliation. We should add a cache for each cloud provider and implement eviction and refreshing logic after the cached token expires.

aryan9600:
Add a new package `auth` which handles multiple authentication use cases for various cloud providers and SaaS. It inculdes the following sub packages:
* `aws`: For AWS and ECR auth
* `gcp`: For GCP and GAR auth
* `azure`: For Azure and ACR auth
* `github`: For GitHub Apps auth
* `registry`: For OCI registry auth
* `git`: For auth against Git providers

The packages `auth/registry` and `auth/git` also support caching the authentication credentials via `AuthOptions.CacheOptions.Key`. Caching needs to explicitly enabled by either calling `auth.InitCache()` or specifying a cache via `AuthOptions.CacheOptions.Cache`.

Fixes #642


auth/cache.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+var once sync.Once
+var cache Store
+
+// InitCache intializes the pacakge cache with the provided cache object.
+// Consumers that want automatic caching when using `GetRegistryAuthenticator()`
+// or `GetGitCredentials()` must call this before. It should only be called once,
+// all subsequent calls will be a no-op.
+func InitCache(s Store) {
```
hiddeco, line 31: I think it would be good if this returned an error for subsequent calls, to indicate it was already set.

auth/options.go
```diff
@@ -0,0 +1,57 @@
...
+	Secret *corev1.Secret
+
+	// ProviderOptions specifies the options to configure various authentication
+	// providers.
+	ProviderOptions ProviderOptions
+
+	// CacheKey is the key to use for caching the authentication credentials.
+	// Consumers must make sure to call `InitCache()` in order for caching to
+	// be enabled.
+	CacheKey string
```
hiddeco, line 47: I think it would be relatively easy to allow providing a custom `Cache` here instead of relying on the global default.

Poltergeisen:
Is this still planned for the next minor release of flux? I'm looking forward to using this soon if possible to connect to Azure Devops Git.

ageisen2000:
Will this work with the flux bootstrap command when implemented?

stefanprodan:
> Also awaiting this functionality, any updates?

We used to have an Azure account to test all of this but MSFT has canceled our subscription. Also https://github.com/fluxcd/flux2/discussions/4544

evandam:
I'm just wondering if it would be easier/more manageable if this was broken down a bit? For example it would probably be easier to test GitHub App auth and making related changes to the GitHub alert provider without worrying about AWS/Azure/GCP mixed in too?

Granted I'm a bit biased since I opened https://github.com/fluxcd/flux2/discussions/4356 and would love to see it implemented. No worries though and appreciate the effort everyone puts in to make Flux awesome!

AdrianBalcan:
Hi guys, this feature can be a solution differentiator for FluxCD over other solutions. It could bring full automation IaC between terraform AKS deployment and FluxCD which would populate the Kubernetes cluster and removing manual steps.

yashwanth-l:
Hello!

Any updates further on this issue?

This feature could be really help in avoiding using PAT and SSH Keys (GH doesn't even allow the same keys in multiple repos)

stefanprodan:
Closing, this is being implemented in stages in https://github.com/fluxcd/pkg/tree/main/auth

Poltergeisen:
Is there an issue in this new repository to track workload identity auth for azure DevOps?

stefanprodan:
@Poltergeisen the Azure DevOps OIDC-auth was released back in September. I suggest you subscribe to the Flux releases in the [flux2 repo](https://github.com/fluxcd/flux2). Here is the blog post: https://fluxcd.io/blog/2024/09/flux-v2.4.0/

PR #675, opened 2023-10-25
[oci] Extend oci.Push to create layer with static uncompressed contents

somtochiama:
This pull request adds different layer types when pushing an OCI image. The default is the `tarball` where the layer contains a tarball created from the sourcePath. The `static` type creates a layer containing the uncompressed contents of the file.

It also adds an option for configuring the media type extension of the layer and reactors the Push methods with functional options.

### DIFF
--- a/oci/client/push.go
+++ b/oci/client/push.go
@@ -28,45 +28,100 @@ import (
 	gcrv1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/empty"
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/static"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 
 	"github.com/fluxcd/pkg/oci"
 )
 
-// Push creates an artifact from the given directory, uploads the artifact
+// LayerType is an enumeration of the supported layer types
+// when pushing an image.
+type LayerType string
+
+const (
+	// LayerTypeTarball produces a layer that contains a gzipped archive
+	LayerTypeTarball LayerType = "tarball"
+	// LayerTypeStatic produces a layer that contains the contents of a
+	// file without any compression.
+	LayerTypeStatic LayerType = "static"
+)
+
+// PushOptions are options for configuring the Push operation.
+type PushOptions struct {
+	layerType LayerType
+	layerOpts layerOptions
+	meta      Metadata
+}
+
+// layerOptions are options for configuring a layer.
+type layerOptions struct {
+	mediaTypeExt string
+	ignorePaths  []string
+}
+
+// PushOption is a function for configuring PushOptions.
+type PushOption func(o *PushOptions)
+
+// WithPushLayerType set the layer type that will be used when creating
+// the image layer.
+func WithPushLayerType(l LayerType) PushOption {
+	return func(o *PushOptions) {
+		o.layerType = l
+	}
+}
+
+// WithPushMediaTypeExt configures the media type extension for the image layer.
+// This is only used when the layer type is `LayerTypeStatic`.
+// The final media type will be prefixed with `application/vnd.cncf.flux.content.v1`
+func WithPushMediaTypeExt(extension string) PushOption {
+	return func(o *PushOptions) {
+		o.layerOpts.mediaTypeExt = extension
+	}
+}
+
+// WithPushIgnorePaths configures ignore paths for PushOptions
+func WithPushIgnorePaths(paths ...string) PushOption {
+	return func(o *PushOptions) {
+		o.layerOpts.ignorePaths = append(o.layerOpts.ignorePaths, paths...)
+	}
+}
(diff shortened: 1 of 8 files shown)


oci/client/push.go
```diff
@@ -34,36 +34,101 @@ import (
...
+			return "", err
+		}
+	} else {
+		// we only want a single file when we aren't archiving
+		f, err := os.Stat(sourceDir)
+		if err != nil {
+			return "", err
+		}
+		if f.IsDir() {
+			return "", fmt.Errorf("can only specify files when not compressing")
```
stefanprodan, line 109:
```suggestion
			return "", fmt.Errorf("can only specify a single file when not compressing")
```

oci/client/push.go
```diff
@@ -34,36 +34,101 @@ import (
...
+		}
+		defer os.RemoveAll(tmpDir)
+
+		layerFile = filepath.Join(tmpDir, "artifact.tgz")
+		if err := c.Build(layerFile, sourceDir, o.ignorePaths); err != nil {
+			return "", err
+		}
+	} else {
+		// we only want a single file when we aren't archiving
+		f, err := os.Stat(sourceDir)
```
stefanprodan, line 104: Let's rename `sourceDir` to `sourcePath`.

oci/client/push.go
```diff
@@ -34,36 +34,101 @@ import (
...
+// argument must be a file and not a directory.
+func WithSkipArchive() PushOption {
+	return func(o *PushOptions) {
+		o.skipArchive = true
+	}
+}
+
+// WithMediaType configures the media type for PushOptions
+// The final media type will be prefixed with `application/vnd.cncf.flux.content.v1`
+func WithMediaType(mediaType string) PushOption {
```
stefanprodan, line 59:
```suggestion
func WithMediaTypeExtension(extension string) PushOption {
```

oci/client/push.go
```diff
@@ -34,36 +34,101 @@ import (
...
+	} else {
+		// we only want a single file when we aren't archiving
+		f, err := os.Stat(sourcePath)
+		if err != nil {
+			return "", err
+		}
+		if f.IsDir() {
+			return "", fmt.Errorf("can only specify a single file when not compressing")
+		}
+		layerFile = sourcePath
```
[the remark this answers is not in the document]
somtochiama: I don't think it requires a separate function since all we want to do is check if the path is a directory. It can also be tested with the current setup.

oci/client/push.go
```diff
@@ -34,36 +34,101 @@ import (
 	"github.com/fluxcd/pkg/oci"
 )
 
+// PushOptions are options for configuring the Push operation.
+type PushOptions struct {
+	skipArchive  bool
```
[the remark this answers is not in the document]
somtochiama: I think archive is a better term as we are skipping archiving the directory into a tarball. The layer is still compressed.

oci/client/push_pull_test.go
```diff
@@ -37,74 +38,123 @@ func Test_Push_Pull(t *testing.T) {
...
+
+		// Build and push the artifact to registry
+		_, err := c.Push(ctx, url, tt.dir, opts...)
+		g.Expect(err).ToNot(HaveOccurred())
+
+		// Verify that the artifact and its tag is present in the registry
+		tags, err := crane.ListTags(fmt.Sprintf("%s/%s", dockerReg, repo))
+		g.Expect(err).ToNot(HaveOccurred())
+		g.Expect(len(tags)).To(BeEquivalentTo(i + 1))
+		g.Expect(tags[i]).To(BeEquivalentTo(tt.tag))
```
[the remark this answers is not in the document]
somtochiama: I have changed to use a function from gomega:
```
g.Expect(tags).To(ContainElement(tt.tag))
```

oci/client/push.go
```diff
@@ -84,3 +146,28 @@ func (c *Client) Push(ctx context.Context, url, sourceDir string, meta Metadata,
...
 	return ref.Context().Digest(digest.String()).String(), err
 }
+
+// createTarballLayer creates a layer from a gzipped tarball of a directory.
+func (c *Client) createTarballLayer(path string, mediaType types.MediaType, dir string, ignorePaths []string) (gcrv1.Layer, error) {
+	layerFile := filepath.Join(dir, "artifact.tgz")
+	if err := c.Build(layerFile, path, ignorePaths); err != nil {
+		return nil, err
+	}
+	return tarball.LayerFromFile(layerFile, tarball.WithMediaType(mediaType))
```
stefanprodan, line 156:
```suggestion
	return tarball.LayerFromFile(layerFile, tarball.WithMediaType(mediaType), tarball.WithCompression(compression.GZip), tarball.WithCompressedCaching)
```

We don't push a plain tarball, but a compressed one.

somtochiama: > tarball.WithCompression(compression.GZip)

This is the default compression. I think we can leave it out

stefanprodan (approved):
LGTM

Thanks @somtochiama 🥇

stefanprodan:
@somtochiama next please do `pull` for static content, where the layer is just streamed to the destination dir without decompression.

PR #678, opened 2023-10-26
Upgrade go-git to 5.10.0

makkes:
This will fix https://github.com/fluxcd/source-controller/issues/1239


pjbgf (approved):
Nice one following up on this @makkes. :bow:

stefanprodan (approved):
🎉

makkes:
> With go-git returning proper error for remote errors, do we still need this https://github.com/fluxcd/pkg/blob/git/gogit/v0.14.2/git/gogit/clone.go#L615 anymore?

Yes, as of go-git 5.10.0 the library shouldn't return such errors, anymore, so `goGitError` would never replace anything.

makkes:
Sorry for the confusion. Next try:

As of go-git 5.10.0 the library shouldn't return such errors, anymore, so `goGitError` would never replace anything. For this reason we can remove `goGitError`.

PR #691, opened 2023-11-27
ssa/jsondiff: cluster and desired object in Diff

hiddeco:
By including the objects in full, it becomes possible to for example apply the JSON Patch without having to retrieve the object again.

In addition, it becomes possible to e.g. render a diff including any ignore rules (using dyff) by applying the JSON Patch to a copy of the cluster object.

Note that when you want to use the JSON Patch result to perform an actual patch operation, any Secret masking should be done while displaying information and not using the available option. As otherwise the patch would be issued with masked data (and potentially illegal base64 payloads).

### DIFF
--- a/ssa/jsondiff/diff.go
+++ b/ssa/jsondiff/diff.go
@@ -18,8 +18,8 @@ package jsondiff
 
 import (
 	"github.com/wI2L/jsondiff"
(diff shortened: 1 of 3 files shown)


stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #692, opened 2023-11-27
ssa: prevent unnecessary `DeepCopy`

hiddeco:
When attempting to retrieve the object, setting the GVK on the Unstructured is sufficient to be able to retrieve it from the cluster.

This avoids deep-copying the object in full and then `nil` writing it again when it does not exist in the cluster. Which can potentially save quite some duplicate data when e.g. dealing with Custom Resource Definitions or other large object kinds.


stefanprodan (approved):
LGTM

Great optimisation, thanks @hiddeco 🎖️

PR #694, opened 2023-11-28
Use `gomega` in sourceignore package

stefanprodan:
Changes:
- Update `golang.org` dependencies in all packages
- Replace `gotest.tools/assert` with gomega for sourceignore tests


makkes (approved):
🙌🏻

Issue #696, opened 2023-12-01
Extend `ssa.Apply` with field ignore rules


stefanprodan:
To make kustomize-controller compatible with VPA and other controllers which mutate Flux-managed fields in-cluster, we need to wire the `ssa/jsondiff` logic into the SSA apply functions.

Tasks:
- Move the `saa` functions used in `jsondiff` to a dedicated package 
- Add `[]jsondiff.IgnoreRule` to the `ssa.ApplyOptions`
- Remove the ignored fields before the final apply

Example:

```go
	appliedObject := object.DeepCopy()

	if existingObject.GetResourceVersion() != "" {
		var sm = make(map[*jsondiff.SelectorRegex][]string, len(opts.DriftIgnoreRule))
		for _, ips := range opts.DriftIgnoreRule {
			sr, err := jsondiff.NewSelectorRegex(ips.Selector)
			if err != nil {
				return nil, fmt.Errorf("failed to create ignore rule selector: %w", err)
			}
			sm[sr] = ips.Paths
		}

		var ignorePaths jsondiff.IgnorePaths
		for sr, paths := range sm {
			if sr.MatchUnstructured(appliedObject) {
				ignorePaths = append(ignorePaths, paths...)
			}
		}

		if len(ignorePaths) > 0 {
			patch := jsondiff.GenerateRemovePatch(ignorePaths...)
			if err := jsondiff.ApplyPatchToUnstructured(appliedObject, patch); err != nil {
				return nil, err
			}
		}
	}

	if err := m.apply(ctx, appliedObject); err != nil {
		return nil, fmt.Errorf("%s apply failed: %w", FmtUnstructured(appliedObject), err)
	}
```

alxbse:
Would like to contribute to getting this solved as flux has been great for my use cases until I now need to manage a CRD with a conversion webhook and a cert-manager caBundle that would keep getting removed/added.

Is it "just" a lack of priority and contributions that is blocking this issue or is there something else related to adding the ignore rules that makes this extra complicated to solve?

stefanprodan:
This has been already implemented in HelmRelease. I don't think we should come up with a different API, IMO the Flux Kustomization will have a `.spec.driftDetection.ignore` with the array of JSON pointers.

PR #697, opened 2023-12-01
ssa: avoid potential cyclic import

hiddeco:
This moves things into separate packages to avoid a potential cyclic import as soon as we would like to utilize `jsondiff` in `ssa` itself.

Part of: https://github.com/fluxcd/pkg/issues/696


stefanprodan (approved):
LGTM

Thanks @hiddeco 🏅

PR #705, opened 2023-12-12
Bump kustomize GH Action to 5.3.0

stefanprodan:
Sync with Flux 2.2.0

### DIFF
--- a/actions/kustomize/action.yml
+++ b/actions/kustomize/action.yml
@@ -6,9 +6,9 @@ branding:
   icon: command
 inputs:
   version:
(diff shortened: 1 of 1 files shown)


actions/kustomize/action.yml
```diff
@@ -49,6 +49,10 @@ runs:
 
           echo "Downloading kustomize ${VERSION} for ${OS}/${ARCH}"
           KUSTOMIZE_TARGET_FILE="kustomize_${VERSION}_${OS}_${ARCH}.tar.gz"
+          if [[ "$OS" == "windows" ]]; then
```
hiddeco, line 52: For backwards compatibility, this should likely also introduce a version check :grimacing:

hiddeco (approved):

PR #706, opened 2023-12-13
http/fetch: allow configuring a logger on client

hiddeco:
This allows for the configuration of a logger on the retryable HTTP client. While adding builtin support for providing a `logr.Logger` instance, which will be wrapped in  a `retryablehttp.LeveledLogger` to only log errors.

Part of: https://github.com/fluxcd/flux2/issues/4472

### DIFF
--- a/http/fetch/go.mod
+++ b/http/fetch/go.mod
@@ -14,6 +14,7 @@ replace github.com/opencontainers/go-digest => github.com/opencontainers/go-dige
 require (
 	github.com/fluxcd/pkg/tar v0.4.0
 	github.com/fluxcd/pkg/testserver v0.5.0
+	github.com/go-logr/logr v1.3.0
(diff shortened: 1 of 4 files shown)


http/fetch/go.mod
```diff
@@ -14,6 +14,7 @@ replace github.com/opencontainers/go-digest => github.com/opencontainers/go-dige
 require (
 	github.com/fluxcd/pkg/tar v0.4.0
 	github.com/fluxcd/pkg/testserver v0.5.0
+	github.com/go-logr/logr v1.2.4
```
stefanprodan, line 17: Please bump it to v1.3.0 like we have in runtime https://github.com/fluxcd/pkg/blob/0603b64045a617a2599566baed7f868c55cf561c/runtime/go.mod#L18

stefanprodan (approved):
LGTM

PR #707, opened 2023-12-13
runtime: Trim the condition message to the max accepted length

stefanprodan:
xref: https://github.com/fluxcd/flux2/issues/4470

### DIFF
--- a/runtime/conditions/setter.go
+++ b/runtime/conditions/setter.go
@@ -28,8 +28,9 @@ import (
 	"sort"
 	"time"
 
-	"github.com/fluxcd/pkg/apis/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/fluxcd/pkg/apis/meta"
 )
 
 // Setter is an interface that defines methods a Kubernetes object should implement in order to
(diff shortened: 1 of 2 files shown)


runtime/conditions/setter.go
```diff
@@ -223,3 +227,15 @@ func hasSameState(i, j *metav1.Condition) bool {
 		i.Reason == j.Reason &&
 		i.Message == j.Message
 }
+
+const maxMessageLength = 32768
+
+// trimConditionMessage trims the condition message to a maximum length of 32K characters.
+func trimConditionMessage(msg string) string {
```
hiddeco, line 234: Would be good to have some test coverage for this, as the reslicing can later on become a potential victim to out-of-bound errors.

To not have to test the full `maxMessageLength`, you could change it to `trimConditionMessage(msg string, max int)` and do `condition.Message = trimConditionMessage(condition.Message, maxMessageLength)` instead.

runtime/conditions/setter.go
```diff
@@ -223,3 +227,15 @@ func hasSameState(i, j *metav1.Condition) bool {
 		i.Reason == j.Reason &&
 		i.Message == j.Message
 }
+
+const maxMessageLength = 32768
```
hiddeco, line 231:
```suggestion
const (
    maxMessageLength = 32768
    trimmedMessageSufix = "..."
)
```

runtime/conditions/setter.go
```diff
@@ -223,3 +227,15 @@ func hasSameState(i, j *metav1.Condition) bool {
...
+
+// trimConditionMessage trims the condition message to a maximum length of 32K characters.
+func trimConditionMessage(msg string) string {
+	if len(msg) <= maxMessageLength {
+		return msg
+	}
+
+	trimmedMsg := msg[:maxMessageLength-3] + "..."
```
hiddeco, line 239:
```suggestion
	trimmedMsg := msg[:maxMessageLength-len(trimmedMessageSuffix)] + trimmedMessageSuffix
```

hiddeco (approved):

PR #708, opened 2023-12-17
Fix a silently dropping Kustomization remote resources

allenporter:
Fix an issue where kustomizations with remote resources are silently dropped from `flux build ks`. The behavior change appears to have been introduced in https://github.com/fluxcd/flux2/pull/3763 and https://github.com/fluxcd/pkg/pull/528 when adding filters to ignore the source path. The behavior skips urls since they don't match file paths, but that had the side effect of not including them in the generated `Kustomization`. This PR reverts to the old behavor of ignoring remote urls and keeping them in the kustomization -- which is fine since source ignore paths are about paths in the source tree and not remote resources.

Fixes https://github.com/fluxcd/flux2/issues/4146 which describes how to reproduce the issue. An example test is included that exercises the issue ensuring the remote resource is included in the filtered kustomization, though does not actually try to build a remote resource.

### DIFF
--- a/kustomize/filters.go
+++ b/kustomize/filters.go
@@ -81,23 +81,25 @@ func filterKsWithIgnoreFiles(ks *kustypes.Kustomization, dirPath string, ignore 
 	return nil
 }
 
+func isUrl(s string) bool {
+	u, err := url.ParseRequestURI(s)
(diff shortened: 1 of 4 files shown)


hiddeco (approved):
Thank you @allenporter :bow:

Issue #715, opened 2024-01-16
Improve kstatus output for `failed early due to stalled resources`


stefanprodan:
When waiting for resources exits early due to stalled resources, the returned list should only contain the stalled resources without the ones in an Unknown state.

Source: https://github.com/fluxcd/pkg/blob/main/ssa/manager_wait.go

Ref: https://github.com/fluxcd/flux2/issues/4529

PR #717, opened 2024-01-16
ignore 'unknown' state of stalled resources

funkymcb:
This would be my idea of fixing issue #715.

### DIFF
--- a/ssa/manager_wait.go
+++ b/ssa/manager_wait.go
@@ -139,8 +139,11 @@ func (m *ResourceManager) WaitForSet(set object.ObjMetadataSet, opts WaitOptions
 				errors = append(errors, fmt.Sprintf("%s (unknown status)", utils.FmtObjMetadata(rs.Identifier)))
 			} else if lastStatus[id].Status != status.CurrentStatus {
 				var builder strings.Builder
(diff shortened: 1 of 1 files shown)


ssa/manager_wait.go
```diff
@@ -139,8 +139,11 @@ func (m *ResourceManager) WaitForSet(set object.ObjMetadataSet, opts WaitOptions
 				errors = append(errors, fmt.Sprintf("%s (unknown status)", utils.FmtObjMetadata(rs.Identifier)))
 			} else if lastStatus[id].Status != status.CurrentStatus {
 				var builder strings.Builder
-				builder.WriteString(fmt.Sprintf("%s status: '%s'",
-					utils.FmtObjMetadata(rs.Identifier), lastStatus[id].Status))
+				if strings.ToLower(lastStatus[id].Status.String()) != "unknown" {
+					// ignore status 'unknown' to only list failed resources
+					builder.WriteString(fmt.Sprintf("%s status: '%s'",
+						utils.FmtObjMetadata(rs.Identifier), lastStatus[id].Status))
+				}
```
stefanprodan, line 146: This needs to happen only if there are failed resources, otherwise you would not know which resources have timed out and the error will come back empty.

stefanprodan:
Superseded by #722

PR #722, opened 2024-01-19
ssa: Improve wait error reporting

stefanprodan:
If the context deadline has not been reached, include only the failed resources in the returned error. This makes the health check error message concise, instead of listing potentially hundreds of resources in progress, we only list the ones that actually failed.   

Part of: https://github.com/fluxcd/flux2/issues/4556
Fixes: https://github.com/fluxcd/pkg/issues/715
Supersedes: https://github.com/fluxcd/pkg/pull/717

### DIFF
--- a/ssa/manager_wait.go
+++ b/ssa/manager_wait.go
@@ -19,6 +19,7 @@ package ssa
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -83,7 +84,6 @@ func (m *ResourceManager) WaitForSet(set object.ObjMetadataSet, opts WaitOptions
 	eventsChan := m.poller.Poll(ctx, set, pollingOpts)
 
 	lastStatus := make(map[object.ObjMetadata]*event.ResourceStatus)
-	var failedResources int
 
 	done := statusCollector.ListenWithObserver(eventsChan, collector.ObserverFunc(
 		func(statusCollector *collector.ResourceStatusCollector, e event.Event) {
@@ -96,7 +96,7 @@ func (m *ResourceManager) WaitForSet(set object.ObjMetadataSet, opts WaitOptions
 				// skip DeadlineExceeded errors because kstatus emits that error
 				// for every resource it's monitoring even when only one of them
 				// actually fails.
-				if rs.Error != context.DeadlineExceeded {
+				if !errors.Is(rs.Error, context.DeadlineExceeded) {
 					lastStatus[rs.Identifier] = rs
 				}
 
@@ -105,7 +105,6 @@ func (m *ResourceManager) WaitForSet(set object.ObjMetadataSet, opts WaitOptions
 				}
 				rss = append(rss, rs)
 			}
-			failedResources = countFailed
 
 			desired := status.CurrentStatus
 			aggStatus := aggregator.AggregateStatus(rss, desired)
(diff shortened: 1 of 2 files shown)


ssa/manager_wait.go
```diff
@@ -122,32 +121,36 @@ func (m *ResourceManager) WaitForSet(set object.ObjMetadataSet, opts WaitOptions
...
-		if ctx.Err() == context.DeadlineExceeded {
-			msg = "timeout waiting for"
-		}
-
-		var errors = []string{}
-		for id, rs := range statusCollector.ResourceStatuses {
-			if rs == nil {
-				errors = append(errors, fmt.Sprintf("can't determine status for %s", utils.FmtObjMetadata(id)))
-				continue
+	var errs []string
```
souleb, line 124: wdyt think about 

```go
 var errs []error
	for id, rs := range statusCollector.ResourceStatuses {
		switch {
		case rs == nil || lastStatus[id] == nil:
			errs = append(errs, fmt.Errorf("resource %s not found", utils.FmtObjMetadata(id)))
		case lastStatus[id].Status == status.FailedStatus:
			var builder strings.Builder
			builder.WriteString(fmt.Sprintf("%s status: '%s'",
				utils.FmtObjMetadata(rs.Identifier), lastStatus[id].Status))
			if rs.Error != nil {
				builder.WriteString(fmt.Sprintf(": %s", rs.Error))
			}
			errs = append(errs, errors.New(builder.String()))
...
	if len(errs) > 0 {
		msg := "failed early due to stalled resources"
		if errors.Is(ctx.Err(), context.DeadlineExceeded) {
			msg = "timeout waiting for"
		}
		return fmt.Errorf("%s: %w", msg, errors.Join(errs...))
	}
```

Wrapping the `errors` can make it easier to inspect for calling funtions.

stefanprodan: This is more about humans than code, keep in mind that the resulting error ends up in Slack, PagerDuty, etc. Is error wrapping with 100+ errors easier to read and parse than an array?

souleb: `e.Error()` will return a string still. The difference is that for `Join`. It says `The error formats as the concatenation of the strings obtained by calling the Error method of each element of errs, with a newline between each string.`

stefanprodan: Ok but this will change the format and if people rely on the current array, their parser will fail. I would consider change it in Flux minor release instead of a patch like now.

ssa/manager_wait.go
```diff
@@ -122,32 +121,36 @@ func (m *ResourceManager) WaitForSet(set object.ObjMetadataSet, opts WaitOptions
...
-					builder.WriteString(fmt.Sprintf(": %s", rs.Error))
-				}
-				errors = append(errors, builder.String())
+			errs = append(errs, builder.String())
+		case errors.Is(ctx.Err(), context.DeadlineExceeded) && lastStatus[id].Status != status.CurrentStatus:
+			var builder strings.Builder
+			builder.WriteString(fmt.Sprintf("%s status: '%s'",
+				utils.FmtObjMetadata(rs.Identifier), lastStatus[id].Status))
+			if rs.Error != nil {
+				builder.WriteString(fmt.Sprintf(": %s", rs.Error))
```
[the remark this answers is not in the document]
stefanprodan: I did like this so it's easier to read, failure vs timeout.

PR #752, opened 2024-03-28
Return a DiffTypeExclude when exclusion annotation is set in cluster

souleb:
We want to enable this behaviour for use cases where different controllers must coordinate in order to mutate a resource.

ref: https://github.com/fluxcd/helm-controller/issues/922


stefanprodan (approved):
LGTM

Thanks @souleb 🥇

PR #753, opened 2024-03-29
Adding a method to retrieve a source artifact.

souleb:
This method will permit retrieving source artifacts from other controllers.


stefanprodan (commented):
I don't think this is needed. All consumers should use the Source interface https://github.com/fluxcd/source-controller/blob/0c81d04c8939903d2861a70c62571208afc9c303/api/v1/source.go#L44

PR #758, opened 2024-04-08
Add envsubst package

stefanprodan:
This PR contains a fork of https://github.com/gomodules/envsubst and incorporates https://github.com/drone/envsubst/pull/34.

This **permanent fork** is needed because envsubst powers a GA feature of Flux, and the Drone's project looks like is no longer maintained.

xref: https://github.com/fluxcd/flux2/issues/4694

### DIFF
--- a/envsubst/README.md
+++ b/envsubst/README.md
@@ -1,0 +1,39 @@
+# envsubst
+
+`github.com/fluxcd/pkg/envsubst` is a Go package for expanding variables in a string using `${var}` syntax.
+Includes support for bash string replacement functions.
+
+This package is a fork of [drone/envsubst](https://github.com/drone/envsubst).
+
+## Supported Functions
+
+| __Expression__                | __Meaning__                                                         |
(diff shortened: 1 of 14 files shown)


makkes:
I think we should accompany this PR with an upstream PR to gomodules/envsubst to add the feature of https://github.com/drone/envsubst/pull/34 so that we can remove this fork from our repo once gomodules/envsubst has been released with the feature.

stefanprodan:
@makkes we can't do that, `gomodules/envsubst` has diverged from upstream and requires all vars to be specified.

makkes:
> @makkes we can't do that, `gomodules/envsubst` has diverged from upstream and requires all vars to be specified. Also we may want to add features specific to Flux, for example functions to cope with YAML numbers & booleans.

Ah, I understand. So this is meant to be a permanent fork. 👍🏻

makkes (approved):

souleb (approved):
LGTM!

PR #761, opened 2024-04-09
kustomize: Add support for `namePrefix` and `nameSuffix`

stefanprodan:
Part of: https://github.com/fluxcd/kustomize-controller/issues/1133


souleb (approved):
LGTM!

PR #765, opened 2024-04-23
ssa: Skip suspended resources from wait result errors

stefanprodan:
Fix: https://github.com/fluxcd/flux2/discussions/4738

### DIFF
--- a/ssa/manager_wait.go
+++ b/ssa/manager_wait.go
@@ -136,6 +136,12 @@ func (m *ResourceManager) WaitForSet(set object.ObjMetadataSet, opts WaitOptions
 			errs = append(errs, builder.String())
 		case errors.Is(ctx.Err(), context.DeadlineExceeded) && lastStatus[id].Status != status.CurrentStatus:
 			var builder strings.Builder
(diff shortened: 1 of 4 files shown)


ssa/manager_wait.go
```diff
@@ -136,6 +136,12 @@ func (m *ResourceManager) WaitForSet(set object.ObjMetadataSet, opts WaitOptions
 			errs = append(errs, builder.String())
 		case errors.Is(ctx.Err(), context.DeadlineExceeded) && lastStatus[id].Status != status.CurrentStatus:
 			var builder strings.Builder
+			if utils.IsSuspended(lastStatus[id].Resource) {
```
souleb, line 139: Is it safe to do this? If I have several resources deployed by a `kustomization`, and I want them all `healthy` before triggering another dependent `kustomization`, skipping a suspended kustomization might have unforseen consequences.

stefanprodan:
Found some issue with this approach, since we remove the suspended resources at from the result, this means wait has to timeout. We need to find a way to filter the objects before passing them to the wait function... but without DDoS-ing the API and make new queries for each one.

PR #766, opened 2024-04-25
Introduce in-memory caching package

souleb:
This PR will allow caching the authentication credentials retrieved by `pkg/oci/auth`. It should also enable future usage of the caching underlying mechanism.

Part of: #642

## Store design

The store is K/V store that can store arbitrary objects. The store interface is a simplified version of client go store. If an object can be cached there, it should be in our store implementations. This is desirable, because the primary envisioned place for the store usage is during a reconciliation of custom resources retrieved from a shared informer cache.

The `keys`  are generated dynamically with deterministic function. Accepting a function instead of `key strings` enables user to determine themselves the keys uniqueness constraint.

They store must be thread safe, it should support concurrent programs.

There are 2 main uses cases:
- 1. cache authentication credentials. A credentials key is written, read and overwritten multiple times by a single goroutine. If we plan to store a credential and refresh it every 6 hours and read every 30 minutes(reconciliation interval), the read/write ratio is 12:1. This is a likely scenario. Authentication credentials are expirable.
- 2. cache helm indexes. The new `Store` interface and underlying implementation should allow replacing `https://github.com/fluxcd/source-controller/tree/main/internal/cache` which is today used to cache `helm indexes`. A `helm index` is written, read and overwritten multiple times by multiple goroutines. Usually an index is stored and read multiple times (reconciliation of all dependent custom resources * interval) before being overwritten by a new version. It is read intensive. An index is read from local storage before being cached, if the cache is full, we could have several Custom Resources loading the same index in memory, in order to avoid that, we should evict older keys to make room.

Based on the two scenario above, the store should be optimized for `reads`.

We also have a scenario that needs `keys` to expirable and another that need them be evicted based on usage.

Hence two implementations are provided:
- An expirable store for use case 1
- An LRU cache for use case 2


cache/cache.go
```diff
@@ -0,0 +1,328 @@
...
+	if found {
+		return KeyError{object, fmt.Errorf("key already exists")}
+	}
+
+	if c.MaxItems > 0 && len(c.Items) < c.MaxItems {
+		c.set(key, object)
+		return nil
+	}
+
+	return KeyError{object, fmt.Errorf("Cache is full")}
```
stefanprodan, line 108: Instead of erroring out, I think we should evict the key with the closest expiry.

souleb: If I set an `expiration`, it should be guaranteed.

cache/metrics.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+const (
+	// CacheEventTypeMiss is the event type for cache misses.
+	CacheEventTypeMiss = "cache_miss"
+	// CacheEventTypeHit is the event type for cache hits.
+	CacheEventTypeHit = "cache_hit"
+)
```
stefanprodan, line 31: Let's add StatusSuccess & StatusFailed here and use these constants in the instrumentation code.

cache/cache.go
```diff
@@ -0,0 +1,414 @@
...
+	janitor    *janitor[T]
+	closed     bool
+
+	mu sync.RWMutex
+}
+
+var _ Expirable[any] = &Cache[any]{}
+
+// New creates a new cache with the given configuration.
+func New[T any](maxItems int, keyFunc KeyFunc[T], opts ...Options[T]) (*Cache[T], error) {
```
stefanprodan, line 57: Could we add `Dump` & `Load` functions so that the cache would be serialised to JSON on disk when the controller receives sigterm, and loaded into memory at startup?

souleb: do we have a use case for persisting to disk? For the ones listed, values are either already persisted or can be requested and cached again with an api call (those are cheap as we are talking about credentials).

stefanprodan: I was thinking about image scanning and tag caching. For this we are using badger in IRC but maybe a simpler solution would suffice.

souleb: This could be a good idea. What can be done here is optionally persist the cache in a file. The file would be periodically rotated to take into account the cache changes. The difference with badger is that we would have no LSM Tree implementation but a file that only grows. I think for IRC this is fine because the data is small, but this will have to be benchmarked. So It will have to be done in another PR.

stefanprodan: > a file that only grows

I don't think this is the case, the dump will contain whatever is in-memory when at controller shutdown. If keys have expired, the janitor would've removed them.

souleb: yes, with expiration and eviction. Those are optional, but you are right, we have control.

stefanprodan: Same for LRU, the file will never grow beyond the limit

souleb: for LRU, the capacity is in number of items. It doesn't limit the underlying file size. For expiration, if we set it to 0, its unlimited. We just need to be aware of the behavior anyway.

errordeveloper:
Thanks for working on this, @souleb!

> The store interface is a simplified version of client go store.

It would be handy to have a link in the code. At the moment this is only mentioned in a commit message and PR description.

souleb:
> Thanks for working on this, @souleb!
> 
> > The store interface is a simplified version of client go store.
> 
> It would be handy to have a link in the code. At the moment this is only mentioned in a commit message and PR description.
> 
> What may need to be made more explicit is that the sore is persisted to a file.

There is a comment in `store.go` pointing to https://pkg.go.dev/k8s.io/client-go/tools/cache#Store.

Persisting to a file will be made in a follow up PR where we will outline the use case we have for it.

cache/cache.go
```diff
@@ -0,0 +1,496 @@
...
+	items []*item[T]
+	// sorted indicates whether the items are sorted by expiration time.
+	// It is initially true, and set to false when the items are not sorted.
+	sorted bool
+	// capacity is the maximum number of index the cache can hold.
+	capacity   int
+	metrics    *cacheMetrics
+	labelsFunc GetLvsFunc[T]
+	janitor    *janitor[T]
+	closed     bool
```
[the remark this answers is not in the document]
souleb: We have an goroutine that handle eviction of expired items. When we close the cache with `Close()`, that goroutine stops. The sentinel value, is there for adding items in a cache where the cleaning goroutine is not running.

cache/store.go
```diff
@@ -0,0 +1,133 @@
...
+// ExplicitKey can be passed to IdentifiableObjectFunc if you have the key for
+// the objectect but not the objectect itself.
+type ExplicitKey string
+
+// IdentifiableObjectKeyFunc is a convenient default KeyFunc which knows how to make
+// keys from IdentifiableObject objects.
+func IdentifiableObjectKeyFunc[T any](object T) (string, error) {
+	if key, ok := any(object).(ExplicitKey); ok {
+		return string(key), nil
+	}
```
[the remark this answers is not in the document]
souleb: When you always use the `Get()` function  and never `GetByKey()` you might have some objects for which you only have a `key`. The `ExplicitKey` solve that.

cache/cache.go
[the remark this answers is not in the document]
souleb: There is another PR for persisting the cache. We will persist to disk when calling `Close()` if persistence is enabled.

cache/cache.go
```diff
@@ -0,0 +1,496 @@
...
+	opt := storeOptions[T]{}
+	for _, o := range opts {
+		err := o(&opt)
+		if err != nil {
+			return nil, err
+		}
+	}
+	if opt.registerer == nil {
+		return nil, ErrNoRegisterer
+	}
```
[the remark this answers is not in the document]
souleb: It's not optional at the moment. I'll have to refactor how we call the metrics funcs. I'll do that.

cache/cache.go
```diff
@@ -0,0 +1,496 @@
...
+	return nil
+}
+
+// Add an item to the cache, existing index will not be overwritten.
+// To overwrite existing index, use Update.
+// If the cache is full, Add will return an error.
+func (c *Cache[T]) Add(object T) error {
+	key, err := c.keyFunc(object)
+	if err != nil {
+		return KeyError{object, err}
```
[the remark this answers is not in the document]
souleb: if the keyFunc is wrong, many actions will result in error due to user provided parameter. This might have a consequence of too many metrics. Also this error is not on the cache side, so I wouldn't record it.

cache/cache.go
```diff
@@ -0,0 +1,496 @@
...
+		keys = append(keys, k)
+	}
+	c.mu.RUnlock()
+	c.metrics.incCacheRequests(StatusSuccess)
+	return keys
+}
+
+// Resize resizes the cache and returns the number of index removed.
+func (c *cache[T]) Resize(size int) int {
+	overflow := len(c.items) - size
```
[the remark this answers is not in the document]
souleb: I'll make size an `uint`

cache/cache.go
```diff
@@ -0,0 +1,496 @@
...
+
+	c.mu.Lock()
+	if c.closed {
+		c.mu.Unlock()
+		c.metrics.incCacheRequests(StatusFailure)
+		return KeyError{object, ErrClosed}
+	}
+	item, ok := c.index[key]
+	if ok {
+		item.expiration = time.Now().Add(expiration).UnixNano()
```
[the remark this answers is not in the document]
souleb: What is the issue with setting negative expiration? If won't panic. The consequences I see is for it to be expired, and so it will be garbage collected, so that's fine. There can be overflow, but we can't resonnably guard against that.

cache/cache.go
```diff
@@ -0,0 +1,496 @@
...
+	capacity   int
+	metrics    *cacheMetrics
+	labelsFunc GetLvsFunc[T]
+	janitor    *janitor[T]
+	closed     bool
+
+	mu sync.RWMutex
+}
+
+var _ Expirable[any] = &Cache[any]{}
```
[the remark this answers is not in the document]
souleb: `Store` is embedded in `Expirable` so it is fine.

cache/lru.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil {
+		c.metrics.incCacheRequests(StatusFailure)
+		return KeyError{object, err}
+	}
+
+	// if node is head or tail, do nothing
+	if key == c.head.key || key == c.tail.key {
+		c.metrics.incCacheRequests(StatusSuccess)
+		return nil
+	}
```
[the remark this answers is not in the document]
souleb: Your understanding is correct. I just wanted to safeguard for empty `string` keys. Doing it this way seems fine to me, if ever anyone changes `head` and `tail` keys to any other string.

cache/cache.go
```diff
@@ -0,0 +1,525 @@
...
+	item, ok := c.index[key]
+	if ok {
+		item.expiresAt = expiration
+		// mark the items as not sorted
+		c.sorted = false
+		recordRequest(c.metrics, StatusSuccess)
+	} else {
+		recordRequest(c.metrics, StatusFailure)
+		return KeyError{object, ErrNotFound}
+	}
```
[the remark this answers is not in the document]
souleb: Good catch! This came with the `time.Time` refactoring

errordeveloper:
Exciting to see this merged! 🌟

PR #783, opened 2024-06-19
feat: add un-taring plain, unzipped tar files

Skarlso:
Alternatively I could just NOT error when Gzip header fails. The choice is up to you dear maintainers.


stefanprodan (approved):
LGTM

Thanks @Skarlso

Issue #311, opened 2022-08-10
Support cross cloud federation


danquack:
GCP supports authenticating from different trusted identities. One possible authentication story is a user wants to auth from AWS against a GCloud Container Registry. Currently the [auth code](https://github.com/fluxcd/pkg/blob/main/oci/auth/gcp/auth.go) only supports querying the internal metadata url via the const GCP_TOKEN_URL. 

References:
https://cloud.google.com/iam/docs/configuring-workload-identity-federation
https://cloud.google.com/iam/docs/using-workload-identity-federation#aws_3

thejosephstevens:
I'm going to look into taking this on. I had assumed that the GCP provider operated on the standard GCP auth resolution, but as it works here you have to either pass a secret into the upstream resource (e.g. OCIRepository), or you have to be running in GCP and using a metadata endpoint for auth. I run in all 3 clouds, and wanted to distribute all artifacts from a centralized place in GCP, and I was really trying to avoid shipping service account json creds around everywhere. I don't want to use node-level permissions because that requires allowing a pod to access node permissions, which we lock down by default, and we use a custom CNI in GKE which doesn't play well with the GKE Metadata Server daemon, so that's also not an option.

matheuscscp:
Supporting this for object-level workload identity is easy: Among the three cloud providers only GCP needs changes in our code.

The `auth.Provider.GetAudience()` interface needs to receive a `serviceAccount` parameter. In the GCP provider we look up a custom Flux annotation in this service account, e.g. `gcp.auth.fluxcd.io/workloadIdentityProvider`, containing the full resource name of a GCP Workload Identity Provider. This is the required audience for issuing the Kubernetes service account token, and also the audience required for exchanging the Kubernetes service account token for a GCP access token.

matheuscscp:
Fixed here: https://github.com/fluxcd/pkg/pull/908

PR #815, opened 2024-10-13
Enable standard forms of GCP auth for oci sources

thejosephstevens:
This is related to #311 . We use k8s-based OIDC for all of our GCP auth, which allows for us to configure workloads in clusters in a simple, templated way to get access to cloud resources without ever having to exchange secret materials. This has some fairly straightforward security benefits, in addition to removing operational concerns of secret distribution, rotation, invalidation, etc...

The updated code piggybacks on the [`DefaultTokenSource`](https://pkg.go.dev/golang.org/x/oauth2/google#DefaultTokenSource) method from the google auth package in order to support all standard methods of obtaining an auth token, in addition to the previously supported method requiring the well-known metadata endpoint to be available (only applicable in GCE and GKE clusters with Workload Identity enabled).

Note - This technically constitutes a breaking change for auth behavior, but it should be very uncommon for it to negatively impact users as they would have had to set up Flux with some form of ambient auth, and leave it in place even though it's not taking effect and they're falling back to metadata endpoint auth. If, for example, someone has a flux workload with `GOOGLE_APPLICATION_CREDENTIALS` configured, with the existing code that would be ignored, but with this change in place their auth behavior would change. I'm happy to write up full details on what scenarios would be impacted by this for release notes, where would be best for me to put that?

Edit:
I just wanted to note that I was able to create a build of source-controller based off of this `pkg/oci` revision and it appears to work fine. I'm not seeing any issues introduced, and I was able to pull a GCP Artifact Registry OCI repo with k8s-based OIDC from an EKS cluster.


stefanprodan:
Hey @thejosephstevens can you signoff your commits so we can run the tests. See https://github.com/fluxcd/pkg/pull/815/checks?check_run_id=31453839284

thejosephstevens:
> Hey @thejosephstevens can you signoff your commits so we can run the tests. See https://github.com/fluxcd/pkg/pull/815/checks?check_run_id=31453839284

Done!

stefanprodan:
@thejosephstevens please run `make tidy` and force push.

stefanprodan:
@darkowlzz can you fork this and run our GCP e2e tests please.

stefanprodan:
@thejosephstevens could you please run `make test-oci` and fix the failing ones.

thejosephstevens:
> @thejosephstevens could you please run `make test-oci` and fix the failing ones.

Just got them all working!

matheuscscp:
Hi @thejosephstevens, thanks for all the work here! :medal_sports: 

One question:

> pull a GCP Artifact Registry OCI repo with k8s-based OIDC from an EKS cluster

Here, are you specifically referring to GCP Workload Identity *Federation*?

oci/auth/gcp/auth.go
```diff
@@ -64,69 +63,57 @@ func WithProxyURL(proxyURL *url.URL) Option {
 	}
 }
 
+// WithTokenSource sets a custom token source for the client.
+func (c *Client) WithTokenSource(ts oauth2.TokenSource) *Client {
```
matheuscscp, line 67: What's the use case for option?

thejosephstevens: This is primarily to enable passing in a stub for testing. In general use I'm not seeing a reason to use anything other than the default resolve for fetching the token source (that keeps the behavior aligned with conventional GCP client behavior that people will expect)

thejosephstevens:
> Hi @thejosephstevens, thanks for all the work here! 🏅
> 
> One question:
> 
> > pull a GCP Artifact Registry OCI repo with k8s-based OIDC from an EKS cluster
> 
> Here, are you specifically referring to GCP Workload Identity _Federation_?

I am! We generate cross-cloud identities for our workloads backed by k8s OIDC through Workload Identity Federation (or equivalent) in each cloud. We just happen to run our core operations primarily in GCP, which is why I jumped in here to contribute.

At a super specific level, we use `GOOGLE_APPLICATION_CREDENTIALS` to point to an external_account json file (mounted by configmap) which maps to the GCP identity, and attach a service account token with the appropriate audience for that GCP Workload Identity Pool Provider.

oci/auth/gcp/auth.go
matheuscscp: Since our tests use the same `package gcp` as the main code, we don't need to expose an exported API for specifying this mock/stub token source, you can directly assign a value to the unexported field `c.tokenSource` in the tests. I think I prefer it that way, so we don't introduce a new API without a concrete use case for it besides tests.

matheuscscp: I see that we have a very similar function `WithTokenURL()` here, so I assume you just followed the same style :+1: 

But since we're changing this, I think I prefer not introducing this API publicly.

@darkowlzz Thoughts?

thejosephstevens: > I see that we have a very similar function WithTokenURL() here, so I assume you just followed the same style 👍

Pretty much, yeah. I think the argument to not make that API public without a use case makes sense, the issue I'm running into is that if it's not on a public method I can't override the private field directly from tests that aren't in the same package (specifically `oci/auth/login/login_test`). There's a couple options that come to mind:
* I can create a distinct function in `oci/auth/gcp/auth_test` to return a client with a passed-in token source (functionally not very different, but it can have a super clear name like `NewClientWithStubTokenSource`)
* I can implement the token source override as an `Option` like `WithProxyUrl` (again, not terrifically different in functionality, but that maybe makes it less confusing)

I don't see another way around this without refactoring the `login` and `oci` packages, which seems like it would be a lot to add to this PR. If you've got another idea I'd be happy to give it a shot, my go is a tad rusty, so there may be something I'm not thinking of.

matheuscscp: > the issue I'm running into is that if it's not on a public method I can't override the private field directly from tests that aren't in the same package (specifically oci/auth/login/login_test).

Fair enough!

matheuscscp:
Hi @thejosephstevens!

So, yes, like @darkowlzz mentioned in the previous comment, this PR is currently removing support for proxy configuration when talking to the internal metadata server of GCE/GKE. The `proxyURL` option is no longer being passed anywhere. We need to keep proxy support at least for the code path that already exists, as it was introduced in the last Flux release for adding per-object proxy configuration for `OCIRepository`.

On one hand, I confess that I'd normally be very tempted to accept the current solution of the PR as is. I appreciate very much the simplicity of just calling a single function from the Google library. This has many benefits, including the ease of maintenance, and I believe it's also what most projects would do.

On the other hand, the function [`DefaultTokenSource`](https://pkg.go.dev/golang.org/x/oauth2/google#DefaultTokenSource) does not accept any non-global form of configuring a proxy URL for fetching the token, so if we just use it we lose per-object proxy configuration support.

I think a good compromise would be to copy, modify and maintain the function [`FindDefaultCredentialsWithParams`](https://pkg.go.dev/golang.org/x/oauth2/google#FindDefaultCredentialsWithParams), which is what `DefaultTokenSource` ultimately calls.

As the first line of the documentation says, the function `FindDefaultCredentialsWithParams` searches for Google's "Application Default Credentials", which are the methods used by the Google libraries for retrieving authentication from the environment of the process. Those methods include both the one we currently have, which is talking to the GCE/GKE internal metadata server, and the one which reads a JSON configuration file located either at `os.Getenv("GOOGLE_APPLICATION_CREDENTIALS")` or [`wellKnownFile()`](https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.23.0:google/default.go;l=303-309). The JSON configuration file method is the one that covers Workload Identity Federation, and also JSON Service Account Keys, along a few other methods.

The function `FindDefaultCredentialsWithParams` is simple and short enough for us to copy, modify and maintain. We would just need to replace [their implementation for talking to the internal metadata server](https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.23.0:google/default.go;l=216-236) with ours, which supports proxy configuration. This would unfortunately change the interface of this fuction to optionally accept the proxy URL, but that's the whole point of making this choice and not a big deal. I'd also be fine copying the other small, unexported functions called by this one, as they are all very simple and short as well, but I'd be fine with inlining them as well. To simplify even further, I think it would also be okay to change the return type from `*google.Credentials` to `oauth2.TokenSource`. The latter is the interface that we will need to implement using our current code (that fetches the token from the internal metadata server) in order to conform with the JSON configuration file code that we will leverage from the library, which will be handed to us also through this interface.

I'd like to note that some of the JSON configurations may still require talking to external HTTP endpoints. Full support for per-object proxy configuration would mean supporting those HTTP calls as well. Workload Identity Federation is probably one of those, as the OIDC ServiceAccount Token issued by Kubernetes must still be exchanged for a Google Access Token, which is done by talking to a Google HTTP API. It would be harder to support proxy for that, as it would require us to copy and adapt a lot more code from the Google library. But, as I mentioned previously, I'd be fine introducing support for new use cases without removing proxy support for the code path that we already have. And we probably need an issue to track proxy support for the remaining code paths.

thejosephstevens:
Hmmmm, that seems like a lot of code to copy and maintain that's currently just handled by Google. My initial reaction is it seems like if the requirement is to maintain per-object proxy support for requesting metadata-service tokens, it seems like it would be simplest to just split logic on the presence of a `proxyUrl` override (which could be served by the existing logic), and have all remaining calls fall through the Google-maintained `defaultTokenSource` path.

I may be missing some context here though, I know that Flux supports auth strategies defined at the object level, but I'm not actually aware of when that would be used. From my experience I would expect:
* For teams where Flux is operated by one group and consumed by others, I would expect objects to either pass in their own creds, or leverage the existing "ambient" access that Flux has (set up by SRE/Infra/whoever)
* For teams where Flux and the objects are operated by the same group I would expect similar patterns (but more preference for "ambient" auth)

I understand global proxy requirements (something-something network complexities), but in order to need per-object proxy support I _think_ that would require consumers to be effectively shoving their own identities into Flux for auth by way of proxies, but as it stands today the metadata URL is hardcoded in Flux, so I just don't know how that would even function to produce different auth resolutions. From the base issue [here](https://github.com/fluxcd/image-reflector-controller/issues/612) it looks like maybe this is focused on consistency across providers and the inherent per-object auth model that Flux uses, but I don't know if I'm missing use cases from the community. I'd be interested if there's context you could point me to (or raise) that I'm missing.

I don't have a strong opinion on implementation details, but I think if the consensus is to head down the route of copying chunks of the Google auth providers and working them in I may ask if one of you could pick up the remainder of the PR, as I just hadn't earmarked that much more time for this work.

matheuscscp:
Superseded by #908

PR #817, opened 2024-10-17
Simplify the cache

darkowlzz:
Based on the cache implementation improvement/simplification plan discussed
before, this change refactors it along the same lines with additional
improvements. See
https://gist.github.com/darkowlzz/34bc8b37d016b8aeb1de6660173652eb for the
initial improvement idea. It also updates oci/auth package which used the cache,
to remove the auth token caching. Auth token caching will be done at a higher
level by an OCI client.

At the time of writing the initial improvement plan, I was focused on the cache
interface and the exposed API. But while implementing the changes, I realized
how tightly the cache and their metrics were coupled, which may have contributed
to the verbose API. I'll explain some of it below for the record along with some
other explanations related to the changes.

#### dependency on kubernetes

The initial implementation was based on client-go's cache.Store interface and
was implemented around kubernetes style objects with object metadata. Due to
this, the implementation used kubernetes related go packages. Any user of the
cache, for example the git package for caching the git auth tokens, when
importing the cache, also imports all the kubernetes dependencies. This is
undesirable to keep simple packages independent of kubernetes dependencies. The
simplification changes the design around objects and makes it based on generic
value.

#### verbose interface

The initial Store interface, which the cache implementations implement, had too
many behaviors defined. It seemed to leak the implementation detail on the
interface. A cache interface needs ways to perform Get, Set and Delete. All the
other operations like resize, listing the keys, etc can be done per
implementation or build on top of the simple base interface if needed. It's
possible that we may have use case for some of the extra behaviors in the
future, but at present, we don't seem to have them. To simplify the interface,
the Store interface now has Get(), Set() and Delete() which work on string key
and generic type value.

#### complexity due to StoreObject

Because the initial implementation was based on kubernetes like objects,
any/most operations require building a StoreObject which is passed to the
cache for setting Get/Set/Delete. This made the cache usage code complex for
simple use. Most of the cache use cases in flux will store string or byte slice.
There's no known use case for caching kubernetes style object. Which makes the
need to create StoreObject for every use unnecessary. The new implementation
explicitly requires a string key and generic type value.

#### cache metrics

The initial cache was implemented to be self-instrumented. For all the
operations, it records metrics on its own. But cache event metrics which
recorded cache hit or miss were defined to be associated with object being
reconciled, similar to the metrics in the runtime package. It is desirable to
provide metrics for cache usage by flux object reconciliation. Since the initial
implementation required a StoreObject for every operation, it extracted the
associated reconciling object details from the StoreObject being passed. This
creates a strong coupling between the cache and the metrics, which is
reflected in the API. It's a nice property but at the cost of complexity on the
API. In addition, the implementation wasn't consistent in regards to the
associated object information on the metric. Some operations like GetByKey()
didn't consider the need for the reconciled object.
The new implementation decouple the cache and the metrics, making the metrics
that require the associated reconciling object to be explicitly called by the
caller who may have the information about the reconciling object.

The new implementation also introduced methods to delete the metrics when the
associated objects are deleted, like we do for any metric that's associated with
flux objects.

#### cache metrics prefix

The cache `Options` now support setting a metrics prefix to be added to all the
metrics registered by a cache. This is to prevent metrics name conflict when
using multiple instances of the cache in a program with the same metrics
registerer. Refer https://github.com/fluxcd/source-controller/issues/1644 for more details.

#### remove cache from OCI auth 

Initially, the cache was used in OCI auth package to cache the obtained auth
token. But later on, it was decided to not cache at the login manager level and
let the caller of such login functions to perform the cache as they have more
context about the scenario. https://github.com/fluxcd/pkg/pull/776 introduced
the cache in oci/auth package along with expiry time from the auth providers.
The cache is now removed and to return the expiry time to the caller, a new
function `LoginWithExpiry()` is introduced. A auth token caching client can call
`LoginWithExpiry()` to obtain the TTL of the obtained token and use it in the
cache. 

Following is a list of overall changes in brief:

- Remove dependency on kubernetes.
- Simplify the Store interface to only have Get, Set and Delete.
- Add new RecordCacheEvent() method in the cache implementations for recording the cache events, hit or miss, along with labels of the associated flux object being reconciled. The labels name, namespace and kind for this metric are not required to be configured when creating the cache. They are predefined in the created metrics recorder, similar to the runtime metrics recorder for the rest of flux. RecordCacheEvent() has to be called by the caller of cache operations explicitly along with the label information of the associated object being reconciled. The cache no longer has the knowledge of the object being reconciled, decoupled.
- With the decoupled cache and reconciling object for metrics, the storeOptions no longer need to have the generic type defined. Simplifying the usage.
- Add new DeleteCacheEvent() method in the cache implementations for deleting the cache events, hit or miss, which are associated with the object being reconciled. When the reconciling object is deleted, these metrics associated with the object must also be deleted.
- Updated all the tests to use simple string cache.
- Get() now returns a pointer instead of a separate `exists` boolean. If the pointer is nil, the item is not found in the cache.
- Get() takes a key and returns a cached item for it. GetByKey() is removed as Get() does the same now.
- In order to prevent the cache metrics to conflict when multiple caches are used against a common metrics registerer, the metrics constructor now accepts a prefix. This can be used to add a prefix to all the registered metrics of a cache.
- Revert the cache introduced in oci/auth in https://github.com/fluxcd/pkg/pull/776, keeping the expiration time which can be used by a cache at a higher level. Introduce LoginWithExpiry() to return the auth expiry time, keeping the existing Login() signature as it is. Auth token caching for OCI will be implemented separately in the OCI client.
- Add doc.go for the package to show its usage.

OCI integration test run with the above changes: https://github.com/fluxcd/pkg/actions/runs/11743637711

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -168,35 +168,34 @@ func (c *cache[T]) set(key string, value T) {
 	c.items = append(c.items, &item)
 }
 
-// Get returns a pointer to an item in the cache for the given key. If no item
-// is found, it's a nil pointer.
+// Get returns an item in the cache for the given key. If no item is found, an
+// error is returned.
 // The caller can record cache hit or miss based on the result with
 // Cache.RecordCacheEvent().
-func (c *Cache[T]) Get(key string) (*T, error) {
+func (c *Cache[T]) Get(key string) (T, error) {
+	var res T
 	c.mu.RLock()
 	if c.closed {
 		c.mu.RUnlock()
 		recordRequest(c.metrics, StatusFailure)
-		return nil, ErrCacheClosed
+		return res, ErrCacheClosed
 	}
 	item, found := c.index[key]
 	if !found {
 		c.mu.RUnlock()
 		recordRequest(c.metrics, StatusSuccess)
-		return nil, nil
+		return res, ErrNotFound
 	}
 	if !item.expiresAt.IsZero() {
 		if item.expiresAt.Compare(time.Now()) < 0 {
 			c.mu.RUnlock()
 			recordRequest(c.metrics, StatusSuccess)
-			return nil, nil
+			return res, ErrNotFound
 		}
 	}
 	c.mu.RUnlock()
 	recordRequest(c.metrics, StatusSuccess)
-	// Copy the value to prevent writes to the cached item.
-	r := item.value
-	return &r, nil
+	return item.value, nil
 }
 
 // Delete an item from the cache. Does nothing if the key is not in the cache.
--- a/cache/cache_test.go
+++ b/cache/cache_test.go
@@ -40,8 +40,8 @@ func TestCache(t *testing.T) {
 		key1 := "key1"
 		value1 := "val1"
 		got, err := cache.Get(key1)
-		g.Expect(err).ToNot(HaveOccurred())
-		g.Expect(got).To(BeNil())
+		g.Expect(err).To(Equal(ErrNotFound))
+		g.Expect(got).To(BeEmpty())
 
 		// Add an item to the cache
 		err = cache.Set(key1, value1)
@@ -50,13 +50,13 @@ func TestCache(t *testing.T) {
 		// Get the item from the cache
 		got, err = cache.Get(key1)
 		g.Expect(err).ToNot(HaveOccurred())
-		g.Expect(*got).To(Equal(value1))
+		g.Expect(got).To(Equal(value1))
 
 		// Writing to the obtained value doesn't update the cache.
-		*got = "val2"
+		got = "val2"
 		got2, err := cache.Get(key1)
 		g.Expect(err).ToNot(HaveOccurred())
-		g.Expect(*got2).To(Equal(value1))
+		g.Expect(got2).To(Equal(value1))
 
 		// Add another item to the cache
 		key2 := "key2"
@@ -68,7 +68,7 @@ func TestCache(t *testing.T) {
 		// Get the item from the cache
 		got, err = cache.Get(key2)
 		g.Expect(err).ToNot(HaveOccurred())
-		g.Expect(*got).To(Equal(value2))
+		g.Expect(got).To(Equal(value2))
 
 		// Update an item in the cache
 		key3 := "key3"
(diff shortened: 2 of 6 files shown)


stefanprodan (approved):
LGTM

Thanks @darkowlzz 🥇

cache/cache.go
```diff
@@ -181,86 +168,41 @@ func (c *cache[T]) set(key string, object T) {
...
 		recordRequest(c.metrics, StatusFailure)
-		return res, false, ErrCacheClosed
+		return nil, ErrCacheClosed
 	}
 	item, found := c.index[key]
 	if !found {
 		c.mu.RUnlock()
 		recordRequest(c.metrics, StatusSuccess)
-		return res, false, nil
+		return nil, nil
```
dipti-pai, line 186: In the previous implementation the bool return value (exists) indicated if the element is not present in the cache. With the current implementation, the caller cannot deduce a cache miss scenario since no error is returned when the cache element is not found. This leads to a segmentation fault in the caller while dereferencing the item. This can be fixed by returning ErrNotFound here if the item is not found in the cache. The caller can then check for ErrNotFound to figure out if this is a cache miss.

cache/cache.go
```diff
@@ -181,86 +168,41 @@ func (c *cache[T]) set(key string, object T) {
...
 	}
 	if !item.expiresAt.IsZero() {
 		if item.expiresAt.Compare(time.Now()) < 0 {
 			c.mu.RUnlock()
 			recordRequest(c.metrics, StatusSuccess)
-			return res, false, nil
+			return nil, nil
```
dipti-pai, line 192: Return an explicit error here since the caller cannot figure out if the item has expired and that value is not being returned.

dipti-pai:
Reviewed the changes and tested the following scenarios by using the cache to store credentials used for github app authentication -

1. Cache miss - Credentials are not present in the cache. This scenario is currently broken (left comments) and can be fixed by returning ErrNotFound from the cache. Rest of the tests were done after fixing this. 
2. Set credentials into the cache after successful authentication. 
3. Cache hit - Credentials are present in the cache and retrieved successfully.

cache/cache.go
dipti-pai: Ok, my vote is to return an error for not found and not return a pointer as it feels more intuitive to use and keeps the implementation in sync with the expiration methods.

dipti-pai (approved):
LGTM! Thanks.

stefanprodan:
@darkowlzz please rebase

PR #824, opened 2024-11-20
Update `ToUnstructured` to handle already unstructured case

errordeveloper:
This version of the function is more robus, it's what is used in Timoni, updating it hear would alow avoid copies in other places.


errordeveloper:
Eh, I suppose `ToUnstructured(object metav1.Object)` and `ToUnstructured(object runtime.Object)` are two quite distinct functions after all, despite looking so similar! Could be the case that `metav1.Object` cannot be unstructured actually?

errordeveloper:
@stefanprodan the intuition is to make PR to Timoni and drop a copy of this function from there.

stefanprodan (approved):
LGTM

Thanks @errordeveloper

PR #827, opened 2024-11-27
allow proxy during SSH host key scan in FluxCD CLI

adone:
`ssh.Dial` uses `net.DialTimeout` under the hood [src](https://cs.opensource.google/go/x/crypto/+/refs/tags/v0.29.0:ssh/client.go;l=176)
and there is no possibility to use a proxy when running command like `flux create source git --url ssh://git@example.com/repository.git` (unlike for the [source controller](https://fluxcd.io/flux/installation/configuration/proxy-setting/#git-repository-access-via-socks5-ssh-proxy) with `ALL_PROXY`)

so I replaced `ssh.Dial` with its almost full internal implementation
except `net.DialTimeout` replaced with `proxy.Dial`
like it is done in [go-git](https://github.com/go-git/go-git/blob/v5.0.0/plumbing/transport/ssh/common.go#L147)

### DIFF
--- a/ssh/go.mod
+++ b/ssh/go.mod
@@ -3,15 +3,16 @@ module github.com/fluxcd/pkg/ssh
 go 1.26.0
 
 require (
+	github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5
 	github.com/onsi/gomega v1.40.0
 	golang.org/x/crypto v0.47.0
+	golang.org/x/net v0.49.0
 )
 
 require (
 	github.com/google/go-cmp v0.7.0 // indirect
 	github.com/kr/text v0.2.0 // indirect
 	go.yaml.in/yaml/v3 v3.0.4 // indirect
-	golang.org/x/net v0.49.0 // indirect
 	golang.org/x/sys v0.40.0 // indirect
 	golang.org/x/text v0.33.0 // indirect
 )
--- a/ssh/go.mod
+++ b/ssh/go.mod
@@ -3,7 +3,7 @@ module github.com/fluxcd/pkg/ssh
 go 1.26.0
 
 require (
-	github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5
+	github.com/firefart/gosocks v0.4.2
 	github.com/onsi/gomega v1.40.0
 	golang.org/x/crypto v0.47.0
 	golang.org/x/net v0.49.0
--- a/ssh/go.sum
+++ b/ssh/go.sum
@@ -1,3 +1,5 @@
+github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5 h1:0CwZNZbxp69SHPdPJAN/hZIm0C4OItdklCFmMRWYpio=
+github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5/go.mod h1:wHh0iHkYZB8zMSxRWpUBQtwG5a7fFgvEO+odwuTv2gs=
 github.com/creack/pty v1.1.9/go.mod h1:oKZEueFk5CKHvIhNR5MUki03XCEU+Q6VDXinZuGJ33E=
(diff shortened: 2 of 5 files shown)


makkes (commented):
I'm generally in favor of this change. What I can't judge right now is which Flux CLI commands are affected by this. I couldn't find any controller code affected by this.

It will be very important to document this change with respect to the affected commands.

adone:
I've tried to improve `flux create source git`, but `flux bootstrap git` with SSH protocol is affected too

as I understand another PR should be made to document `ALL_PROXY` variable support in CLI (with `pkg` version update also)

P.S. and tests for go-git's clone are broken, I'm trying to understand why

adone:
ah! `ScanHostKey` ignores any SSH/network errors if it manages to get host keys

gecube:
I triaged fluxcd/flux2#4353 (full write-up [here](https://github.com/fluxcd/flux2/issues/4353#issuecomment-4990834085)) and it leads straight to this PR, so a few data points that may help revive it.

> What I can't judge right now is which Flux CLI commands are affected by this.

@makkes the enumeration (as of flux2 v2.9.2 — all of them funnel into `sourcesecret.GenerateGit` → `ssh.ScanHostKey`):

- `flux create source git` with an `ssh://` URL, unless `--secret-ref`/`--export` is given;
- `flux create secret git` with an `ssh://` URL — including with `--export` (generation happens before the export branch);
- `flux bootstrap github|gitlab|git|gitea|bitbucket-server` over SSH; `flux bootstrap git --private-key-file` also hits a second call site in [`bootstrap_git.go`](https://github.com/fluxcd/flux2/blob/v2.9.2/cmd/flux/bootstrap_git.go#L418) when building the CLI's own clone auth.

Related issue reports of the same root cause: fluxcd/flux2#4353, fluxcd/flux2#2509, fluxcd/flux2#3430 (closed), discussion fluxcd/flux2#4414.

Validation: the branch no longer builds against the current modules (its `ssh/go.mod` pins pre-2025 deps), so I applied the `host_key.go` change onto `ssh/v0.25.0` and built flux v2.9.2 with it. Behind a SOCKS5-only network path (sshd resolvable/reachable only through the proxy): scan with `ALL_PROXY=socks5://…` succeeds and pins the correct host key; without `ALL_PROXY` direct connections behave exactly as before. Since go-git already honors `ALL_PROXY` for the SSH transport, this makes `flux bootstrap git` over SSH work end-to-end behind a SOCKS5 proxy.

@adone are you still interested in finishing this? If not, I'm happy to submit a rebased version with a regression test, plus the flux2 docs PR documenting `ALL_PROXY` support for the affected commands.

adone:
hi @gecube 
my own original issue was resolved in a different way

honestly I would like to have a contribution to the project
so I will try to revive this PR on this weekend
if I fail you can take the opportunity

gecube:
Hi @adone,

Great, go for it — I'd much rather see you land it. Ping me if you need a review (at least I would be able to invest some efforts in testing stage)

One thing though: I'm still curious how your original issue got resolved. Even a short write-up would be useful for anyone else hitting the same thing.

adone:
I've added a test but it uses a little dirty trick to update cached ENV variables in `x/net/proxy` package
if you have any better ideas please point to right direction

ssh/go.mod
```diff
@@ -3,15 +3,16 @@ module github.com/fluxcd/pkg/ssh
 go 1.26.0
 
 require (
+	github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5
```
stefanprodan: We can not take on this dependency which is abandoned for 10+ years.

stefanprodan (changes requested):

ssh/go.mod
adone: i've replaced it with [github.com/firefart/gosocks](https://pkg.go.dev/github.com/firefart/gosocks)

PR #844, opened 2024-12-13
chore: add function to group entries by action in changeset output

94DanielBrown:
This can then be used to improve cardinality of logs in kustomize-controller.
I have tested and will make a PR to kustomize-controller which does this:  
```
log.Info("server-side apply for cluster definitions completed", "output", changeSet.ToMap())
+                       log.Info("server-side apply for cluster definitions completed", "output", changeSet.ToGroupedMap()
```

It helps with https://github.com/fluxcd/flux2/issues/4036

### DIFF
--- a/ssa/changeset.go
+++ b/ssa/changeset.go
@@ -76,6 +76,14 @@ func (c *ChangeSet) String() string {
 	return strings.TrimSuffix(b.String(), "\n")
 }
 
+func (c *ChangeSet) ToGroupedMap() map[Action][]string {
+	res := make(map[Action][]string)
+	for _, entry := range c.Entries {
+		res[entry.Action] = append(res[entry.Action], entry.Subject)
+	}
+	return res
+}
+
 func (c *ChangeSet) ToMap() map[string]Action {
 	res := make(map[string]Action, len(c.Entries))
 	for _, entry := range c.Entries {


stefanprodan:
We can't just change the log format of a GA controller, we'll break the filters people have for these logs just to comply with a small subset that are using elastic. If we decide to get this in, it needs to be under a feature flag in kustomize-controller.

94DanielBrown:
Yeah that makes sense 

But is there any harm getting the function added into here for now and then I can look at adding the feature flag in the controller and using it there after, if you agree that's an okay thing to do?

94DanielBrown:
Can it not just be added as a feature flag for just the kustomize controller as this is where the issue are? 

Also if it's added here as a function it doesn't have to be used but it makes it easier to use if someone want to run a fork of the controller to sort out the logging in ES/OS ?

stefanprodan:
> Can it not just be added as a feature flag for just the kustomize controller as this is where the issue are?

Yes, what I'm saying is that there may be other places where the log keys are high cardinality, and the feature flag should fix this everywhere.  I'll run some queries on my clusters and see if such logs are in other places besides here.

stefanprodan:
@94DanielBrown if you have time this week to help with this we can include this in Flux 2.5 release. Based on my log analysis, kustomize-controller apply and delete logs could be put behind a feature flag to switch the format. Please rebase this PR.

stefanprodan:
I've tagged this as `ssa/v0.45.1` so you can bump the ssa package in kustomize-controller and add a [feature flag](https://github.com/fluxcd/kustomize-controller/blob/main/internal/features/features.go) to switch the log format.

PR #848, opened 2024-12-23
[RFC-0008] Custom Event Metadata from Annotations

matheuscscp:
Part of: https://github.com/fluxcd/flux2/pull/4809


runtime/events/recorder.go
```diff
@@ -154,6 +156,16 @@ func (r *Recorder) AnnotatedEventf(
 		r.Log.Error(err, "failed to get object reference")
 	}
 
+	// Add object annotations to the annotations.
+	annotations := maps.Clone(inputAnnotations)
```
matheuscscp, line 160: This function was not modifying the metadata before so now that it does, in order to avoid a side-effect, it is safer to make a shallow copy of the map here (which in the case of a `map[string]string` is the same as a deep copy).

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #850, opened 2025-01-06
git: Add support for sparse checkout

kane8n:
Add the sparse checkout option to the gogit package.
This change is in preparation for adding the sparse checkout option to [ImageAutomationController](https://github.com/fluxcd/image-automation-controller).

Reference:
https://github.com/go-git/go-git/blob/master/_examples/sparse-checkout/main.go


stefanprodan:
> This change is in preparation for adding the sparse checkout option to [ImageAutomationController](https://github.com/fluxcd/image-automation-controller).

Hey @kane8n, sparse checkout could be very useful in source-controller as well.

PS. please run `make tidy` and force push so that CI can run.

stefanprodan:
Hey @pjbgf is sparse checkout stable in go-git, can it be used in IAC to commit changes back to the upstream branch? I'm reading the go-git code and I I assume this can't work, especially when IAC needs to push a new branch.

matheuscscp (approved):
Thanks for this contribution @kane8n!

git/gogit/clone_test.go
```diff
@@ -757,6 +759,308 @@ func Test_cloneSubmodule(t *testing.T) {
...
+		expectedTag               string
+		expectedErr               string
+		expectedEmpty             bool
+	}{
+		{
+			testType:                  "branch",
+			name:                      "Default branch",
+			branch:                    "master",
+			filesCreated:              map[string]string{"dir1/branch": "init1"},
+			sparseCheckoutDirectories: []string{"dir1"},
```
pjbgf, line 875: As the API enables multiple dirs to be set, we should add a test case that covers more than one dir

pjbgf (commented):
> Hey @pjbgf is sparse checkout stable in go-git, can it be used in IAC to commit changes back to the upstream branch? I'm reading the go-git code and I assume this can't work, especially when IAC needs to push a new branch.

@stefanprodan this would certainly be useful for SC. For IAC, it should also work however I haven't tested pushing from a repository that was sparse checkout using go-git. If that ends up being a problem, we have two options: fix the issue in `go-git` which we should be able to do at pace these days, or b) just opt-out of sparse checkout in IAC.

git/gogit/clone_test.go
kane8n: @pjbgf 
Thank you for your review!
I have added test cases and pushed the changes.

pjbgf (approved):
LGTM!

stefanprodan:
@kane8n did you tested that go-git can push commits from a repository that was sparse checkout?

stefanprodan:
@kane8n thanks, based on your findings we can make a plan for this how this feature will be implemented in the Flux controllers.

kane8n:
@stefanprodan @pjbgf 
I have tested go-git sparse-checkout!
Below are the results and evidence.

## Test Repository
I have prepared a Repository with files arranged as follows.
```bash
[go-git-test(main =)] $ tree
.
├── README.md
├── dir1
│   └── test.txt
├── dir2
│   └── test.txt
├── dir3
│   └── test.txt
└── root.txt

4 directories, 5 files
[go-git-test(main =)] $ 
```

## Summary
The pattern of editing a file in a sparse-checkouted branch and doing a commit -> push worked.
However, when I run switch branch from the sparse-checkouted branch, I get the error `worktree contains unstaged changes`.
If I check the status in worktree.Status(), I see that the unchecked out files are marked as deleted.
When I run the `git status` command on the sparse-checkouted directory in go-git, I get `working tree clean`, so I'm not sure why go-git's worktree.Status is not able to account for sparse-checkouts. I think the cause is that go-git's worktree.Status does not take sparse-checkout into account.

## Evidense
### The pattern of editing a file in a sparse-checkouted branch and doing a commit -> push
#### test code
https://go.dev/play/p/8r9ZMFhxud4

#### execution log
```bash
 $ go run main.go         
dir1
└── test.txt
 $ 
 $ cat /tmp/go-git-test/dir1/test.txt 
test write at 2025-01-22 19:52:47.731592 +0900 JST m=+1.974122043%                                                                                                                                                                                                                                                                                      
 $ 
```

#### Files on remote side after execution
![CleanShot 2025-01-22 at 20 26 43](https://github.com/user-attachments/assets/82b0bcf9-07ca-47dd-b2d6-8a515ce7164f)

### Pattern of running switch branch from sparse-checkouted branch
#### test code
https://go.dev/play/p/1Cd4nSqS97v

#### execution log
```bash
 $ go run main.go                    
dir1
└── test.txt
panic: worktree contains unstaged changes

goroutine 1 [running]:
main.main()
        /Users/takumi.kaneda/asobiba/main.go:64 +0x3bc
exit status 2
 $ 
```


#### status check test code
https://go.dev/play/p/9-o8v2HPujF

#### execution log
```bash
 $ go run main.go         
dir1
└── test.txt
dirty
README.md D
dir2/test.txt D
dir3/test.txt D
root.txt D
dir1/test.txt D
 $ 
```

#### Result of git status command execution
```
 $ tree                 
.
└── dir1
    └── test.txt

2 directories, 1 file
 $ 
 $ git status
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
 $
```

pjbgf:
@kane8n great stuff, thanks for testing this. Would you mind opening [an issue on go-git](https://github.com/go-git/go-git/issues/new?template=Blank+issue) to provide support to push sparse checked-out repositories when new branches are created?

In my opinion, this could be implemented as-is. IAC would have a slight caveat, which would be to disable sparse checkout automatically when a push branch is defined. This is similar to some of previous changes we did around Azure support a few years back.

A new issue on IAC would be needed to track the removal of that caveat, which would be blocked by the go-git issue.

@stefanprodan any concerns on the approach above?

stefanprodan:
> IAC would have a slight caveat, which would be to disable sparse checkout automatically when a push branch is defined.

I think it's fine, but then most users will not be able to benefit from sparse checkout. The majority of users have the main branch protected (for good security reasons) and usually Flux pushes to a new branch from which a PR is opened.

stefanprodan:
@kane8n please rebase with upstream main and force push

stefanprodan (approved):
LGTM

Thanks @kane8n 🏅 

PS. I have created an issue in source-controller for the next implementation step

kane8n:
@stefanprodan 
>rebase with upstream main and force push

done!

PR #859, opened 2025-01-27
[RFC-0009] Add InvalidCELExpressionReason to apis/meta

matheuscscp:
Part of: https://github.com/fluxcd/flux2/pull/5151
Part of: https://github.com/fluxcd/notification-controller/pull/948


stefanprodan (approved):
LGTM

PR #865, opened 2025-02-03
runtime: downgrade github.com/google/cel-go and cel.dev/expr

aerfio:
Downgrade these deps to versions used in k8s.io/apiserver:
- https://github.com/kubernetes/apiserver/blob/v0.32.1/go.mod#L21
- https://github.com/kubernetes/apiserver/blob/v0.32.1/go.mod#L70

Reason: this change prevents compilation error when the project's dependencies (or the project itself) imports both `k8s.io/apiserver/pkg/cel/environment` and `github.com/fluxcd/pkg/runtime/*`. 

Reproduction:
```go
package main

import (
	"fmt"

	"github.com/fluxcd/pkg/runtime/client"
	"k8s.io/apiserver/pkg/cel/environment"
)

func main() {
	_ = environment.StoredExpressions
	_ = client.GetConfigOrDie
	fmt.Println("exit")
}
```

`go.mod`:
```go
module github.com/aerfio/repoduction

go 1.23.5

require (
	github.com/fluxcd/pkg/runtime v0.53.0
	k8s.io/apiserver v0.32.1
)


require (
	cel.dev/expr v0.19.1 // indirect
	github.com/google/cel-go v0.23.1 // indirect
        // other indirect deps
)
```

compiling this simple app results in:
```
# k8s.io/apiserver/pkg/cel/environment
/Users/aerfio/go/pkg/mod/k8s.io/apiserver@v0.32.1/pkg/cel/environment/base.go:176:19: cannot use ext.TwoVarComprehensions (value of type func(options ...ext.TwoVarComprehensionsOption) "github.com/google/cel-go/cel".EnvOption) as func() "github.com/google/cel-go/cel".EnvOption value in argument to UnversionedLib
```

Trying to force correct deps also fails:
```
❯ go get -u github.com/google/cel-go@v0.22.0 cel.dev/expr@v0.18.0 github.com/fluxcd/pkg/runtime@v0.53.0
go: github.com/fluxcd/pkg/runtime@v0.53.0 requires github.com/google/cel-go@v0.23.1, not github.com/google/cel-go@v0.22.0
```

### DIFF
--- a/runtime/go.mod
+++ b/runtime/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/fluxcd/pkg/apis/kustomize v1.9.0
 	github.com/fluxcd/pkg/apis/meta v1.10.0
 	github.com/go-logr/logr v1.4.2
-	github.com/google/cel-go v0.23.1
+	github.com/google/cel-go v0.22.0
 	github.com/google/go-cmp v0.6.0
 	github.com/hashicorp/go-retryablehttp v0.7.7
 	github.com/kylelemons/godebug v1.1.0
@@ -41,7 +41,7 @@ require (
 replace gopkg.in/yaml.v3 => gopkg.in/yaml.v3 v3.0.1
 
 require (
-	cel.dev/expr v0.19.1 // indirect
+	cel.dev/expr v0.18.0 // indirect
 	github.com/Azure/go-ansiterm v0.0.0-20230124172434-306776ec8161 // indirect
 	github.com/MakeNowJust/heredoc v1.0.0 // indirect
 	github.com/antlr4-go/antlr/v4 v4.13.0 // indirect
--- a/runtime/go.sum
+++ b/runtime/go.sum
@@ -1,5 +1,5 @@
-cel.dev/expr v0.19.1 h1:NciYrtDRIR0lNCnH1LFJegdjspNx9fI59O7TWcua/W4=
-cel.dev/expr v0.19.1/go.mod h1:MrpN08Q+lEBs+bGYdLxxHkZoUSsCp0nSKTs0nTymJgw=
+cel.dev/expr v0.18.0 h1:CJ6drgk+Hf96lkLikr4rFf19WrU0BOWEihyZnI2TAzo=
+cel.dev/expr v0.18.0/go.mod h1:MrpN08Q+lEBs+bGYdLxxHkZoUSsCp0nSKTs0nTymJgw=
 github.com/AdaLogics/go-fuzz-headers v0.0.0-20240806141605-e8a1dd7889d6 h1:He8afgbRMd7mFxO99hRNu+6tazq8nFF9lIwo9JFroBk=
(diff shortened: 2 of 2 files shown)


matheuscscp:
@aerfio Flux doesn't seem to be using `k8s.io/apiserver/pkg/cel/environment` anywhere:

https://github.com/search?q=org%3Afluxcd%20k8s.io%2Fapiserver%2Fpkg%2Fcel%2Fenvironment&type=code

This seems to be a bug/regression in upstream CEL for users in your specific situation. Did you raise an issue with upstream CEL about this?

aerfio:
@matheuscscp No, it's not about CEL libs. The `fluxcd/pkg/runtime` depends on cel libs (`cel-go` and `cel.dev/expr`) and on several k8s.io libs @ v0.32.1. If I have the project that uses `fluxcd/pkg/runtime` and I'd like to to also use the `k8s.io/apiserver/pkg/cel/environment` (or have any other dependency/import that uses that package), then I get a compilation error. This PR downgrades the CEL libs in fluxcd/pkg/runtime to the same version used in the k8s.io/apiserver@v0.32.1, which is the same version used for other k8s deps.

IIUC this issue is primarily caused by the fact that cel-go has a breaking change in it's v0.23.0 minor version, but they're still v0 so they're allowed to do this. k8s.io/apiserver still uses _older_ version of this lib, `github.com/fluxcd/runtime` uses newest version, so they clash. But you do not use any feature from newest version of these CEL libs.

It's a choice of compatibility with k8s.io/apiserver or keeping newest `cel-go` and `cel.dev/expr` libs. controller-runtime recently introduced a tool that keeps its deps synced with k/k: https://github.com/kubernetes-sigs/controller-runtime/pull/2774

matheuscscp:
> But you do not use any feature from newest version of these CEL libs.

Actually we do, we bumped to `v0.23.0` specifically to support the functionality landed in CEL upstream by @bigkevmcd in this PR:

https://github.com/google/cel-go/pull/1067

Kevin was initially introducing this functionality as a custom CEL library in this PR:

https://github.com/fluxcd/notification-controller/pull/948

And using it like this:

https://github.com/fluxcd/notification-controller/pull/948/files#diff-59ed780af7ec7af60ec655b3c2252349c39224916bfb104ee8fc2773fec0773dR96-R113

Now bumping to `v0.23.0` we can support this use case without the custom CEL library.

What's your use case for including both `fluxcd/pkg/runtime` and `k8s.io/apiserver/pkg/cel/environment`?

aerfio:
> What's your use case for including both fluxcd/pkg/runtime and k8s.io/apiserver/pkg/cel/environment?
I'm using `fluxcd/pkg/*` libs and `k8s.io/apiserver` through the whole codebase for various things

> fluxcd/pkg/runtime is v0 too
Yup, I get that, and all the implications that stem from it.

Ok, given that you already use newest cel-go somewhere else I don't think this PR can be merged, I'll try to find another way or wait for k8s.io/apiserver to catch up. Thank you for your attention @matheuscscp

PR #870, opened 2025-02-22
Add cache specialized for access tokens

matheuscscp:
Introduce a cache that is specialized for access tokens. This is a thin abstraction on top of the LRU cache adding requirements that are important for managing the lifetime of tokens:

* We consider tokens expired after 80% of their lifetime. This is what kubelet does for rotating ServiceAccount tokens inside pods ([ref](https://github.com/kubernetes/kubernetes/blob/4032177faf21ae2f99a2012634167def2376b370/pkg/kubelet/token/token_manager.go#L172-L174)).
* We need a pessimistic approach for considering tokens expired. With this in mind we store two time points: the usual `time.Now().Add(duration)`, which has a monotonic clock, and also a Unix time point wihtout a monotonic clock. If either report `t.Before(time.Now()) == true`, the token is considered expired. This is important because Flux may be running in local machines, which are often put to sleep. When they come back up, the monotonic clock will resume from where it stopped, i.e. for the monotonic clock the machine sleep time is considered a "pause". This would cause tokens to be kept beyond their actual expiration, so we use both clocks to ensure we don't return expired tokens on `TokenCache.Get()`.

### DIFF
--- a/cache/token_test.go
+++ b/cache/token_test.go
@@ -1,0 +1,66 @@
+/*
+Copyright 2025 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cache_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	. "github.com/onsi/gomega"
+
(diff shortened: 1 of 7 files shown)


cache/token_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"time"
+
+	. "github.com/onsi/gomega"
+
+	"github.com/fluxcd/pkg/cache"
+)
+
+type testToken struct {
+	duration time.Duration
+}
```
stefanprodan, line 30: Can you please make this test case closer to the actual implementation in let's say source-controller & GitHub App.

I expect for this type to hold a GitHub token and also the kind, name, namespace. Then a function that composes the cache key.

An implementation here would help reviewing and also help consumers of this package.

matheuscscp: In order not to duplicate all this stuff here and in the actual implementation I created the following combo PR based on this one: https://github.com/fluxcd/pkg/pull/871

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #871, opened 2025-02-23
Add caching for GitHub App tokens

matheuscscp:
Builds on #870


auth/github/client.go
```diff
@@ -158,16 +166,41 @@ func WithProxyURL(proxyURL *url.URL) OptFunc {
...
+	return time.Until(at.ExpiresAt)
+}
+
 // GetToken returns the token that can be used to authenticate
 // as a GitHub App installation.
 // Ref: https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation
 func (p *Client) GetToken(ctx context.Context) (*AppToken, error) {
+	cacheKey := p.buildCacheKey()
+	if p.cache != nil {
+		if token := p.cache.Get(cacheKey); token != nil {
```
stefanprodan, line 197: Let's assume the GH App key is updated in the secret and the old one is invalidated. This will make the cached token to be rejected by GitHub and since we always read it from cache, source-controller will error out until the cache TTL runs out. We really need to delete the token from the cache if calling the GitHub API fails, so we have to expose some function in the client for e.g. `DeleteTokenFromCache`. In source-controller we'll have to purge the cache and retry once, then bail out.

matheuscscp: Check the logic of `p.buildCacheKey()`, `sha256(privateKey)` is included in the cache key. If the key is updated in the secret, the cache key will change. And there's no point in retrying an API call that fails due to authentication if the secret is not updated with a valid key, so this solves the problem with a much simpler solution

stefanprodan: Ok I guess the App ID can't change without changing the private key so this should solve it. Thanks!

stefanprodan (approved):
LGTM

PR #875, opened 2025-02-25
Use impersonated rest mapper for status poller

matheuscscp:


runtime/cel/status_reader_test.go
```diff
@@ -34,47 +34,6 @@ import (
 	"github.com/fluxcd/pkg/runtime/cel"
 )
 
-func TestNewStatusReader(t *testing.T) {
```
matheuscscp, line 37: This test was just testing the error in constructing the evaluator internally, now it is injected, so the test is no longer needed.

stefanprodan:
I propose we refactor the client into:

```
NewImpersonator(kubeClient, opts ...ImpersonatorOption)
WithScheme(scheme)
WithServiceAccount(defaultName, name, namespace)
WithKubeConfig(reference, options)
WithPolling(options, readers)
```

stefanprodan (approved):
LGTM

PR #880, opened 2025-03-05
Add option for custom event namespace label in cache metrics

matheuscscp:


stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #882, opened 2025-03-11
support string evaluation in CEL expressions

kathleenfrench:
related: https://github.com/fluxcd/notification-controller/issues/589#issuecomment-2688342204

in order to leverage the `cel` flux package for the purposes of evaluating a commit status/id expression, it's necessary to expose a function for evaluating strings ([similar](https://github.com/fluxcd/pkg/blob/8b1f852228b117123efcc7b5708ce112801416b6/runtime/cel/expression.go#L128) to `EvaluateBoolean`)

### DIFF
--- a/runtime/cel/expression.go
+++ b/runtime/cel/expression.go
@@ -136,3 +136,16 @@ func (e *Expression) EvaluateBoolean(ctx context.Context, data map[string]any) (
 	}
 	return bool(result), nil
 }
+
+// EvaluateString evaluates the expression with the given data and returns the result as a string.
+func (e *Expression) EvaluateString(ctx context.Context, data map[string]any) (string, error) {
+	val, _, err := e.prog.ContextEval(ctx, data)
+	if err != nil {
+		return "", fmt.Errorf("failed to evaluate the CEL expression '%s': %w", e.expr, err)
+	}
+	result, ok := val.(types.String)
+	if !ok {
+		return "", fmt.Errorf("failed to evaluate CEL expression as string: '%s'", e.expr)
+	}
+	return string(result), nil
+}
--- a/runtime/cel/expression_test.go
+++ b/runtime/cel/expression_test.go
@@ -219,3 +219,143 @@ func TestExpression_EvaluateBoolean(t *testing.T) {
 		})
 	}
 }
+
+func TestExpression_EvaluateString(t *testing.T) {
+	for _, tt := range []struct {
+		name   string
+		expr   string
+		opts   []cel.Option
+		data   map[string]any
+		result string
+		err    string
+	}{
+		{
+			name: "non-existent field",
+			expr: "foo",
+			data: map[string]any{},
+			err:  "failed to evaluate the CEL expression 'foo': no such attribute(s): foo",
+		},
+		{
+			name:   "string field",
+			expr:   "foo",
+			data:   map[string]any{"foo": "some-value"},
+			result: "some-value",
+		},
+		{
+			name: "non-string field",
+			expr: "foo",
+			data: map[string]any{"foo": 123},
+			err:  "failed to evaluate CEL expression as string: 'foo'",
+		},
+		{
+			name:   "nested string field",
+			expr:   "foo.bar",
+			data:   map[string]any{"foo": map[string]any{"bar": "some-value"}},
+			result: "some-value",
+		},
+		{
+			name:   "compiled expression returning string",
+			expr:   "foo.bar",
+			opts:   []cel.Option{cel.WithCompile(), cel.WithStructVariables("foo")},
+			data:   map[string]any{"foo": map[string]any{"bar": "some-value"}},
+			result: "some-value",
+		},
+		{
+			name: "compiled expression returning string multiple variables",
+			expr: "foo.bar + '/' + foo.baz + '/' + bar.biz",
+			opts: []cel.Option{
+				cel.WithCompile(),
+				cel.WithStructVariables("foo", "bar"),
+			},
+			data: map[string]any{
+				"foo": map[string]any{
+					"bar": "some-value",
+					"baz": "some-other-value"},
+				"bar": map[string]any{
+					"biz": "some-third-value",
+				},
+			},
+			result: "some-value/some-other-value/some-third-value",
+		},
+		{
+			name: "compiled expression with string manipulation and zero index",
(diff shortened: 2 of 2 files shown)


runtime/cel/expression_test.go
```diff
@@ -219,3 +219,79 @@ func TestExpression_EvaluateBoolean(t *testing.T) {
...
+			expr:   "foo.bar",
+			opts:   []cel.Option{cel.WithCompile(), cel.WithStructVariables("foo")},
+			data:   map[string]any{"foo": map[string]any{"bar": "some-value"}},
+			result: "some-value",
+		},
+		{
+			name:   "compiled expression returning string multiple variables",
+			expr:   "foo.bar + '/' + foo.baz + '/' + bar.biz",
+			opts:   []cel.Option{cel.WithCompile(), cel.WithStructVariables("foo", "bar")},
+			data:   map[string]any{"foo": map[string]any{"bar": "some-value", "baz": "some-other-value"}, "bar": map[string]any{"biz": "some-third-value"}},
```
matheuscscp, line 267: At this point I think we can start breaking some lines :sweat_smile:

runtime/cel/expression_test.go
```diff
@@ -219,3 +219,79 @@ func TestExpression_EvaluateBoolean(t *testing.T) {
...
+			result: "some-value/some-other-value/some-third-value",
+		},
+		{
+			name:   "compiled expression with string manipulation",
+			expr:   "foo.bar + '/' + foo.baz + '/' + bar.uid.split('-')[0].lowerAscii()",
+			opts:   []cel.Option{cel.WithCompile(), cel.WithStructVariables("foo", "bar")},
+			data:   map[string]any{"foo": map[string]any{"bar": "some-value", "baz": "some-other-value"}, "bar": map[string]any{"uid": "AKS2J23-DAFLSDD-123J5LS"}},
```
matheuscscp, line 274: Ditto

runtime/cel/expression_test.go
```diff
@@ -219,3 +219,79 @@ func TestExpression_EvaluateBoolean(t *testing.T) {
...
+		{
+			name:   "compiled expression returning string multiple variables",
+			expr:   "foo.bar + '/' + foo.baz + '/' + bar.biz",
+			opts:   []cel.Option{cel.WithCompile(), cel.WithStructVariables("foo", "bar")},
+			data:   map[string]any{"foo": map[string]any{"bar": "some-value", "baz": "some-other-value"}, "bar": map[string]any{"biz": "some-third-value"}},
+			result: "some-value/some-other-value/some-third-value",
+		},
+		{
+			name:   "compiled expression with string manipulation",
+			expr:   "foo.bar + '/' + foo.baz + '/' + bar.uid.split('-')[0].lowerAscii()",
```
matheuscscp, line 272: Here it would nice to test `split('-').first()`, a Flux contributor landed `.first()` and `.last()` in the upstream CEL lib a while a go, can you pls try that? :pray:

matheuscscp (commented):
@kathleenfrench Really glad to see a contribution in this lib I created recently, thanks very much! :grin:

runtime/cel/expression_test.go
kathleenfrench: updated 👍🏻  -- it needs a `.value()` after `first()` / `last()` otherwise the expression fails, so there is also a test for that failure case

matheuscscp: @bigkevmcd Is this `.value()` the thing you said about them asking for `.first()`/`.last()` be optional?

runtime/cel/expression.go
```diff
@@ -136,3 +136,16 @@ func (e *Expression) EvaluateBoolean(ctx context.Context, data map[string]any) (
 	}
 	return bool(result), nil
 }
+
+// EvaluateString evaluates the expression with the given data and returns the result as a string.
+func (e *Expression) EvaluateString(ctx context.Context, data map[string]any) (string, error) {
+	val, _, err := e.prog.ContextEval(ctx, data)
+	if err != nil {
+		return "", fmt.Errorf("failed to evaluate the CEL expression '%s': %w", e.expr, err)
```
bigkevmcd, line 144: This could be `%q` instead of `'%s'` ?

> %q	a double-quoted string safely escaped with Go syntax

runtime/cel/expression.go
```diff
@@ -136,3 +136,16 @@ func (e *Expression) EvaluateBoolean(ctx context.Context, data map[string]any) (
...
+	}
+	result, ok := val.(types.String)
+	if !ok {
+		return "", fmt.Errorf("failed to evaluate CEL expression as string: '%s'", e.expr)
+	}
```
bigkevmcd, line 149: Same as above.

runtime/cel/expression_test.go
bigkevmcd: Yes, upstream the `.first()` and `.last()` functions work on CEL optionals...because there could be no elements in the array.

bigkevmcd (approved):
Looks good to me.

runtime/cel/expression.go
kathleenfrench: addressed in https://github.com/fluxcd/pkg/pull/882/commits/df922dece8b1bd710d976d6ea6672f314fc56f5e

runtime/cel/expression.go
kathleenfrench: addressed in https://github.com/fluxcd/pkg/pull/882/commits/df922dece8b1bd710d976d6ea6672f314fc56f5e

runtime/cel/expression.go
matheuscscp: I actually prefer `'%s'`, when these strings show up in JSON for whatever reason the double-quotes are escaped, I really prefer the single-quote alternative, and this was copied from `EvaluateBoolean`, so we're now introducing the two versions in the repo, I prefer this change to be reverted

kathleenfrench: k, no problem -- i'll revert

matheuscscp:
@kathleenfrench Please revert the quote change and squash into a single commit :pray:

matheuscscp:
@kathleenfrench I pushed the tag `runtime/v0.55.0` so you can bump `fluxcd/pkg/runtime` in notification-controller when implementing the commit status ID feature :+1:

PR #885, opened 2025-03-12
Take the hash of all the parts in the GitHub App token cache key

matheuscscp:


auth/github/client.go
```diff
@@ -230,12 +230,13 @@ func (p *Client) GetToken(ctx context.Context) (*AppToken, error) {
 }
 
 func (p *Client) buildCacheKey() string {
-	privateKeyDigest := sha256.Sum256(p.privateKey)
 	keyParts := []string{
 		fmt.Sprintf("%s=%s", AppIDKey, p.appID),
 		fmt.Sprintf("%s=%s", AppInstallationIDKey, p.installationID),
 		fmt.Sprintf("%s=%s", AppBaseUrlKey, p.apiURL),
-		fmt.Sprintf("%sDigest=%x", AppPrivateKey, privateKeyDigest),
+		fmt.Sprintf("%s=%s", AppPrivateKey, string(p.privateKey)),
```
matheuscscp, line 237: This `string()` cast is fine, the private key is supposed to be in the PEM format, which is string

stefanprodan (approved):

PR #904, opened 2025-04-13
Support `ClientCert` and `ClientKey` (mTLS) configuration in Git

hiddeco:
Based on the current HEAD of go-git's `releases/v5.x` branch, pending the official release of `v5.16.0` with https://github.com/go-git/go-git/pull/1510.

Related to: https://github.com/fluxcd/source-controller/issues/1761


hiddeco:
Tests are failing due to a non-related change upstream in go-git. Figuring out what the issue is, but it looks like the fallback to the system `known_hosts` file broke when a feature got added.

stefanprodan:
I guess the fallback to the system known_hosts not working would break the Flux CLI bootstrap but the controllers would be unaffected.

hiddeco:
Paulo will have another look tomorrow and see if my gut feeling is right. If so, it will likely justify a patch on the go-git side instead of us having to jump through hoops. Will be continued...

stefanprodan:
I fixed the SOPS action in https://github.com/fluxcd/pkg/pull/905 please approve it.

git/options.go
```diff
@@ -38,6 +38,18 @@ const (
...
+	dataKeyIdentity    = "identity"
+	dataKeyKnownHosts  = "known_hosts"
+	dataKeyUsername    = "username"
+	dataKeyPassword    = "password"
+	dataKeyBearerToken = "bearerToken"
+	dataKeyTLSKey      = "tls.key"
+	dataKeyTLSCert     = "tls.crt"
+	dataKeyCAFile      = "caFile"
+	dataKeyCACert      = "ca.crt"
+)
```
stefanprodan, line 51: Thanks for refactoring these!

stefanprodan (approved):
LGTM

Thanks @hiddeco 🥇

PR #906, opened 2025-04-14
[RFC-0010] Add core auth library

matheuscscp:
This PR introduces the core library from [RFC-0010](https://github.com/fluxcd/flux2/blob/main/rfcs/0010-multi-tenant-workload-identity/README.md).

PR Stack:

* [`azure`](https://github.com/fluxcd/pkg/pull/909)
* [`gcp`](https://github.com/fluxcd/pkg/pull/908)
* [`aws`](https://github.com/fluxcd/pkg/pull/907)
* `core` <---


auth/options.go
```diff
@@ -0,0 +1,97 @@
...
+
+// Options contains options for configuring the behavior of the provider methods.
+// Not all providers/methods support all options.
+type Options struct {
+	Client          client.Client
+	Cache           *cache.TokenCache
+	ServiceAccount  *client.ObjectKey
+	InvolvedObject  cache.InvolvedObject
+	Scopes          []string
+	ImageRepository string
```
stefanprodan, line 38: I suggest using `Artifact` instead of `Image` throughout the APIs:

- `ArtifactRepository`
- `WithArtifactRepository`
- `GeArtifactCacheKey`
- `ArtifactRegistry` 
- `NewArtifactRegistryToken`
- `ArtifactRegistryCredentials`

matheuscscp: Fixed, thanks!

auth/provider.go
```diff
@@ -0,0 +1,65 @@
...
+	NewTokenForServiceAccount(ctx context.Context, oidcToken string,
+		serviceAccount corev1.ServiceAccount, opts ...Option) (Token, error)
+
+	// GetArtifactCacheKey extracts the part of the artifact repository that must be
+	// included in cache keys when caching registry credentials for the provider.
+	GetArtifactCacheKey(artifactRepository string) string
+
+	// NewRegistryToken takes an artifact repository and an access token and returns a token
+	// that can be used to authenticate with the container registry of the artifact.
+	NewRegistryToken(ctx context.Context, artifactRepository string,
```
stefanprodan, line 63:
```suggestion
	NewArtifactRegistryToken(ctx context.Context, artifactRepository string,
```

auth/options.go
matheuscscp: Oops, missed a few, fixing

auth/token.go
```diff
@@ -0,0 +1,47 @@
...
+	// relative to approximately time.Now(). This is used to determine when the token should
+	// be refreshed.
+	GetDuration() time.Duration
+}
+
+// RegistryCredentials is a particular type implementing the Token interface
+// for credentials that can be used to authenticate with a container registry
+// from a cloud provider. This type is compatible with all the cloud providers
+// and should be returned when the artifact repository is configured in the options.
+type RegistryCredentials struct {
```
stefanprodan, line 39:
```suggestion
type ArtifactRegistryCredentials struct {
```

auth/provider.go
matheuscscp: Fixed, thanks!

auth/token.go
matheuscscp: Fixed, thanks!

auth/get_token.go
```diff
@@ -0,0 +1,178 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to get provider audience: %w", err)
+		}
+
+		// Get provider identity.
+		providerIdentity, err = provider.GetIdentity(serviceAccount)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get provider identity from service account '%s/%s' annotations: %w",
+				serviceAccount.Namespace, serviceAccount.Name, err)
+		}
```
dipti-pai, line 70: providerIdentity is used only if caching is used and can be retrieved just before building the cache key.

auth/options.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// Apply applies the given slice of Option(s) to the Options struct.
+func (o *Options) Apply(opts ...Option) {
+	for _, opt := range opts {
+		opt(o)
+	}
+}
+
+// GetHTTPClient returns a *http.Client with the configured proxy URL
```
dipti-pai, line 100: Is this function called somewhere ?

matheuscscp: Yes, in the other 3 provider PRs

auth/get_token.go
matheuscscp: This function call here gives an early validation of the provider identity format with an informative error message to the user containing the service account reference. The alternative is repeating this error message inside the code of each provider when they get the identity internally for performing the token exchange. Since this needs to run even if cache is disabled, so I prefer to keep it up here

matheuscscp: Now that you mentioned I think I'm even gonna write a couple of test cases for this asserting the error message: invalid identity with cache disabled/enabled

matheuscscp: Tests added, thanks!

PR #907, opened 2025-04-15
[RFC-0010] Add aws auth library

matheuscscp:
This PR introduces the `aws` library from [RFC-0010](https://github.com/fluxcd/flux2/blob/main/rfcs/0010-multi-tenant-workload-identity/README.md).

Part of: https://github.com/fluxcd/flux2/issues/5022

PR Stack:

* [`azure`](https://github.com/fluxcd/pkg/pull/909)
* [`gcp`](https://github.com/fluxcd/pkg/pull/908)
* `aws` <---
* [`core`](https://github.com/fluxcd/pkg/pull/906)

PR #909, opened 2025-04-15
[RFC-0010] Add azure auth library

matheuscscp:
This PR introduces the `azure` library from [RFC-0010](https://github.com/fluxcd/flux2/blob/main/rfcs/0010-multi-tenant-workload-identity/README.md).

PR Stack:

* `azure` <---
* [`gcp`](https://github.com/fluxcd/pkg/pull/908)
* [`aws`](https://github.com/fluxcd/pkg/pull/907)
* [`core`](https://github.com/fluxcd/pkg/pull/906)


dipti-pai:
Wanted to register that I tried the azure library to enable single and multi-tenant workload identity with Azure Event Hub in this [PR ](https://github.com/fluxcd/notification-controller/pull/1106) and it works great! :) Also validated the proxy usage and caching of tokens. Thanks @matheuscscp!

auth/azure/client.go
```diff
@@ -51,6 +51,8 @@ type OptFunc func(*Client)
 // https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#NewDefaultAzureCredential
 // The default scope is to ARM endpoint in Azure Cloud. The scope is overridden
 // using OptFunc.
+//
+// Deprecated: Use auth.GetToken() with azure.Provider{} or NewTokenCredential().
```
stefanprodan, line 55: I propose we delete the client, the auth package is at v0.10.

dipti-pai (commented):
Would be nice to have some tests (where possible) for the new functionality.

auth/azure/new_default_azure_credential.go
```diff
@@ -0,0 +1,91 @@
...
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+)
+
+// newDefaultAzureCredential is like azidentity.newDefaultAzureCredential, but
+// does not call azidentity.NewAzureCLICredential().
+func newDefaultAzureCredential(options azidentity.DefaultAzureCredentialOptions) (azcore.TokenCredential, error) {
+	var (
```
dipti-pai, line 32: nit: can be const

auth/azure/options.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func getTenantID(serviceAccount corev1.ServiceAccount) (string, error) {
+	if tenantID, ok := serviceAccount.Annotations["azure.workload.identity/tenant-id"]; ok {
+		return tenantID, nil
+	}
+	if tenantID := os.Getenv("AZURE_TENANT_ID"); tenantID != "" {
+		return tenantID, nil
+	}
+	return "", fmt.Errorf("azure tenant ID not found in the service account annotations nor in the environment variable AZURE_TENANT_ID")
```
dipti-pai, line 49: can provide additional context in the error message by specifying the service account name.

auth/azure/options.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func getClientID(serviceAccount corev1.ServiceAccount) (string, error) {
+	if clientID, ok := serviceAccount.Annotations["azure.workload.identity/client-id"]; ok {
+		return clientID, nil
+	}
+	return "", fmt.Errorf("azure client ID not found in the service account annotations")
```
dipti-pai, line 56: can add service account name in the error message for additional context.

auth/azure/options.go
```diff
@@ -0,0 +1,79 @@
...
+
+	var conf *cloud.Configuration
+	switch {
+	case strings.HasSuffix(o.ArtifactRepository, ".azurecr.cn"):
+		conf = &cloud.AzureChina
+	case strings.HasSuffix(o.ArtifactRepository, ".azurecr.us"):
+		conf = &cloud.AzureGovernment
+	default:
+		conf = &cloud.AzurePublic
+	}
```
dipti-pai, line 72: this can be moved to a separate function since it is more about figuring out the cloud environment.

auth/azure/client.go
matheuscscp: Fixed, thanks!

auth/azure/new_default_azure_credential.go
matheuscscp: Fixed, thanks!

auth/azure/options.go
matheuscscp: Fixed, thanks!

auth/azure/options.go
matheuscscp: We already have it here (see our discussion in the core lib PR, we should have this in a single place): https://github.com/fluxcd/pkg/blob/7eae0914af240f06cbe139d8f649d51dc2e8b518/auth/get_token.go#L65-L70

auth/azure/options.go
matheuscscp: We already have it here (see our discussion in the core lib PR, we should have this in a single place): https://github.com/fluxcd/pkg/blob/7eae0914af240f06cbe139d8f649d51dc2e8b518/auth/get_token.go#L65-L70

matheuscscp:
> Would be nice to have some tests (where possible) for the new functionality.

IMO testing only errors does not add much value. Integration tests would be valuable, but we don't have credits for Azure, only AWS and GCP

auth/azure/options.go
dipti-pai: Thanks for clarifying, missed the core lib context as I was looking at this PR in isolation.

git/go.mod
```diff
@@ -3,7 +3,6 @@ module github.com/fluxcd/pkg/git
 go 1.24.0
 
 replace (
-	github.com/fluxcd/pkg/auth => ../auth
```
stefanprodan, line 6: Was this removed by mistake? If we don't replace it here, then we can't debug the git package with local changes to auth.

matheuscscp: No, it was removed intentionally because the git package depends on the old auth/azure code. I don't want to change the git libraries to use the new auth/azure package now in this PR, I want to do it separately when we implement the RFC for the git APIs down the road, so I'm freezing the packages that depend on the old auth/azure code for now. I'll re-add this line when we do it.

matheuscscp: Reverted and changed the packages depending on auth/azure to use the new code, thanks!

git/gogit/client.go
```diff
@@ -343,7 +343,7 @@ func (g *Client) providerAuth(ctx context.Context) error {
 			}
 			switch g.authOpts.ProviderOpts.Name {
 			case git.ProviderAzure:
-				g.authOpts.ProviderOpts.AzureOpts = append(g.authOpts.ProviderOpts.AzureOpts, azure.WithProxyURL(proxyURL))
+				g.authOpts.ProviderOpts.AuthOpts = append(g.authOpts.ProviderOpts.AuthOpts, auth.WithProxyURL(*proxyURL))
```
stefanprodan, line 346: Why is it `*proxyURL` and not `proxyURL` like for github?

auth/azure/token_credential.go
```diff
@@ -0,0 +1,58 @@
...
+func NewTokenCredential(ctx context.Context, opts ...auth.Option) azcore.TokenCredential {
+	return &tokenCredential{ctx, opts}
+}
+
+// GetToken implements exported.TokenCredential.
+// The context is ignored, use the constructor to set the context.
+// This is because some callers of the library pass context.Background()
+// when calling this method, so to ensure we have a real context we pass
+// it in the constructor.
+func (t *tokenCredential) GetToken(_ context.Context, tokenOpts policy.TokenRequestOptions) (azcore.AccessToken, error) {
```
stefanprodan, line 44: What stops the called from setting `context.Background()` in the contractor? Also who are these caller that we can't fix?

matheuscscp: SOPS

stefanprodan: Ok so mention sops

git/gogit/client.go
matheuscscp: Because this way when you call WithProxyURL() you can't pass nil, this function receives a copy of the object, with the intention to say: "hey, a proxy URL is not optional in this function, you have have a real URL object here to call this function"

auth/azure/token_credential.go
matheuscscp: Our code passes the right context in the constructor, while SOPS passes context.Background() to the Azure libraries.

matheuscscp: Fixed, thanks!

stefanprodan (approved):
LGTM

PR #917, opened 2025-05-02
[RFC-0010] Add tests for auth providers

matheuscscp:


stefanprodan (approved):
LGTM

Thanks @matheuscscp 🥇

PR #923, opened 2025-05-13
ssa: Implement `WaitForSetTermination`

stefanprodan:
This PR adds a helper function that ca be used in the Flux Garbage Collector to determine when the Kubernetes GC has finalized the objects marked for deletion.


matheuscscp (approved):
LGTM

PR #928, opened 2025-05-17
[RFC-0010] Add provider audience to cache key and decouple packages

matheuscscp:

After #925 the provider audience is no longer unique per provider, hence it needs to be part of the token cache key.

We are also decoupling the `auth` package from the `oci`, `git` and `git/gogit` packages, which makes it much easier to release changes in `auth` and also avoids forcing these dependants from including loads of indirect dependencies that are extremely unneeded.


stealthybox (approved):

matheuscscp:
All integration tests passing:

https://github.com/fluxcd/pkg/actions/runs/15111467033
https://github.com/fluxcd/pkg/actions/runs/15111472375
https://github.com/fluxcd/pkg/actions/runs/15111476445

PR #932, opened 2025-05-25
[RFC-0010] Add integration test for GCP workload identity federation

matheuscscp:

Important fix for GCP WIF. While writing this test I found out that the audience string can't have the `https:` prefix. It needs to start with `//`.

This test is passing locally with my personal GCP account, but for passing in our CI it needs the GCP IAM role [`roles/iam.workloadIdentityPoolAdmin`](https://cloud.google.com/iam/docs/roles-permissions/iam#iam.workloadIdentityPoolAdmin) to be added to the GCP Service Account that we use for applying terraform in the GitHub Action.

@stefanprodan Can you please add this role to the GH Action GCP SA?

Note: We don't need a test like this for the other clouds as the code path is only different for GCP. For AWS and Azure the cross-cloud and same-cloud code path is exactly the same, and for both of them the cloud setup involves registering the Issuer URL of the cluster. This was never the case for GCP up until introducing this new test, built-in federation for GKE registers the cluster Issuer URL automatically.

### DIFF
--- a/auth/gcp/options.go
+++ b/auth/gcp/options.go
@@ -54,5 +54,5 @@ func getWorkloadIdentityProviderAudience(serviceAccount corev1.ServiceAccount) (
 		return "", fmt.Errorf("invalid %s annotation: '%s'. must match %s",
 			key, wip, workloadIdentityProviderPattern)
 	}
-	return fmt.Sprintf("https://iam.googleapis.com/%s", wip), nil
+	return fmt.Sprintf("//iam.googleapis.com/%s", wip), nil
 }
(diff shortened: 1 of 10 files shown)


matheuscscp:
To add the role via the `gcloud` CLI, assuming you have ran `gcloud auth login` with Flux's GCP account, you can run this command:

```bash
gcloud projects add-iam-policy-binding <project-ID> --member=serviceAccount:<INCLUDE_HERE_THE_GCP_SA_EMAIL_USED_FOR_THE_GITHUB_ACTION> --role=roles/iam.workloadIdentityPoolAdmin
```

Or in the GCP Console you can go to the IAM page of the project and grant the role to the GCP SA, the screen should look like this:

Screenshot from 2025-05-25 13-16-50

stefanprodan:
I've added `IAM Workload Identity Pool Admin` to the GCP Service Account assigned to this repo.

stefanprodan (approved):
LGTM

PR #936, opened 2025-06-04
Introduce authentication for public.ecr.aws

drsherluck:
Some helm charts require authentication against the `public.ecr.aws` ECR registery e.g. [aws-node-termination-handler](https://github.com/aws/aws-node-termination-handler/tree/main/config/helm/aws-node-termination-handler).

Current flux v2.6.1 does not allow me to update my aws-node-termination-handler deployment due to this error: 
```
HelmChart 'flux-system/flux-system-aws-node-termination-handler' is not ready: unknown build error: failed to get credential from 'aws': failed to parse artifact repository 'public.ecr.aws/aws-ec2/helm': invalid AWS registry: 'public.ecr.aws'. must match ([0-9+]*).dkr.ecr(?:-fips)?\.([^/.]*)\.(amazonaws\.com[.cn]*|sc2s\.sgov\.gov|c2s\.ic\.gov|cloud\.adc-e\.uk|csp\.hci\.ic\.gov)
```
So this PR fixes that error by making `public.ecr.aws` a valid AWS registry.

### DIFF
--- a/auth/aws/provider.go
+++ b/auth/aws/provider.go
@@ -214,6 +214,12 @@ func (Provider) ParseArtifactRepository(artifactRepository string) (string, erro
 		return "", err
 	}
 
+	// Region is required to be us-east-1 for public.ecr.aws:
+	// https://docs.aws.amazon.com/AmazonECR/latest/public/public-registry-auth.html#public-registry-auth-token
+	if registry == "public.ecr.aws" {
+		return "us-east-1", nil
+	}
+
 	parts := registryRegex.FindAllStringSubmatch(registry, -1)
 	if len(parts) < 1 || len(parts[0]) < 3 {
 		return "", fmt.Errorf("invalid AWS registry: '%s'. must match %s",
--- a/auth/aws/provider_test.go
+++ b/auth/aws/provider_test.go
@@ -320,6 +320,11 @@ func TestProvider_ParseArtifactRepository(t *testing.T) {
 			artifactRepository: "gcr.io/foo/bar:baz",
 			expectValid:        false,
 		},
+		{
+			artifactRepository: "public.ecr.aws/foo/bar",
+			expectedRegion:     "us-east-1",
+			expectValid:        true,
+		},
 	}
 
 	for _, tt := range tests {


auth/aws/provider.go
```diff
@@ -214,6 +214,11 @@ func (Provider) ParseArtifactRepository(artifactRepository string) (string, erro
 		return "", err
 	}
 
+	// Region is required to be us-east-1 for public.ecr.aws (https://docs.aws.amazon.com/AmazonECR/latest/public/public-registry-auth.html#public-registry-auth-token)
```
matheuscscp, line 217:
```suggestion
	// Region is required to be us-east-1 for public.ecr.aws:
	// https://docs.aws.amazon.com/AmazonECR/latest/public/public-registry-auth.html#public-registry-auth-token
```

matheuscscp (commented):
Thanks very much for this PR!

JFYI this is actually not a fix, but a new feature. The previous versions of Flux never used authentication for this registry, Flux just silently ignored the `spec.provider` field set to `aws`. You can actually fix your deployment today by simply removing this field. See more information about this breaking change here:

https://fluxcd.io/blog/2025/05/flux-v2.6.0/#breaking-changes

Please amend the commit message to match what I put in the PR title :pray:

drsherluck:
> Please amend the commit message to match what I put in the PR title 🙏

I combined your suggestion with the amend of the commit message 👍🏻 

> I tested and this chart is really public, no authentication needed:

I'll remove the aws provider field from my HelmRepository. Thanks for the help!

I don't get why they add the authentication step in the documentation of the node-termination-handler chart installation guide. Makes it look like it's required 🤷🏻

stefanprodan:
> I don't get why they add the authentication step in the documentation of the node-termination-handler chart installation guide.

Amazon ECR Public does not require authentication but I suspect that once authenticated, the rate limits applied by AWS are more permissive. I see nothing wrong with us adding auth support for `public.ecr.aws`, but most people will not needed, unless the cluster pulls hundreds of charts on a short interval.

matheuscscp (approved):
Thanks very much @drsherluck! :rocket:

PR #938, opened 2025-06-05
Fix support for public.ecr.aws

matheuscscp:
I tested the current code for public.ecr.aws and it doesn't work. It's probably because public ECR has a different endpoint and we were trying to reach it through the Go SDK for normal ECR, but there's a dedicated Go SDK for public ECR. This PR is fixing this by using the appropriate Go SDK.

xref: #936

### DIFF
--- a/oci/tests/integration/terraform/aws/main.tf
+++ b/oci/tests/integration/terraform/aws/main.tf
@@ -23,6 +23,12 @@ module "test_ecr" {
   tags = var.tags
 }
 
+resource "aws_ecrpublic_repository" "test_ecr_public" {
+  repository_name = local.name
+  tags            = var.tags
+  force_destroy   = true
+}
+
 module "test_ecr_cross_reg" {
   source = "git::https://github.com/fluxcd/test-infra.git//tf-modules/aws/ecr"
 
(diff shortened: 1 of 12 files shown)


oci/tests/integration/terraform/aws/main.tf
```diff
@@ -23,6 +23,12 @@ module "test_ecr" {
   tags = var.tags
 }
 
+resource "aws_ecrpublic_repository" "test_ecr_public" {
+  repository_name = local.name
+  tags            = var.tags
+  force_destroy   = true
+}
```
stefanprodan, line 30: Do we clean this up at the end of the CI run?

matheuscscp: Yes, on terraform destroy. But we may need to add code in the reaper, same for GCP WI pools and providers

stefanprodan:
I've added Elastic Container Registry Public with Full access to the `oci-e2e` policy.

matheuscscp:
@stefanprodan Can you please add also a statement allowing `sts:GetServiceBearerToken` on the resource `*`?

```
An error occurred (AccessDeniedException) when calling the GetAuthorizationToken operation: User: arn:aws:sts::***:assumed-role/***/OCI_GH_Actions is not authorized to perform: sts:GetServiceBearerToken on resource: * because no identity-based policy allows the sts:GetServiceBearerToken action
```

PR #951, opened 2025-06-23
runtime/tls: remove deprecated package

### LINKED ISSUE #949: Consolidate the handling of secrets
In Flux we have a few widespread patterns of handling Secrets as part of the Flux APIs. Most notably:

- `.spec.secretRef` for authentication secrets, e.g. basic auth, docker config, bearer tokens, GitHub App, etc.
- `.spec.certSecretRef` for TLS secrets, i.e. CA and mTLS (fields `ca.crt`, `tls.crt` and `tls.key`, and deprecated fields `caFile`, `certFile` and `keyFile`).
- `.spec.proxySecretRef` for holding an HTTP/S proxy configuration (fields `address`, `username` and `password`).

The fact is however that the implementation is different in each place. As a result of [this](https://cloud-native.slack.com/archives/C01A402P0R2/p1750448321600989) CNCF Slack thread, we decided to consolidate the implementation of all these patterns across Flux in a library in this repository.

Action items:
- [x] Introduce a package `runtime/secrets` including:
  - [x] `func TLSConfigFromSecret(ctx context.Context, c client.Client, name, namespace string, opts ...Option) (*tls.Config, error)`
  - [x] `func ProxyURLFromSecret(ctx context.Context, c client.Client, name, namespace string) (*url.URL, error)`
  - [x] `func BasicAuthFromSecret(ctx context.Context, c client.Client, name, namespace string) (string, string, error)`
  - [x] `func PullSecretsFromServiceAccount(ctx context.Context, c client.Client, name, namespace string) ([]corev1.Secret, error)`
  - [x] `Make` counterpart functions to generate `*corev1.Secret` from the respective inputs. Ideally, all inputs from [here](https://github.com/fluxcd/flux2/blob/de594183bd8fe381d4e17d06a40972903a3e15ed/pkg/manifestgen/sourcesecret/sourcesecret.go#L134) should be supported.
- [x] Remove the `runtime/tls` package

The `TLSConfigFromSecret` function is the only one that needs to optionally support deprecated behavior, i.e. `caFile`, `certFile` and `keyFile`. By default the function should not support these fields, an option must be specified for support to be enabled. When support is enabled, preference should be given to a standard key, a deprecated key should be a fallback.

cappyzawa:
Closes #949 

Remove the deprecated `runtime/tls` package which uses deprecated field names (certFile, keyFile, caFile) for TLS secrets.

This is phase 2 of addressing https://github.com/fluxcd/pkg/issues/949, following https://github.com/fluxcd/pkg/pull/950 which added the replacement `runtime/secrets` package. The functionality is now superseded by the `runtime/secrets` package which provides better API design and supports both standard field names (tls.crt, tls.key, ca.crt) and deprecated field names via options.

GitHub search confirms this package is not used by any Flux components: https://github.com/search?q=org%3Afluxcd%20fluxcd%2Fpkg%2Fruntime%2Ftls&type=code


matheuscscp (approved):
Thanks!

PR #954, opened 2025-06-26
fix: Support rsa-sha2-512 and rsa-sha2-256 host key algos when ssh-rsa is supported

matheuscscp:
This PR fixes a regression in host key negotiation when the algorithms `rsa-sha2-512` or `rsa-sha2-256` are required.

This regression was introduced in [this](https://github.com/fluxcd/pkg/pull/943) PR, which was a fix for another regression, this one in the `go-git` dependency. Context [here](https://github.com/fluxcd/flux2/issues/5385).

This PR also introduces integration tests for Azure DevOps with SSH, which would have caught the regression this PR is fixing.

### DIFF
--- a/.github/workflows/integration-azure.yaml
+++ b/.github/workflows/integration-azure.yaml
@@ -50,6 +50,22 @@ jobs:
         run: cat .env
       - name: Build test app
         run: make docker-build
+      - name: Prepare Git SSH secrets
+        run: |
+          mkdir -p azure
+          cat <<EOF > azure/identity
+          $GIT_SSH_IDENTITY
+          EOF
+          cat <<EOF > azure/identity.pub
+          $GIT_SSH_IDENTITY_PUB
+          EOF
+          cat <<EOF > azure/known_hosts
+          $GIT_SSH_KNOWN_HOSTS
+          EOF
+        env:
+          GIT_SSH_IDENTITY: ${{ secrets.GIT_SSH_IDENTITY }}
+          GIT_SSH_IDENTITY_PUB: ${{ secrets.GIT_SSH_IDENTITY_PUB }}
+          GIT_SSH_KNOWN_HOSTS: ${{ secrets.GIT_SSH_KNOWN_HOSTS }}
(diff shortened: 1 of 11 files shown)


.github/workflows/integration-azure.yaml
```diff
@@ -50,6 +50,10 @@ jobs:
         run: cat .env
       - name: Build test app
         run: make docker-build
+      - name: Prepare Git SSH secret
+        run: echo $GIT_SSH_SECRET > git-ssh-secret.yaml
+        env:
+          GIT_SSH_SECRET: ${{ secrets.GIT_SSH_SECRET }}
```
matheuscscp, line 56: @stefanprodan Please add this secret to the repo, it's in the file `git-ssh-key.yaml` of the item `Matheus Azure Int Tests` in the fluxcd 1password (same we used for Azure int tests)

stefanprodan: Done

stefanprodan: Storing a whole YAML file in a secret is bad practice, GitHub could show the secret in logs due to formatting changes.

pjbgf (approved):
The git-related changes LGTM.

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

git/gogit/transport.go
```diff
@@ -146,6 +146,12 @@ func (a *CustomPublicKeys) ClientConfig() (*gossh.ClientConfig, error) {
 		config.Config.KeyExchanges = git.KexAlgos
 	}
 
+	// Whenever ssh-rsa is being used, prioritise sha2-512 and sha2-256.
+	// TODO: deprecate SHA1 signing scheme.
```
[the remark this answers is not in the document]
matheuscscp: In this repo or flux2?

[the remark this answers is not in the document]
stefanprodan: Let's do it in flux2 for max visibility

PR #955, opened 2025-06-26
runtime/secrets: add legacy field logging to TLS functions

### LINKED ISSUE #953: Add deprecated field detection logging to runtime/secrets package
## Summary

Enhance the `runtime/secrets` package with optional logging functionality to detect and warn when deprecated TLS field names are used in Kubernetes secrets. This will help Flux component maintainers identify and migrate away from deprecated field names (`certFile`, `keyFile`, `caFile`) to standard ones (`tls.crt`, `tls.key`, `ca.crt`).

**Alternative approaches or better solutions are welcome!** Please feel free to suggest different implementations or entirely different approaches to solve this visibility problem.

## Motivation

This enhancement was identified while implementing [notification-controller#1131](https://github.com/fluxcd/notification-controller/issues/1131), which migrates from the deprecated `runtime/tls` package to the new `runtime/secrets` package. During implementation, we discovered that while the `runtime/secrets` package supports both standard and deprecated field names through the `WithDeprecatedFieldSupport()` option, there's no mechanism to detect when these deprecated fields are actually being used in production environments.

Without visibility into deprecated field usage, it's challenging for component maintainers to:

1. Identify secrets using deprecated field names in their deployments
2. Plan migration strategies from deprecated to standard fields
3. Eventually phase out deprecated field support
4. Plan migration timelines for users
5. Understand the impact of eventually removing deprecated field support
6. Provide targeted guidance to users who need to update their secret configurations

This enhancement would provide visibility into deprecated field usage without requiring duplicate Secret fetches.

## Proposed Solution

Add a new `WithDeprecatedFieldLogging(logger logr.Logger)` option to the `TLSConfigFromSecret` function that:

1. Enables deprecated field support automatically
2. Logs warnings when deprecated fields are detected and used
3. Provides detailed information about which deprecated field was used and its recommended standard equivalent

### API Changes

```go
// New option function
func WithDeprecatedFieldLogging(logger logr.Logger) Option

// Usage example
logger := logr.FromContext(ctx)
tlsConfig, err := secrets.TLSConfigFromSecret(ctx, client, "my-secret", "default",
    secrets.WithDeprecatedFieldLogging(logger))
```

### Log Output Example

When a secret uses deprecated fields, the logger would output:

```
level=0 msg="using deprecated field in secret" secret="default/my-tls-secret" field="certFile" preferred="tls.crt"
level=0 msg="using deprecated field in secret" secret="default/my-tls-secret" field="keyFile" preferred="tls.key"
```

### Options struct updates

```go
type options struct {
    supportDeprecatedFields bool
    deprecatedFieldLogger   *logr.Logger
}
```

---

**Related Issues**:

- #949 (consolidate secret handling)
- #950 (add runtime/secrets package)

cappyzawa:
Resolves #953 .

## Summary

Add logging functionality to `TLSConfigFromSecret` that warns when legacy field names (`certFile`, `keyFile`, `caFile`) are used instead of standard Kubernetes field names (`tls.crt`, `tls.key`, `ca.crt`).

## Background

This addresses feedback from #953 discussion where the community decided to:

1. Make legacy field support mandatory across all Flux APIs for consistency
2. Require logging to encourage migration to standard field names
3. Prepare for eventual v2 API transition


matheuscscp:
Hey @cappyzawa since the discussion in #953 moved into standardizing even more and making the deprecated fields available everywhere, even for new APIs, I think we can actually remove the option we introduced with the first version of the library and change behavior as if it was set to `true` on every call`. Furthermore, we should still log (a nasty error :smiling_imp:) in case people use it to call out for using the standard Kubernetes keys. Furthermore, we should change the wording everywhere from `deprecation` to `legacy`, which gives more the notion of something that will be supported forever but should not be used. We should change this wording everywhere, code, docs, logs, etc.

So for this PR I'd remove the option that was merged in the first version of the library and make the logger a required argument.

cappyzawa:
Sorry! :cry: Added fixup commit dd7df7d to include all changes (some files weren't staged in the previous commit).

cappyzawa:
@matheuscscp DONE 👍

PR #961, opened 2025-07-02
runtime/secrets: use context-based logger retrieval

cappyzawa:
Addresses https://github.com/fluxcd/notification-controller/pull/1137#discussion_r2180561747

Updates TLSConfigFromSecret function to retrieve logger from context using log.FromContext(ctx) instead of accepting logger as a parameter, aligning with Flux's standard pattern.


matheuscscp (approved):
LGTM! Thanks very much!!! :pray:

PR #962, opened 2025-07-04
Fix scanManifests to respect ignore patterns

dgunzy:
Fixes #4921 by passing the ignorePaths string into the `scanManifests` function - and added a new function `shouldIgnoreFile` to process if it should be ignored based on the pattern. I added some test coverage but I can add more, or expand the tests. Let me know if this is the right approach @matheuscscp 

I compiled flux and imported the package locally and this was working compared to the release flux:
```
~/OpenSource/flux/flux2-fork fix/ignore-paths-diff !2 ?3 ❯ ./bin/flux diff kustomization test-app --path ./test-ignore --ignore-paths ".sops.yaml" --namespace test-flux-k-diff                          ⎈ default
✓  Kustomization diffing...
► Kustomization/test-flux-k-diff/test-app drifted

metadata
  + one map entry added:
    labels:
    │ kustomize.toolkit.fluxcd.io/name: test-app
    │ kustomize.toolkit.fluxcd.io/namespace: test-flux-k-diff

⚠️ identified at least one change, exiting with non-zero exit code
~/OpenSource/flux/flux2-fork fix/ignore-paths-diff !2 ?3 ❯ flux diff kustomization test-app --path ./test-ignore --ignore-paths ".sops.yaml" --namespace test-flux-k-diff                                ⎈ default
✓  Kustomization diffing...
✗ failed to generate kustomization.yaml: failed to decode Kubernetes YAML from /Users/danguns/OpenSource/flux/flux2-fork/test-ignore/.github/workflows/ci.yml: missing Resource metadata <nil> <nil>
~/OpenSource/flux/flux2-fork fix/ignore-paths-diff !2 ?3 ❯ 
```

### DIFF
--- a/kustomize/filters.go
+++ b/kustomize/filters.go
@@ -102,3 +102,39 @@ func filterSlice(ks *kustypes.Kustomization, path string, s *[]string, t string,
 	*s = (*s)[:start]
 	return nil
 }
+
+// shouldIgnoreFile returns true if the given file should be ignored based on
+// ignore patterns loaded from the base path and any additional ignore patterns provided.
+func shouldIgnoreFile(filePath, basePath, ignorePatterns string) bool {
+	if ignorePatterns == "" {
+		return false
+	}
+
+	absBase, err := filepath.Abs(basePath)
+	if err != nil {
+		return false
+	}
+
+	absFile, err := filepath.Abs(filePath)
+	if err != nil {
+		return false
+	}
+
+	ignoreDomain := strings.Split(absBase, string(filepath.Separator))
+	ps, err := sourceignore.LoadIgnorePatterns(absBase, ignoreDomain)
+	if err != nil {
+		return false
+	}
+
+	if ignorePatterns != "" {
+		ps = append(ps, sourceignore.ReadPatterns(strings.NewReader(ignorePatterns), ignoreDomain)...)
+	}
+
+	info, err := os.Lstat(absFile)
+	if err != nil {
+		return false
+	}
+
+	filter := ignoreFileFilter(ps, ignoreDomain)
+	return filter(absFile, info)
+}
--- a/kustomize/kustomize_generator.go
+++ b/kustomize/kustomize_generator.go
@@ -471,7 +471,7 @@ func (g *Generator) generateKustomization(dirPath string) (Action, string, error
 		return UnchangedAction, "", err
 	}
 
-	files, err := scanManifests(fs, abs)
+	files, err := scanManifests(fs, abs, g.ignore)
 	if err != nil {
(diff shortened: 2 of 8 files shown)


kustomize/filters.go
```diff
@@ -102,3 +102,39 @@ func filterSlice(ks *kustypes.Kustomization, path string, s *[]string, t string,
...
+		return false
+	}
+
+	absFile, err := filepath.Abs(filePath)
+	if err != nil {
+		return false
+	}
+
+	ignoreDomain := strings.Split(absBase, string(filepath.Separator))
+	ps, err := sourceignore.LoadIgnorePatterns(absBase, ignoreDomain)
```
matheuscscp, line 124: Looks like this function recursively reads the directory to find ignore files, and we're calling this once per traversed file. So this is close to O(N^2)

The function `filterKsWithIgnoreFiles` is calling this...

I think we need to load everything on a higher place of the code and pass down the patterns to all the places that need them. Would that be possible?

dgunzy: Good catch! I moved up the `LoadIgnorePatters` to the `WriteFile` method, and then passed the loaded patterns down through the call chain to make it more like O(N)  i think. Let me know if this is an ok approach!

kustomize/kustomize_generator.go
```diff
@@ -471,7 +500,12 @@ func (g *Generator) generateKustomization(dirPath string) (Action, string, error
 		return UnchangedAction, "", err
 	}
 
-	files, err := scanManifests(fs, abs)
+	if g.ignore != "" {
+		ignorePatterns = append(ignorePatterns,
+			sourceignore.ReadPatterns(strings.NewReader(g.ignore), ignoreDomain)...)
+	}
```
matheuscscp, line 506: Isn't this a duplication of lines 147-151?

kustomize/kustomize_generator.go
```diff
@@ -40,6 +40,8 @@ import (
 
 	"github.com/fluxcd/pkg/apis/kustomize"
 	securefs "github.com/fluxcd/pkg/kustomize/filesys"
+	"github.com/fluxcd/pkg/sourceignore"
+	"github.com/go-git/go-git/v5/plumbing/format/gitignore"
```
matheuscscp, line 44: Please move this import to the block above, alongside the external imports 🙏

matheuscscp (approved):
Thanks very much @dgunzy!!

@stefanprodan This LGTM, can you please take a look? 🙏

kustomize/kustomize_generator.go
dgunzy: Done!

PR #965, opened 2025-07-10
oci: Add `insecure` option to retry transport

stefanprodan:
Allow insecure registries when using the retry transport.

ref: https://github.com/fluxcd/flux2/issues/5448


matheuscscp (approved):
LGTM

PR #969, opened 2025-07-11
Bugfix: Perform http post in goroutine to prevent locking up the reconciler loop.

jvcdk:
## Brief description

This is a follow-up of PR https://github.com/fluxcd/kustomize-controller/pull/1479 and discussion https://github.com/fluxcd/flux2/discussions/5403 with a (hopefully) good fix to the issue.

The *very* short story of this issue is that *if* the notification controller (for some reason) is not available to receive Annotated Events, then the reconciliation loop of the Kustomize Controller can be locked up, delaying the reconciliation process.

This fix moves the http post action to a goroutine, effectively unblocking the reconciliation loop.

Please see detailed description in the PR above.

@matheuscscp :)

### DIFF
--- a/runtime/events/recorder.go
+++ b/runtime/events/recorder.go
@@ -250,16 +250,20 @@ func (r *Recorder) AnnotatedEventf(
 		return
 	}
 
-	// avoid retrying rate limited requests
-	if res, _ := r.Client.HTTPClient.Post(r.Webhook, "application/json", bytes.NewReader(body)); res != nil &&
-		(res.StatusCode == http.StatusTooManyRequests || res.StatusCode == http.StatusAccepted) {
-		return
-	}
+	// Execute the actual HTTP POST request in a goroutine to avoid blocking the
+	// reconciler loop.
+	go func() {
+		// avoid retrying rate limited requests
+		if res, _ := r.Client.HTTPClient.Post(r.Webhook, "application/json", bytes.NewReader(body)); res != nil &&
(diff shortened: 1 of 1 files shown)


tombtrifork (approved):

matheuscscp:
Thanks for addressing the issue in the right place :+1: 

I, however, don't like this kind of behavior. Imagine a controller handling a large number of objects with very short reconciliation intervals. The default amount of time we keeping retrying is a good amount of seconds. If your NC goes down, this will easily fill up the controller memory with goroutines (8 KiB each) stuck for several seconds in pointless retries. This could OOM the controller.

I honestly think it's better to let something like this surface in controller timeout errors rather than letting this problem be swallowed in error logs people may not be paying attention to. This change will hide malfunction of NC and make people think everything is OK, when they should actually be busy investigating why NC is not working and fixing it, or disabling it entirely if they don't care about notifications.

I'm not for merging this change.

jvcdk:
> > We have a reconciliation interval of 2 seconds
> 
> Why would you use such a low interval, the `Kustomization.spec.interval` is for undoing manual changes to the cluster, with a 2sec interval you are just DDOS-ing your Kubernetes API for nothing.

I might have misunderstood the reference `reconciliation interval`; I had the command line option `--requeue-dependency` in mind (but now that I dug out the option, I clearly see it is not the same). Our `Kustomization.spec.interval` is 15 minuttes.

stefanprodan:
Closing in favour of: #986

PR #970, opened 2025-07-11
runtime/secrets: add client-based API variants for TLS and Proxy functions

cappyzawa:
## Summary

Add client-based variants for TLS and Proxy functions as discussed in https://github.com/fluxcd/flux2/issues/5433#issuecomment-3063253727.

### File Separation
- `reader.go`: API server reading functions (`XXXFromSecretRef`)
- `converter.go`: Secret conversion functions (`XXXFromSecret`)

### New Functions
- `TLSConfigFromSecretRef`: creates TLS config from Secret reference  
- `ProxyURLFromSecretRef`: creates proxy URL from Secret reference

### Use Cases
These client-based variants are ideal for scenarios like:
- `spec.certSecretRef` where Secrets contain only TLS data
- `spec.proxySecretRef` where Secrets contain only proxy configuration
- Auto-fetching without manual Secret retrieval in controllers

## Notes
The core changes can be seen in [c929dfc](https://github.com/fluxcd/pkg/pull/970/commits/c929dfc8af4906efb300cb39611284da0119f902). Will squash commits upon approval.


matheuscscp (approved):
LGTM! 🚀

PR #972, opened 2025-07-12
tools: Introduce CLI tools for helping with Flux development tasks

matheuscscp:
## Release procedure

This process is intended to be run locally, in the clone of a Flux maintainer,
properly configured with commit signing and GitHub credentials.

First, a preparation PR must be created bumping all the Go modules
from this repository that have changed since their latest version.

Run the following commands:

1. `git pull`
2. `make prep`

If there are any changes, commit, open a PR `Prepare for release` and merge.

If no changes are needed, then:

1. `git pull`
2. `make release`

Both `make` commands will show a plan of the changes they
will make and ask for confirmation.


internal/enumerate_bumpable_modules.go
```diff
@@ -0,0 +1,28 @@
...
+package internal
+
+// EnumerateBumpableModules returns a list of Go modules that can be bumped
+// in the github.com/fluxcd/pkg repository. This includes all modules that
+// are taggable, as well as specific modules that are not tagged but should
+// still be considered for version bumps.
+func EnumerateBumpableModules(taggable []string) []string {
+	return append(taggable, []string{
+		"git/internal/e2e",
+		"oci/tests/integration",
```
stefanprodan, line 26: These 2 packages are hardcoded twice. I think we need to document that any new test pkg must be added in code. It would be good to declare this list only once.

matheuscscp: Fixed, thanks!

internal/module_bumps.go
```diff
@@ -0,0 +1,103 @@
...
+func (m *ModuleBumps) PrintTags() {
+	for _, tag := range m.tagsToPush {
+		fmt.Println(tag)
+	}
+}
+
+// PushTags pushes the tags to the remote repository.
+func (m *ModuleBumps) PushTags(ctx context.Context) error {
+	for _, tag := range m.tagsToPush {
+		tagCmd := exec.CommandContext(ctx, "git", "tag", "-s", "-m", tag, tag)
```
stefanprodan, line 89: I don't think this will work, on macOS for tags to be signed with the private key stored in the secure enclave, the shell must load `GPG_TTY` and prompt for fingerprint confirmation, but here there is no shell.

stefanprodan: I guess this would work `exec.Command("/opt/homebrew/bin/bash", "-c"`, ...)`

stefanprodan: Ok this will work if the key is is loaded into memory, so for macOS users this will work as long the Pinentry prompt ran before.

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #976, opened 2025-07-13
runtime/dependency: Simplify dependency sorting algorithm

matheuscscp:
#972 got me feeling a bit nostalgic about graph algorithms, so I decided to refactor our dependency sorting library into a much simpler depth-first search algorithm. Tarjan's SCC is pretty cool but at the same overkill for this use case.

In this PR I:

* Reduced our code.
* Deleted a package.
* Refactored tests to `gomega` (test cases remain the same).
* Added a test for the dependency cycle error string that should bubble up to controllers.


stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #978, opened 2025-07-14
runtime/secrets: Add `MakeSOPSSecret`

stefanprodan:
Add `MakeSOPSSecret` that would enable us to add a new command to Flux CLI `flux create secret sops`.


matheuscscp (approved):
LGTM

PR #981, opened 2025-07-14
ci: Fix missing permissions for `preview-release` workflow

stefanprodan:
Drop permissions to minimum required `contents: read`


matheuscscp (approved):
LGTM

I think the default is `read`, but good to make sure

PR #990, opened 2025-07-18
chartutil: Fix sort not recursing for slices

ymettier:
With chart values that contain double nested slices, sort will fail.

This patch contains:
- a test that contain a map with double nested slices
- a fix for this bug.

Without this patch, this will fail:

```
{
  "a": [[
    "p": "v-p"
    "r": "v-r"
    "q": v-q"
  ]]
}
```

### DIFF
--- a/chartutil/sort.go
+++ b/chartutil/sort.go
@@ -19,8 +19,18 @@ package chartutil
 import (
 	"sort"
 
-	goyaml "sigs.k8s.io/yaml/goyaml.v2"
+	goyaml "go.yaml.in/yaml/v2"
 )
+
+func sortSlice(s []interface{}) {
+	for _, item := range s {
+		if nestedMS, ok := item.(goyaml.MapSlice); ok {
+			SortMapSlice(nestedMS)
+		} else if nestedSlice, ok := item.([]interface{}); ok {
+			sortSlice(nestedSlice)
+		}
+	}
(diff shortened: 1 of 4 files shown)


chartutil/sort.go
```diff
@@ -22,6 +22,16 @@ import (
 	goyaml "sigs.k8s.io/yaml/goyaml.v2"
```
stefanprodan, line 22: Please replace this import in all chartutil files with `goyaml "go.yaml.in/yaml/v2"` and run `make tidy-chartutil`.
After this, in go.mod we should have `go.yaml.in/yaml/v2 v2.4.2`.

The reason for this is that `sigs.k8s.io/yaml/goyaml.v2` has been [deleted from upstream](https://github.com/kubernetes-sigs/yaml/tree/master/goyaml.v2#migration-guide)

ymettier: done (in an additional commit). Thanks for noticing it.

stefanprodan (approved):
LGTM

Thanks @ymettier

ymettier:
Commits squashed, amended with new msg and signed.

matheuscscp:
@ymettier Thanks for this contribution! I just pushed the tag `chartutil/v1.9.0` in case you wanna open a PR to bump it in helm-controller

PR #992, opened 2025-07-27
ssa: Add exact match option to field manager removal

stefanprodan:
Allow a SSA manager to take over field ownership using an exact name match. For example, override Helm CLI without affecting helm-controller.


matheuscscp (approved):
LGTM

PR #993, opened 2025-07-28
runtime/secrets: remove insecure parameter from TLS funcs

cappyzawa:
ref: https://github.com/fluxcd/flux2/issues/5433#issuecomment-3127327847


matheuscscp (approved):
LGTM! 🚀

PR #1000, opened 2025-08-04
auth/gcp: fix short-lived context issues with persistent TokenSource

cappyzawa:
## Problem

Currently, notification-controller and other FluxCD controllers create new `oauth2.TokenSource` instances on every request, leading to several issues:

1. **Short-lived context problems**: Controllers pass request-scoped contexts to `NewTokenSource()`, causing token acquisition to fail when the original request context is cancelled
2. **Performance inefficiency**: Creating new TokenSource instances for every request is wasteful
3. **Inconsistent authentication patterns**: Different controllers implement GCP authentication differently

## Why TokenSource Should Be 1:1 with ServiceAccount

Each ServiceAccount represents a unique identity with specific permissions and configuration:

1. **Identity separation**: Different ServiceAccounts may have different GCP service account annotations, workload identity providers, or scopes
2. **Cache efficiency**: Reusing the same TokenSource for the same identity avoids redundant token requests
3. **Security isolation**: Prevents token leakage between different identities

## Why TokenSource Needs Long-lived Context and Timeout-Equipped HTTP Client

TokenSource instances perform token acquisition through HTTP requests to GCP's STS (Security Token Service). This process requires careful context and HTTP client management:

### Context Lifecycle Issues
According to the [Go oauth2 documentation](https://pkg.go.dev/golang.org/x/oauth2), `TokenSource` automatically refreshes tokens "using the provided context". When controllers pass request-scoped contexts:

1. **Request-scoped context cancellation**: Token acquisition fails if the original request completes before token retrieval
2. **Background token refresh**: TokenSource performs automatic token refresh in background, but cancelled contexts prevent this essential operation
3. **Controller lifecycle mismatch**: TokenSource should live as long as the controller process, not individual requests

### HTTP Client Timeout Protection
The oauth2 documentation states that HTTPClient is used "only for token acquisition". Without proper timeouts:

1. **Hang prevention**: Token acquisition requests to GCP STS can hang indefinitely, blocking controller operations
2. **Network reliability**: 10-second timeout provides reasonable time for token acquisition while preventing indefinite waits
3. **Resource cleanup**: Timeout-equipped HTTP client ensures proper connection cleanup even when GCP services are slow

As documented in the [Go oauth2 package](https://pkg.go.dev/golang.org/x/oauth2), custom HTTP clients with timeouts should be embedded in context using `oauth2.HTTPClient`:

```go
httpClient := &http.Client{Timeout: 10 * time.Second}
ctx := context.WithValue(ctx, oauth2.HTTPClient, httpClient)
```

Note: Due to [known limitations](https://github.com/golang/oauth2/issues/368) in the oauth2 package, only the Transport is inherited from the context client, not timeout settings. Therefore, we configure timeouts at the Transport level.

### Token Acquisition Flow
```
Controller Request → TokenSource.Token() → HTTP Client → GCP STS
                                         ↓ (with timeout)
                  ← OAuth2 Token ← JSON Response ← 200 OK
```

When using short-lived contexts or clients without timeouts:
- Context cancellation terminates the HTTP request prematurely
- Missing timeouts can cause indefinite hangs on network issues
- Background token refresh fails due to cancelled context

This implementation uses `context.Background()` for TokenSource operations while embedding a timeout-equipped HTTP client, ensuring reliable token acquisition regardless of the original request context lifecycle, following [Go's context best practices](https://go.dev/blog/context-and-structs) for long-running background processes.

## Usage Example

See fluxcd/notification-controller#1156 for an example of how to migrate from direct TokenSource creation to the factory pattern.

### DIFF
--- a/oci/tests/integration/go.mod
+++ b/oci/tests/integration/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/elazarl/goproxy v1.7.2
 	github.com/fluxcd/cli-utils v0.36.0-flux.14
 	github.com/fluxcd/pkg/apis/meta v1.18.0
-	github.com/fluxcd/pkg/auth v0.23.0
+	github.com/fluxcd/pkg/auth v0.24.0
 	github.com/fluxcd/pkg/cache v0.10.0
 	github.com/fluxcd/pkg/git v0.35.0
 	github.com/fluxcd/pkg/git/gogit v0.38.0


matheuscscp:
Thanks very much for all the thought you put into this :hugs: 

> 1. **Short-lived context problems**: Controllers pass request-scoped contexts to `NewTokenSource()`, causing token acquisition to fail when the original request context is cancelled

The only controller that does that is notification-controller. It handles HTTP requests from the other controllers and then fires off a goroutine to do the actual send operation to the notification provider. We already solved this problem in notification-controller by using the appropriate context created on the beginning of the goroutine. All the other controllers perform all their operations synchronously with the `ctx` coming from the `Reconcile()` method of the reconciler, and no asynchronous operations are fired off in any other controllers like in notification-controller (if that's the case, my guess would be a mistake that needs to be fixed).

> 2. **Performance inefficiency**: Creating new TokenSource instances for every request is wasteful

Creating a `TokenSource` is practically a no-op:

https://github.com/fluxcd/pkg/blob/f227e67fddd8f2773dce0c4666b9b3eeda88dd6b/auth/gcp/token_source.go#L33-L36

From the caller perspective, it's a handful of `if <is option set> { append option }` statements, so it's not really a concern.

Also, it's a major design choice in Flux to never reuse clients across different reconciliations. As a multi-tenant software, Flux strives to consistently recreate clients on every single reconciliation to avoid client state leaking across tenant boundaries, which can potentially introduce bugs and vulnerabilities. The only exception is the controller-runtime client, for which there is a completely different story. We do share cached tokens when they are produced by the exact same inputs, and that's because we received reports of Flux getting rate-limited by STS services, so it's a carefully made decision.

> 3. **Inconsistent authentication patterns**: Different controllers implement GCP authentication differently

It's usually about the service being either a cloud-specific one, where the cloud offers an SDK that integrates with the respective `TokenSource` abstraction (`oauth2.TokenSource` for `gcp`, `azcore.TokenCredential` for `azure` and `aws.CredentialsProvider` for `aws`), or being an agnostic type, like OCI registries, HTTPS Git repositories and remote Kubernetes clusters. For agnostic services, we need to retrieve/exchange/build the token/credential ourselves, and feed it to the respective library, like `github.com/google/go-containerregistry` for OCI registries, `github.com/go-git/go-git` for Git repositories and `k8s.io/client-go/rest` for Kubernetes. There are also cases where the cloud service SDK is old or not designed to integrate with the token source abstraction, e.g. `azuredevops.NewPatConnection()` is the only way to use an Azure Managed Identity token with Azure DevOps at the moment, even for the commit status API.

The `auth` library was designed around the process of issuing and properly caching the tokens, which is especially needed when handling object-level authentication. But to also integrate well with the cloud provider SDKs, the respective token source interfaces were implemented, and they should be the preferred integration method for all cases unless the contrary is the only alternative, this is our core guideline when using the `auth` library. These token sources already handle caching. If a Flux object reconciliation needs to perform multiple operations using the token, the token source will seamlessly handle issuing new tokens only when needed, and when caching is enabled (the default).

The `oauth2.TokenSource` for `gcp` is unfortunately the only one that suffers from the anti-pattern of not receiving a `ctx` in the retrieve method, and this problem surfaced in the notification-controller due to its unique behavior of firing off asynchronous code. But as I mentioned initially, this is already solved. So this problem was really a problem only in single controller, for a single cloud provider.

From my perspective, most of this PR is introducing a significant redesign/drift between `gcp` and the other two. Such a change would warrant an entire redesign of the library, as we need to keep all providers the same. The changes include introducing a new caching mechanism that overlaps with the one carefully designed during the RFC discussion, and also breaks Flux's pattern of not caching clients for external APIs. I'd prefer not to move forward with this PR without a more holistic discussion in an RFC, but I also don't think the new mechanisms being introduced here are really solving any problems that need a discussion right now.

I'd be happy to see the `*http.Client` timeout improvement, though. To me it sounds reasonable to set this 10sec timeout on the clients used to fetch tokens. In `gcp` we already pass them via `context.WithValue(ctx, oauth2.HTTPClient, hc)`. We already pass the correct `ctx` everywhere, so we rely only the libraries to use it properly, e.g. `http.NewRequestWithContext()`, etc. For `oauth2` it may be that this `ctx` is not correctly used internally, so I'm fine setting this timeout on the `*http.Client`.

PR #1001, opened 2025-08-05
runtime/secrets: add GitHub App support to AuthMethodsFromSecret

cappyzawa:
## Summary

Add GitHub App authentication support to the `runtime/secrets` package to enable unified authentication handling across Flux controllers.

## Motivation

While reviewing fluxcd/source-controller#1860, I noticed there was an opportunity to integrate with the existing `runtime/secrets` pattern used by other controllers. This change provides a `GitHubAppData` type that can be directly used as input to `git/github.WithAppData()`, improving consistency and reducing code duplication.

## Implementation

- Add `GitHubAppData` type alias for direct compatibility with `github.WithAppData()`
- Integrate GitHub App support into `AuthMethodsFromSecret()` pattern
- Provide minimal validation and data extraction only - actual usage is delegated to `git/github.New()` and similar consumers
- Remove unused `GitHubAppAuth` struct to avoid confusion

The implementation maintains the existing philosophy: `runtime/secrets` handles secret parsing and basic validation, while specific authentication logic remains in the appropriate packages.

### DIFF
--- a/oci/tests/integration/go.mod
+++ b/oci/tests/integration/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/fluxcd/pkg/cache v0.10.0
 	github.com/fluxcd/pkg/git v0.35.0
 	github.com/fluxcd/pkg/git/gogit v0.38.0
-	github.com/fluxcd/pkg/runtime v0.78.0
+	github.com/fluxcd/pkg/runtime v0.79.0
 	github.com/fluxcd/test-infra/tftestenv v0.0.0-20250626232827-e0ca9c3f8d7b
 	github.com/go-git/go-git/v5 v5.16.2
 	github.com/google/go-containerregistry v0.20.6


runtime/secrets/secrets.go
```diff
@@ -243,14 +252,6 @@ type SSHAuth struct {
...
 }
 
-// GitHubAppAuth holds GitHub App authentication credentials.
-type GitHubAppAuth struct {
-	AppID          string
-	InstallationID string
-	PrivateKey     []byte
-	BaseURL        string
-}
-
```
stealthybox, line 253: how necessary is it to switch to `map[string]byte` here?
can we update callers to use this more type-safe version?

matheuscscp (approved):
LGTM! 🚀

matheuscscp:
@cappyzawa Can you please leave a comment on the related PR suggesting the use of what we added here? :pray: Thanks!

PR #1003, opened 2025-08-07
auth: add 10-second timeout to HTTP client for token acquisition

cappyzawa:
Following https://github.com/fluxcd/pkg/pull/1000#issuecomment-3160477432, add timeout setting only.

The `GetHTTPClient()` method now always returns an HTTP client with a 10-second timeout to prevent indefinite hangs during token acquisition from cloud provider STS services.

### DIFF
--- a/auth/options.go
+++ b/auth/options.go
@@ -19,6 +19,7 @@ package auth
 import (
 	"net/http"
 	"net/url"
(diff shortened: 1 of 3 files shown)


matheuscscp (approved):
LGTM! 🚀

PR #1007, opened 2025-08-14
runtime/secrets: remove ServerName pinning from TLS config

cappyzawa:
ref: https://github.com/fluxcd/flux2/issues/5477#issuecomment-3188140202


matheuscscp (approved):
LGTM! 🚀

PR #1008, opened 2025-08-15
[RFC-0010] auth: add lockdown support for multi-tenant workload identity

cappyzawa:
Part of: fluxcd/flux2#5465


auth/options.go
```diff
@@ -32,7 +32,6 @@ type Option func(*Options)
 // Options contains options for configuring the behavior of the provider methods.
 // Not all providers/methods support all options.
 type Options struct {
-	Client          client.Client
```
cappyzawa, line 35: Client field removed to prevent ambiguity between explicit client parameter and Options.Client. 
All auth functions now require client.Client as an explicit mandatory parameter to ensure consistent usage and avoid unintended client selection.

auth/utils/restconfig.go
```diff
@@ -84,11 +101,20 @@ func GetRESTConfig(ctx context.Context,
 	}
 
 	// Get provider by name.
-	provider, err := ProviderByName[auth.RESTConfigProvider](cm.Data[meta.KubeConfigKeyProvider])
+	providerName := cm.Data[meta.KubeConfigKeyProvider]
+	provider, err := ProviderByName[auth.RESTConfigProvider](providerName)
```
matheuscscp, line 105: `ProviderByName` should handle creating the provider

```suggestion
	provider, err := ProviderByName[auth.RESTConfigProvider](providerName, kubeClient)
```

auth/generic/provider.go
```diff
@@ -36,7 +36,10 @@ import (
...
 
 // Provider implements the auth.Provider interface for generic authentication.
-type Provider struct{ Implementation }
+type Provider struct {
+	Client client.Client
+	Implementation
+}
 
 // GetName implements auth.RESTConfigProvider.
 func (p Provider) GetName() string {
```
matheuscscp, line 45: Please remove the `p` receiver where it is not needed :pray: 

```suggestion
func (Provider) GetName() string {
```

matheuscscp (commented):
Great work!

auth/generic/provider.go
```diff
@@ -36,7 +36,10 @@ import (
 const ProviderName = "generic"
 
 // Provider implements the auth.Provider interface for generic authentication.
-type Provider struct{ Implementation }
+type Provider struct {
+	Client client.Client
+	Implementation
+}
```
matheuscscp, line 42: Now that `generic.Provider` is a complex type with construction inputs, let's make it fully encapsulated:

```suggestion
type provider struct {
	client client.Client
	Implementation
}

// Option is a functional option for configuring the generic authentication provider.
type Option func(*provider)

// WithImplementation sets the implementation for the generic authentication provider.
func WithImplementation(impl Implementation) Option {
	return func(p *provider) {
		p.Implementation = impl
	}
}

// NewProvider creates a new generic authentication provider.
func NewProvider(kubeClient client.Client, opts ...Option) *provider {
	p := &provider{
		client: kubeClient,
	}
	for _, opt := range opts {
		opt(p)
	}
	return p
}
```

matheuscscp:
Hmm but for this to be possible we also need the namespace, not just the client...

So I guess we need to make the namespace mandatory too :/ Sorry for the late bad news :(

The good news is that the crazy RESTConfig adapter already receives a namespace 😅

cappyzawa:
I had the same concern about how to implement the lockdown logic properly.

Initially, I was planning to implement the lockdown logic in the pkg side as mentioned in the issue (fluxcd/flux2#5465). However, I realized that when a ServiceAccount is already provided in the opts, the ServiceAccount token acquisition logic is already implemented.

So I thought maybe we could control this at the controller side by adding the default ServiceAccount to the options when calling these auth functions.

cappyzawa:
For the controller-side approach, I was imagining something like this:

```go
if obj.Spec.ServiceAccountName != "" {
    opts = append(opts, auth.WithServiceAccount(client.ObjectKey{
        Name:      obj.Spec.ServiceAccountName,
        Namespace: obj.Namespace,
    }))
} else if auth.GetDefaultServiceAccount() != "" {
    // Use default ServiceAccount (same namespace as the resource)
    opts = append(opts, auth.WithServiceAccount(client.ObjectKey{
        Name:      auth.GetDefaultServiceAccount(),
        Namespace: obj.Namespace,
    }))
}
```

matheuscscp:
I usually don't like global vars, my goal with them here was to make it possible for the lib to:

1) Handle this behavior transparently and effectively reduce the complexity in the controller side; and
2) Be a guaranteed final line of defense when the flags are enabled, i.e. eliminate the risk of some controller not implementing the behavior (hence why I'd also like to see a breaking change in this PR to force us to review the `auth` lib integration everywhere, `auth.WithServiceAccount()` is doing the job :+1:).

This is what the current implementation of the feature gate is trying to accomplish. It's not perfect, though, as we still need to handle the error by stalling the object, and this can only be done individually in each controller, as the way we stall is not uniform everywhere right now. So the feature gate is already adding to the burden on the controller side, as each controller integration needs to come with at least one minimal test for the `Stalled` condition/error. Also, note that the lib is [**also**](https://github.com/fluxcd/pkg/blob/4ca8fb011f0c15804649fc7ff178ff13fb8722c4/auth/access_token.go#L59-L64) enforcing the feature gate behavior, as a final line of defense like mentioned above. (By the way, notification-controller is again an exception in all of this, as it is not really a controller.)

If we decided to free the lib from the responsibility of implementing the lockdown behavior, then it would be better to follow the established pattern of passing controller flags into the `Reconciler` structs, and into whatever deeper contexts where the flag may be needed, from one function to another.

So, honestly, I wish integrating the `auth` lib was even simpler than what it is right now, so I'm leaning more towards the original plan...

The fact that we are passing the `kubeClient` as a mandatory argument when it's not really mandatory is also starting to annoy me now, I'm starting to regret this idea :sweat_smile: If the flag and object field are not set, then `kubeClient` is not really needed (except for controller-level in the `generic` provider). The same applies to passing the `namespace` as mandatory. Also, there are places like in the `flux <push|pull|diff|tag|list> artifact` CLI commands where a `kubeClient` and `namespace` are not even available, like everywhere else (these commands have a `--provider` flag and authenticate with environment credentials, i.e. the controller-level code path).

So the controller-level implementation for the `generic` provider is doing this, right now:

https://github.com/fluxcd/pkg/blob/4ca8fb011f0c15804649fc7ff178ff13fb8722c4/auth/generic/provider.go#L52-L54

Which is really just for avoiding a panic in a potential code path where the provider is created with a `nil` `kubeClient`. This would be a bug in the code, and not a configuration mistake, so ideally I'd like the *compiler* to catch it, and not Flux running in production :cry: Maybe we should let it panic, as there should never be a code path passing `nil` to the `generic` provider (at least when using controller-level authentication). Catching a catastrophic mistake is easier than a mistake that surfaces with more moderation. But, at the same time, we have `recover()` in place, so controllers would never get killed by a `panic()` anyway. So it's probably better to have an easier-to-read error... :exploding_head: 😵‍💫

(Note: The `generic` provider may be relevant for more use cases in the future, going [beyond](https://github.com/goharbor/harbor/issues/22027) just kubeconfig.)

With all these things considered ( :weary: ), I think the most correct option from an API design perspective, and from the perspective of trying to reduce the controller side complexity as much as possible and enforce the lockdown without mistakes, we should make it like this:

```go
func WithClient(client client.Client) Option // keep this option, revert mandatory kubeClient arg (sorry!!)
func WithServiceAccountNamespace(namespace string)
func WithServiceAccountName(name string) Option // breaking change. will prevent us from missing any places that will necessarily need to pass WithClient() and WithServiceAccountNamespace() for the lockdown
```

What do you think? Does this all make sense?

cappyzawa:
@matheuscscp 

Your API design proposal makes perfect sense! I completely agree that having the lockdown logic in the auth package ensures consistent behavior across all controllers, and making `kubeClient` and `namespace` optional via Options is much cleaner. The breaking change to `WithServiceAccountName()` is also a good idea to force us to review all call sites.

I should have proposed this approach earlier. The current mandatory `kubeClient` parameter indeed feels awkward when it's not always needed.

So the API would be:
```go
func WithClient(client client.Client) Option                
func WithServiceAccountNamespace(namespace string) Option   
func WithServiceAccountName(name string) Option            
```

And the lockdown logic would check if ServiceAccountName is not set but default SA is configured, then use the default with the provided namespace.

I'll implement this approach in this PR.

auth/options.go
```diff
@@ -32,19 +32,28 @@ type Option func(*Options)
...
+	ClusterAddress          string
+	AllowShellOut           bool
+}
+
+// ShouldGetServiceAccountToken returns true if ServiceAccount token should be retrieved.
+func (o *Options) ShouldGetServiceAccountToken() bool {
+	// ServiceAccount namespace is required because ServiceAccounts are namespace-scoped resources.
+	// ServiceAccountName can be empty as it may be provided by DEFAULT_SERVICE_ACCOUNT or
+	// DEFAULT_KUBECONFIG_SERVICE_ACCOUNT environment variables.
+	return o.ServiceAccountNamespace != ""
```
matheuscscp, line 56: This logic is incomplete. If neither `o.ServiceAccountName` nor the env vars are set, we should not fetch a service account token.

```suggestion
	return o.ServiceAccountNamespace != "" &&
		(o.ServiceAccountName != "" || hasDefaultServiceAccount())
```

And:

```go
func hasDefaultServiceAccount() {
	// Here we can detect a default service account by checking either the default
	// service account or the default kubeconfig service account because these two
	// are supposed to never be set simultaneously. The controller main functions
	// must ensure this property.
	return GetDefaultServiceAccount() != "" || GetDefaultKubeconfigServiceAccount() != ""
}
```

matheuscscp: Edited a little bit.

auth/registry.go
```diff
@@ -121,10 +122,14 @@ func GetArtifactRegistryCredentials(ctx context.Context, provider ArtifactRegist
 	var serviceAccount *corev1.ServiceAccount
 	var providerIdentity string
 	var audiences []string
-	if o.ServiceAccount != nil {
+	if o.ServiceAccountName != "" {
```
matheuscscp, line 125:
```suggestion
	if o.ShouldGetServiceAccountToken() {
```

auth/utils/restconfig.go
```diff
@@ -99,14 +99,10 @@ func GetRESTConfig(ctx context.Context,
...
 	if name, ok := cm.Data[meta.KubeConfigKeyServiceAccountName]; ok {
-		saKey := client.ObjectKey{
-			Name:      name,
-			Namespace: namespace,
-		}
-		opts = append(opts, auth.WithServiceAccount(saKey, ctrlClient))
-	} else {
-		opts = append(opts, auth.WithClient(ctrlClient))
+		opts = append(opts, auth.WithServiceAccountName(name))
+		opts = append(opts, auth.WithServiceAccountNamespace(namespace))
```
matheuscscp, line 105:
```suggestion
```

Namespace here should be passed 100% of the times. If we don't pass it before the `if`, lockdown will always return `false` in `o.ShouldGetServiceAccountToken()`.

matheuscscp (commented):
Last of the batch!

matheuscscp (commented):
Looking much better now!

auth/controller_flags_test.go
```diff
@@ -0,0 +1,98 @@
...
+func TestSetDefaultServiceAccount(t *testing.T) {
+	g := NewWithT(t)
+	auth.SetDefaultServiceAccount("test-sa")
+	t.Cleanup(func() {
+		os.Unsetenv(auth.EnvDefaultServiceAccount)
+	})
+	g.Expect(os.Getenv(auth.EnvDefaultServiceAccount)).To(Equal("test-sa"))
+}
+
+func TestSetDefaultKubeconfigServiceAccount(t *testing.T) {
```
matheuscscp, line 37:
```suggestion
func TestSetDefaultKubeConfigServiceAccount(t *testing.T) {
```

`Kubeconfig` -> `KubeConfig`

auth/controller_flags_test.go
```diff
@@ -0,0 +1,98 @@
...
+	})
+
+	t.Run("returns empty when not set", func(t *testing.T) {
+		g := NewWithT(t)
+
+		g.Expect(auth.GetDefaultServiceAccount()).To(Equal(""))
+	})
+}
+
+func TestGetDefaultKubeconfigServiceAccount(t *testing.T) {
```
matheuscscp, line 70:
```suggestion
func TestGetDefaultKubeconfigServiceAccount(t *testing.T) {
```

`Kubeconfig` -> `KubeConfig`

auth/controller_flags_test.go
cappyzawa: 😢

cappyzawa:
@matheuscscp Done!
Thanks for your suggestions 👍

matheuscscp:
* aws: https://github.com/fluxcd/pkg/actions/runs/17011765679
* azure: https://github.com/fluxcd/pkg/actions/runs/17011767286
* gcp: https://github.com/fluxcd/pkg/actions/runs/17011768464

matheuscscp:
> * aws: https://github.com/fluxcd/pkg/actions/runs/17011765679
> * azure: https://github.com/fluxcd/pkg/actions/runs/17011767286
> * gcp: https://github.com/fluxcd/pkg/actions/runs/17011768464

Passed!

PR #1009, opened 2025-08-16
auth: eliminate unreasonable use of env vars

matheuscscp:
I regret my choice of using env vars for representing internal global state. It's absolutely insane polluting the user-facing properties of the process just to benefit from `t.Setenv()`.

I'm also fixing/refactoring/improving a few tests.

This test will need to be fixed after this PR: https://github.com/fluxcd/source-controller/blob/24412ed278c8d298d47c068598b9db7a74a3e9e5/internal/controller/bucket_controller_test.go#L1395-L1397


cappyzawa (commented):
LGTM 🥇

cappyzawa (approved):

auth/controller_flags.go
```diff
@@ -40,15 +40,15 @@ const (
 const (
 	// EnvDefaultServiceAccount stores the default service account name
 	// for workload identity.
-	EnvDefaultServiceAccount = "DEFAULT_SERVICE_ACCOUNT"
+	EnvDefaultServiceAccount = "PKG_AUTH_DEFAULT_SERVICE_ACCOUNT"
```
stefanprodan, line 43: Rename from `PKG_AUTH` to `WID` please

matheuscscp: Fixed, thanks!

stefanprodan (approved):

PR #1010, opened 2025-08-17
tests: move testdata and integration tests to tests

### LINKED ISSUE #975: Move `oci/tests/integration` to `tests/integration`
The module `oci/tests/integration` is no longer only for `oci`, but rather for all our cloud provider integration tests, including Git and KubeConfig  In fact, the package `oci` is not (!) exercised in this test. OCI integrations are tested, but only through `auth` and `github.com/google/go-containerregistry`.

So we should move `oci/tests/integration` to `tests/integration`.

matheuscscp:
Closes #975


matheuscscp:
> Tests are running:
> 
> * aws: https://github.com/fluxcd/pkg/actions/runs/17020454954
> * azure: https://github.com/fluxcd/pkg/actions/runs/17020455658
> * gcp: https://github.com/fluxcd/pkg/actions/runs/17020456572

They passed

stefanprodan (approved):
LGTM

PR #1011, opened 2025-08-17
auth: add object-level configuration validation

cappyzawa:
ref: https://github.com/fluxcd/source-controller/pull/1872#discussion_r2280877451


matheuscscp:
@cappyzawa `auth/v0.27.0` is released :ok_hand:

PR #1016, opened 2025-09-01
ssa: Add strict field validation option for dry-run

stefanprodan:
⚠️ The `FieldValidation` option is only available for `client.Patch()`. Given that `client.Apply()` is suppose to replace `client.Patch()`, and the `FieldValidation` is not an option of `client.Apply()`, merging this could put us in a tough situation where we'll not be able to upgrade controller-runtime once they remove the deprecated `client.Patch(ctx, obj, client.Apply, opts)`.


stefanprodan:
Closing this PR, Strict mode is the only way SSA works so this is redundant.

PR #1022, opened 2025-09-06
chore: upgrade ssa structured-merge-diff package to v6

ricardo-mng:
Upgrade the only package that is still using structured-merge-diff/v4

This is a blocker for other project upgrades.


stefanprodan (approved):
LGTM

Thanks @ricardo-mng good catch!

PR #1023, opened 2025-09-08
runtime: Add environment variables

stefanprodan:
These env vars are used by all Flux controllers


runtime/controller/env.go
```diff
@@ -0,0 +1,27 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+const (
+	// RuntimeNamespaceEnvKey is the environment variable key used to
+	// specify the namespace in which the controller is running.
+	RuntimeNamespaceEnvKey = "RUNTIME_NAMESPACE"
```
matheuscscp, line 22:
```suggestion
	EnvRuntimeNamespace = "RUNTIME_NAMESPACE"
```

runtime/controller/env.go
```diff
@@ -0,0 +1,27 @@
...
+
+	// SourceLocalhostEnvKey is the environment variable key used to
+	// specify the localhost address of the source-controller file server.
+	SourceLocalhostEnvKey = "SOURCE_CONTROLLER_LOCALHOST"
```
matheuscscp, line 26:
```suggestion
	EnvSourceLocalhost = "SOURCE_CONTROLLER_LOCALHOST"
```

matheuscscp (commented):
Like in `auth`

PR #1031, opened 2025-09-22
ssa: Add force apply config to dry-run diff

stefanprodan:
Part of: https://github.com/fluxcd/flux2/issues/3032


matheuscscp (approved):
LGTM! :rocket:

PR #1038, opened 2025-10-06
runtime/secrets: validate proxy URL scheme and length

cappyzawa:
ref: https://github.com/fluxcd/notification-controller/pull/1190

While reviewing the negative diff in that PR, I noticed that the `address` key in the secret referenced by proxySecretRef is missing some validations.


matheuscscp:
@cappyzawa `runtime/v0.87.0` is released :ok_hand:

PR #1055, opened 2025-12-04
HelmRelease drift detection generates JSON patch that replaces entire `metadata` instead of `annotations` and `labels`

yozel:
### Describe the bug

When drift detection computes a JSON patch for changes limited to `metadata.annotations` and `metadata.labels`, the generated patch targets `/metadata` and replaces the entire `metadata` object, instead of adding only `annotations` and `labels`.

This leads to an incorrect patch that would drop required metadata fields such as `name`, `namespace`, `uid`, etc.

### To Reproduce

Given the following current object in the cluster:

```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: configmap-data
  namespace: default
```

And the desired object rendered by Helm is:

```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: configmap-data
  namespace: default
  annotations:
    foo: bar
  labels:
    baz: qux
```

Drift detection produces the following JSON patch:

```json
[
  {
    "value": {
      "annotations": {
        "foo": "bar"
      },
      "labels": {
        "baz": "qux"
      }
    },
    "op": "add",
    "path": "/metadata"
  }
]
```

In JSON Patch semantics, `op: "add"` with `path: "/metadata"` on an existing key replaces the entire `metadata` object. This means every existing field under `metadata` that is not part of this value is effectively removed.

### Expected behavior

The generated JSON patch should target only the changed subtrees under `metadata`, for example:

```json
[
  {
    "value": {
      "foo": "bar"
    },
    "op": "add",
    "path": "/metadata/annotations"
  },
  {
    "value": {
      "baz": "qux"
    },
    "op": "add",
    "path": "/metadata/labels"
  }
]
```

This way, existing fields under `metadata` such as `name`, `namespace`, `uid`, `resourceVersion`, and any other keys are preserved.

### Fix

This PR preserves metadata.name in copyAnnotationsAndLabels and that prevents the resulting metadata to be empty, which solves the issue with wrong JSON patch

### DIFF
--- a/ssa/jsondiff/testdata/annotated-configmap.yaml
+++ b/ssa/jsondiff/testdata/annotated-configmap.yaml
@@ -1,0 +1,7 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: "configmap-data"
+  annotations:
+    prev: some-previous-value
--- a/ssa/jsondiff/unstructured.go
+++ b/ssa/jsondiff/unstructured.go
@@ -246,6 +246,14 @@ func copyAnnotationsAndLabels(obj *unstructured.Unstructured) *unstructured.Unst
 		Object: make(map[string]interface{}),
 	}
 
+	// When annotations and labels were missing on the cluster object but present on the desired object, drift
+	//	detection produced an "add /metadata" patch. JSON Patch interprets this as replacing the entire metadata block,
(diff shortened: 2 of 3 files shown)


stefanprodan:
Please rebase with upstream main and force push. Thanks!

matheuscscp (approved):
LGTM! :rocket:

PR #1061, opened 2026-01-07
git: Preserve sparse checkout directories when switching branches

kane8n:
# Description
## Problem
When cloning a repository with SparseCheckoutDirectories option, the sparse checkout setting was not preserved when calling SwitchBranch(). This caused a full checkout of all files instead of only the specified directories.

## Root Cause
The SwitchBranch() method in client.go did not pass SparseCheckoutDirectories to wt.Checkout(), unlike the clone functions in clone.go which correctly handle this option.

## Solution
- Added sparseCheckoutDirectories field to the Client struct to store the sparse checkout configuration
- Store the sparse checkout directories during clone operations (cloneBranch, cloneTag, cloneCommit, cloneSemVer)
- Pass the stored sparseCheckoutDirectories to wt.Checkout() in SwitchBranch()

## Changes
- git/gogit/client.go: Added sparseCheckoutDirectories field to Client struct and updated SwitchBranch() to use it
- git/gogit/clone.go: Store SparseCheckoutDirectories from clone options into the client
- git/gogit/client_test.go: Added test case to verify sparse checkout is preserved after branch switch

### DIFF
--- a/git/gogit/client.go
+++ b/git/gogit/client.go
@@ -69,15 +69,16 @@ const ClientName = "go-git"
 // Client implements repository.Client.
 type Client struct {
 	*repository.DiscardCloser
-	path                 string
-	repository           *extgogit.Repository
-	authOpts             *git.AuthOptions
-	storer               storage.Storer
-	worktreeFS           billy.Filesystem
-	credentialsOverHTTP  bool
-	useDefaultKnownHosts bool
-	singleBranch         bool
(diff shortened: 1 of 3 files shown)


matheuscscp (approved):
LGTM

matheuscscp:
Thanks @kane8n! `git/gogit/v0.43.0` is tagged, can you please implement this in source-controller and image-automation-controller? Thanks!

PR #1077, opened 2026-01-19
ssa: introduce custom apply stage

matheuscscp:
This custom stage allows Role and RoleBinding objects to be applied on the same call to `ApplyAllStaged` even if the impersonated ServiceAccount does not have a ClusterRoleBinding for the `cluster-admin` ClusterRole (which bypasses the existence check for the Role on the dry-run of the RoleBinding).

### DIFF
--- a/ssa/manager_apply.go
+++ b/ssa/manager_apply.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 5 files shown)


matheuscscp:
Supersedes #1073

ssa/manager_apply.go
```diff
@@ -154,7 +159,6 @@ func (m *ResourceManager) ApplyAll(ctx context.Context, objects []*unstructured.
 		g, ctx := errgroup.WithContext(ctx)
 		g.SetLimit(m.concurrency)
 		for i, object := range objects {
-			i, object := i, object
```
erikgb: Is this change unrelated to the purpose of the PR, or have I misunderstood? Maybe do it in a separate PR?

ssa/manager_apply.go
```diff
@@ -300,6 +309,19 @@ func (m *ResourceManager) ApplyAllStaged(ctx context.Context, objects []*unstruc
...
 
+	// Apply custom staged objects next, if any, and wait for them to become ready.
+	if len(customStage) > 0 {
+		cs, err := m.ApplyAll(ctx, customStage, opts)
+		if err != nil {
+			return changeSet, err
+		}
+		changeSet.Append(cs.Entries)
+
+		if err := m.WaitForSet(cs.ToObjMetadataSet(), WaitOptions{opts.WaitInterval, opts.WaitTimeout, false}); err != nil {
```
stefanprodan: I don't think waiting here is the right thing, in some cases like custom resources waiting will not work at all unless they use the -1 generation hack like Flux.

erikgb (commented):
This looks great, and I am really looking forward to seeing this in a future FluxCD release.

PR #1082, opened 2026-01-29
ssa: fail faster when only terminal states are observed

matheuscscp:
This fix is similar to https://github.com/helm/helm/pull/31730

It allows helm-controller not to pass `FailFast=true` when waiting for hooks.

### DIFF
--- a/tests/integration/go.mod
+++ b/tests/integration/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/fluxcd/pkg/auth v0.36.0
 	github.com/fluxcd/pkg/cache v0.13.0
 	github.com/fluxcd/pkg/git v0.41.0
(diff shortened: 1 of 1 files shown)


ssa/manager_wait.go
```diff
@@ -112,11 +113,22 @@ func (m *ResourceManager) WaitForSetWithContext(ctx context.Context, set object.
...
+			}
+
+			// If only Failed or Current statuses are present,
+			// we can consider this a terminal state.
+			terminal := true
+			for s := range countPerStatus {
+				if s != status.FailedStatus && s != status.CurrentStatus {
+					terminal = false
+					break
+				}
```
stefanprodan: We already have countFailed, so i don't think we need a countPerStatus.

PR #1085, opened 2026-02-03
ssa: handle Jobs with TTL during health check

stefanprodan:
Job with TTL can be deleted after completion during health checking. This PR allows callers to specify which Jobs have TTL enabled so that `ssa.Wait` will not report not found errors for these.

Changes:
- Added `ExtractJobsWithTTL` helper function to `ssa/utils` package
-  Added `JobsWithTTL` field to `WaitOptions` and updated error handling to treat `NotFoundStatus` for Jobs with TTL as success


matheuscscp (approved):
LGTM! :rocket: 

Great improvement and fix for helm-controller kstatus poller :ok_hand:

PR #1089, opened 2026-02-09
Eliminate go-git from various packages

matheuscscp:
* Remove go-git from sourceignore, by vendoring the small part of gitignore that we use there
* Move git/github to auth/githubapp so we can import GitHub App auth without go-git
* Upgrade go-git to v5.16.5


sourceignore/gitignore/matcher.go
```diff
@@ -0,0 +1,33 @@
+// This package was originally copied from the `go-git` project:
```
stefanprodan: Add Flux copyright in all go files before the "This package was originally copied from the `go-git` project"

PR #1116, opened 2026-02-14
cmd: introduce command for bumping modules in controller repos

matheuscscp:
Automatically bumps `fluxcd/pkg` dependencies in a controller (or flux2) repository:

1. Detects the target repo's branch and controller name.
2. Maps them to the correct `fluxcd/pkg` branch using a baseline version table.
3. Clones pkg in-memory to find the latest module tags reachable from that branch.
4. Updates `go.mod` accordingly.

I tested this command locally in the PR https://github.com/fluxcd/notification-controller/pull/1238

The command has a temporary workaround for Flux 2.8 due to the fact that we branched here before releasing 2.8. This workaround will make our lives easier in the next 2 weeks. After that, we should remove this workaround and 2.8 patches will work as intended without workarounds.

### DIFF
--- a/cmd/internal/bump_deps.go
+++ b/cmd/internal/bump_deps.go
@@ -1,0 +1,284 @@
+/*
+Copyright 2026 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


cmd/internal/bump_deps.go
```diff
@@ -0,0 +1,324 @@
...
+
+	if err := os.WriteFile(gomod, []byte(newContent), 0644); err != nil {
+		return nil, fmt.Errorf("failed to write %s: %w", gomod, err)
+	}
+	return updates, nil
+}
+
+// goModTidy runs "go mod tidy" in the given directory.
+func goModTidy(ctx context.Context, dir string) error {
+	cmd := exec.CommandContext(ctx, "go", "mod", "tidy")
```
stefanprodan: This leaves stale packages in go.sum, we should be running `make tidy` from the controller repo, which deletes go.sum, we also have test modules that need to be tidy up. I would remove this shell out, and leave it to the workflow to run this.

stefanprodan (approved):
LGTM

PR #1129, opened 2026-02-22
refactor: move DependencyReference to apis/meta

### LINKED ISSUE #1034: Extract `DependencyReference` to `apis/meta`
The `DependencyReference` type is now duplicated in KC and HC API packages. We should move it to `apis/meta`  and extend the Flux CLI to accept CEL expression in: `utils.MakeDependsOn`.

Iam-Karan-Suresh:
### Summary

This PR moves the duplicated `DependencyReference` type from controller-specific APIs into `apis/meta` to provide a single shared implementation across Flux components.

### Changes

* Added `DependencyReference` to `apis/meta`
* Updated `ObjectWithDependencies` to use `[]DependencyReference`
* Added `MakeDependsOn` helper with CEL (`@expr`) support
* Updated dependency sorting logic and tests accordingly
* Added deepcopy implementations for the new type

### Motivation

`DependencyReference` was duplicated across KC and HC APIs. Centralizing it in `apis/meta` removes duplication and enables consistent dependency handling, including CEL-based readiness expressions used by the Flux CLI.

### Notes

* Backward behavior preserved.
* Follow-up update will be required in `fluxcd/flux2` to use the shared type in `utils.MakeDependsOn`.

---


apis/meta/reference_types.go
```diff
@@ -43,6 +43,36 @@ func (in NamespacedObjectReference) String() string {
...
+	// of a dependency. When specified, the built-in readiness check
+	// is replaced by the logic defined in the CEL expression.
+	// To make the CEL expression additive to the built-in readiness check,
+	// the feature gate `AdditiveCELDependencyCheck` must be set to `true`.
+	// +optional
+	ReadyExpr string `json:"readyExpr,omitempty"`
+}
+
+// String implements the fmt.Stringer interface for DependencyReference.
+func (in DependencyReference) String() string {
```
matheuscscp, line 70: Is this being used anywhere in this repo or in the controllers? It doesn't encode the `ReadyExpr` field... Wondering if it's really needed

runtime/dependency/sort.go
```diff
@@ -56,7 +56,11 @@ func Sort(objects []Dependent) ([]meta.NamespacedObjectReference, error) {
 			Namespace: obj.GetNamespace(),
 		}
 		vertices = append(vertices, u)
-		for _, v := range obj.GetDependsOn() {
+		for _, dep := range obj.GetDependsOn() {
+			v := meta.NamespacedObjectReference{
```
matheuscscp, line 60: I was gonna ask for changing to `DependencyReference` here everywhere, but I think keeping `NamespacedObjectReference` is better, it's one less string per edge, saves up some memory 👍

apis/meta/dependencies.go
```diff
@@ -16,9 +16,41 @@ limitations under the License.
...
 
+import "strings"
+
 // ObjectWithDependencies describes a Kubernetes resource object with dependencies.
 // +k8s:deepcopy-gen=false
 type ObjectWithDependencies interface {
-	// GetDependsOn returns a NamespacedObjectReference list the object depends on.
-	GetDependsOn() []NamespacedObjectReference
+	// GetDependsOn returns a DependencyReference list the object depends on.
+	GetDependsOn() []DependencyReference
```
matheuscscp, line 25: This is a breaking change, and `apis/meta` is GA (`v1.x`). I'm ok with it, though, so I'll defer to other maintainers

apis/meta/dependencies.go
```diff
@@ -16,9 +16,41 @@ limitations under the License.
...
+}
+
+// MakeDependsOn parses a list of dependency strings into DependencyReference
+// objects. Each dependency string can be in one of the following formats:
+//   - "name" - a dependency in the same namespace with no CEL expression
+//   - "namespace/name" - a dependency in a specific namespace
+//   - "name@readyExpr" - a dependency with a CEL readiness expression
+//   - "namespace/name@readyExpr" - a dependency in a specific namespace with a CEL expression
```
matheuscscp, line 33: This is new when compared to what's in the flux2 repo: https://github.com/fluxcd/flux2/blob/7d27a26665b34463b0ed5fb54fa06642375536f9/internal/utils/utils.go#L234

Wondering if this is the best way to add the feature. Is `@` prohibited in Kubernetes object names and namespaces?

matheuscscp (commented):
After this change is released, we need to replace the types in kustomize-controller and helm-controller in a non-breaking way:

```go
type DependencyReference = meta.DependencyReference
````

File: apis/meta/dependencies.go (hunk 4 above)
Comment (Iam-Karan-Suresh) [reply apis/meta/dependencies.go:33]: The `@` symbol is only used as a separator in the string form of a dependency reference (for example, inside annotations). It is **not part of actual Kubernetes resource names**.

The parts before and after `@` still follow normal Kubernetes naming rules:

* `my-deployment@ready.status == "ok"` → `my-deployment` is a valid resource name
* `prod/my-service@custom.condition` → `prod` (namespace) and `my-service` (name) are both valid

The idea behind using `@` comes from familiar patterns like emails or URLs, where it clearly separates two different pieces of information. Here, it helps distinguish the resource identifier from the condition or expression in a clean and readable way.

We could document this more clearly in code comments if needed. Alternatives like `:` or `||` are possible, but `@` works well because it’s easy to read and doesn’t conflict with Kubernetes-reserved characters like `/` or `:`.

File: apis/meta/reference_types.go (hunk 1 above)
Comment (Iam-Karan-Suresh) [reply apis/meta/reference_types.go:70]: @matheuscscp  I fixed and updated the pr.

File: runtime/go.mod [hunk 5]
```diff
@@ -29,12 +29,12 @@ require (
 	github.com/stretchr/testify v1.11.1
 	go.uber.org/zap v1.27.1
 	golang.org/x/net v0.49.0
-	k8s.io/api v0.35.2
+	k8s.io/api v0.35.0
 	k8s.io/apimachinery v0.35.2
```
matheuscscp, line 33: Do we need this bump in this PR?

apis/meta/go.mod
```diff
@@ -2,16 +2,21 @@ module github.com/fluxcd/pkg/apis/meta
 
 go 1.25.0
 
-require k8s.io/apimachinery v0.35.2
+require (
+	github.com/onsi/gomega v1.38.2
+	k8s.io/apimachinery v0.35.2
```
matheuscscp, line 7: Ditto

apis/meta/dependencies.go
matheuscscp: We should add links to this PR of the k8s docs saying that `@` cannot be part of resource names/namespaces

runtime/go.mod
Iam-Karan-Suresh: The k8s.io/apimachinery v0.35.2 bump is likely incidental from go mod tidy, not intentional for this refactor. Since meta now requires v0.35.2 and runtime depends on meta, the version was automatically resolved during module tidying. I remove the bump if you wish @matheuscscp

Iam-Karan-Suresh:
@matheuscscp  i added the links on the command. happy to iterate based on suggestions

runtime/go.mod
matheuscscp: Ok, let's keep it

apis/meta/dependencies.go
```diff
@@ -16,9 +16,47 @@ limitations under the License.
...
+//
+// The @ symbol is used to separate the resource reference from the CEL expression.
+// Note that @ cannot be part of resource names or namespaces per Kubernetes naming conventions:
+// https://kubernetes.io/docs/concepts/overview/working-with-objects/names-and-ids/
```
matheuscscp: This link is giving me a 404

Iam-Karan-Suresh: Sorry for the typo error. link fixed now

apis/meta/dependencies.go
matheuscscp: @stefanprodan This is the only item left open IMO

stefanprodan (approved):
LGTM

matheuscscp:
@Iam-Karan-Suresh Can you pls rebase and force-push?

matheuscscp:
@Iam-Karan-Suresh There's another couple of features blocked on this, pls rebase as soon as you can :pray:

Iam-Karan-Suresh:
Hi @matheuscscp. I fixed the merge conflit. Sorry for the delay 🙏

PR #1131, opened 2026-02-23
ssa: fix dry-run failure when upgrading API version

mehrdadbn9:
## Summary

When upgrading a custom resource to a new API version (e.g., `v1beta1` to `v1beta3`), the server-side apply dry-run fails with an error like:

```
dry-run failed: .spec.accessPolicy: field not declared in schema
```

This happens because the managed fields in the existing object still reference the old API version, and Kubernetes validates those fields against the new schema during dry-run. Fields that have default values in the new version but didn't exist in the old version cause schema validation errors.

## Fix

This PR introduces a new `migrateAPIVersion` function that:
1. Detects when the desired object has a different API version than the existing object
2. Uses the existing `PatchMigrateToVersion` function to update the API version in managed fields
3. Applies the patch before the dry-run to ensure proper schema validation

The fix is applied in:
- `Apply()` - single object apply
- `ApplyAll()` - batch object apply  
- `Diff()` - drift detection

## Testing

- Added unit tests for `PatchMigrateToVersion` with various scenarios
- Added integration test for API version migration workflow
- Code compiles and passes `go vet`

Fixes: fluxcd/flux2#5715

## Reproduction

The issue can be reproduced with a CRD like Linkerd's `Server` resource:

1. Apply a resource with `policy.linkerd.io/v1beta1`
2. Update the manifest to use `policy.linkerd.io/v1beta3`
3. Flux dry-run fails with schema validation error for default fields

With this fix, the API version migration is handled automatically before dry-run validation.


matheuscscp:
Superseded by: #1174

PR #1133, opened 2026-02-26
ssa: fix WaitForSetWithContext waiting until timeout if set is empty

Orzelius:
Before this fix, if ResourceManager.WaitForSetWithContext is called with an empty set it will wait for the duration`ReconcileTimeout` is set to and return without an error.

fix: just return early if the set is empty

I judged this to be too basic of a change to be worth adding tests for, but let me know if you think otherwise.


Orzelius:
Yep, we're working on adding server side apply functionality to Talos Linux (https://github.com/siderolabs/talos/pull/12829) and have gone with flux's ssa package. I spent a good five minutes trying to figure out what was being waited for just to realize that I had called the Wait method with an empty set and figured that fixing it upstream will help other people who run into the same thing as well.

Also the `ResourceManager.Wait` method already has the same check
https://github.com/Orzelius/fluxcd-pkg/blob/fc6028abd509c6cdf8ecb08b3589fae7603fbaf4/ssa/manager_wait.go#L74-L76

Orzelius:
> Please fix the signoff

fixed

PR #1144, opened 2026-03-08
runtime: Migrate to new Kubernentes event API

### LINKED ISSUE #5761: Implement Kubernetes Event API v1
Transition the Flux ecosystem from legacy `core/v1` events to the modern `events.k8s.io/v1` API. The following changes have to follow a backward compatibility approach where legacy fields  may be bridged to the new schema (e.g., keep `InvolvedObject` regardless of being renamed, add a new `RelatedObject` field) to ensure no breaking changes for users relying on current event logs.

**Action Items**
- [x] Add a new AnnotatedEventRecorder interface upstream in `client-go/tools/events` - kubernetes/kubernetes#138103
- [x] Controller-Runtime - kubernetes-sigs/controller-runtime#3509
- [x] Update `fluxcd/pkg` to use `events.k8s.io/v1` - fluxcd/pkg#1144
- [ ] Update all controllers using the changes done in `runtime/events` from `fluxcd/pkg`. Use new field `relatedObject` to include object's source metadata (as InvolvedObject does) and `Action` as mandatory field.
   - [ ] Mandatory controllers
      - [ ] Notification-Controller - fluxcd/notification-controller#1267
         - [ ]  Adapt the event receiver to handle the events.k8s.io/v1 schema.
         - [ ] Update [spec/v1beta3/events.md](https://github.com/fluxcd/notification-controller/blob/main/docs/spec/v1beta3/events.md) to document new fields.
      - [ ] Helm-Controller fluxcd/helm-controller#1510
      - [ ] Kustomize-Controller fluxcd/kustomize-controller#1621
      - [ ] Source-Controller fluxcd/source-controller#2011
   - [ ] Optional controllers
      - [ ] Source-Watcher fluxcd/source-watcher/pull/343
      - [ ] Image-Automation-Controller fluxcd/image-automation-controller#1038
      - [ ] Image-Reflector-Controller fluxcd/image-reflector-controller#889

**Reference to the new API spec and changes**
- [API Deprecation Guide](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#event-v125) ( `events.k8s.io/v1beta1` deprecated since k8s 1.25) 
- [KEP-383](https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/383-new-event-api-ga-graduation/README.md): New Event API GA Graduation

adri1197:
Event Recorder implementation to fully support the Kubernetes Events API `events.k8s.io/v1` (replacing `k8s.io/client-go/tools/record` to `k8s.io/client-go/tools/events` module). The recorder implements the `events.EventRecorder` interface, enabling compatibility with modern Kubernetes event handling while maintaining backward compatibility for existing webhook-based notifications.

Part of fluxcd/flux2#5761


runtime/events/recorder.go
```diff
@@ -202,12 +203,12 @@ func (r *Recorder) AnnotatedEventf(
...
 	// traces are persisted as Kubernetes events only as normal events.
 	if severity == eventv1.EventSeverityTrace {
-		r.EventRecorder.AnnotatedEventf(object, annotations, corev1.EventTypeNormal, reason, messageFmt, args...)
+		r.EventRecorder.Eventf(object, nil, corev1.EventTypeNormal, reason, action, messageFmt, args...)
 		return
 	}
 
 	// Forward the event to the Kubernetes recorder.
-	r.EventRecorder.AnnotatedEventf(object, annotations, eventtype, reason, messageFmt, args...)
+	r.EventRecorder.Eventf(object, nil, eventtype, reason, action, messageFmt, args...)
```
matheuscscp:
```suggestion
	r.EventRecorder.Eventf(object, related, eventtype, reason, action, messageFmt, args...)
```

matheuscscp (commented):
The lib should support `related` and let callers decide whether or not to use it

matheuscscp:
> The lib should support `related` and let callers decide whether or not to use it

We should also add `related` as a constant in `fluxcd/pkg/apis/event` and send the related object reference under this field to notification-controller

stefanprodan (commented):
Please undo all changes to `events/v1beta1` and add `events/v1`.

Also go.mod changes are unrelated, please undo.

runtime/go.mod
```diff
@@ -11,30 +11,30 @@ replace (
 
 require (
 	github.com/AdaLogics/go-fuzz-headers v0.0.0-20240806141605-e8a1dd7889d6
-	github.com/fluxcd/cli-utils v1.2.2
```
adri1197, line 14: There seems to be a dependency with controller-runtime, which downgrades cli-utils to `github.com/fluxcd/cli-utils v1.2.0`. Shall it be upgraded there first? - fluxcd/cli-utils#38

apis/event/v1/action.go
```diff
@@ -0,0 +1,47 @@
...
+	ActionValidating string = "Validating"
+	// ActionValidated indicates successful validation.
+	ActionValidated string = "Validated"
+	// ActionWaiting indicates waiting for a condition.
+	ActionWaiting string = "Waiting"
+	// ActionProgressing indicates progression through a workflow.
+	ActionProgressing string = "Progressing"
+	// ActionFailed indicates a failed operation.
+	ActionFailed string = "Failed"
+)
```
matheuscscp, line 47: Wondering where are these used across Flux 🤔

runtime/events/recorder.go
```diff
@@ -48,70 +48,65 @@ import (
...
-		Client:              httpClient,
-		EventRecorder:       mgr.GetEventRecorderFor(reportingController),
-		Log:                 log,
-	}, nil
+// NewRecorder creates an events.EventRecorder with a Kubernetes event recorder
+// and an external event recorder based on the given webhook. The recorder
+// performs automatic retries for connection errors and 500-range response codes
+// from the external recorder.
+func NewRecorder(mgr ctrl.Manager, log logr.Logger, webhook, reportingController string, opts ...RecorderOption) (EventRecorder, error) {
+	return NewRecorderForScheme(mgr.GetScheme(), mgr.GetEventRecorder(reportingController), log, webhook, reportingController, opts...)
```
matheuscscp: I think we can delete all these other constructors, and have only the one accepting variadic options

apis/event/v1/action.go
adri1197: These action constants are about aligning with the `events.k8s.io/v1` API which requires an action field. The plan is for Flux controllers to adopt them when they migrate to the new event recorder interface, keeping a value convention across. This is just a proposal, can be changed.

runtime/events/recorder.go
adri1197: Good point! Consolidated into a single `NewRecorder(log, webhook, reportingController, opts...)` with `WithManager(mgr, reportingController)`, `WithScheme(scheme)`, and `WithEventRecorder(er)` options. `NewRecorderForScheme` is removed

stefanprodan:
@adri1197 please rebase to take https://github.com/fluxcd/pkg/pull/1292

adri1197:
> @adri1197 please rebase to take #1292

Done! 😃

runtime/events/recorder.go
```diff
@@ -48,43 +48,59 @@ import (
...
+// NewRecorder creates an EventRecorder with a Kubernetes event recorder and an
+// external event recorder based on the given webhook. The recorder performs
+// automatic retries for connection errors and 500-range response codes from the
 // external recorder.
-func NewRecorder(mgr ctrl.Manager, log logr.Logger, webhook, reportingController string) (*Recorder, error) {
+//
+// The scheme and Kubernetes event recorder can be provided either via
+// WithManager (common case) or via WithScheme and WithEventRecorder
+// (for tests or custom setups).
+func NewRecorder(log logr.Logger, webhook, reportingController string, opts ...RecorderOption) (EventRecorder, error) {
```
matheuscscp, line 103: @adri1197 Before this change, the return type `*Recorder` used to be one that implements the old upstream Go interface. Can we keep returning an upstream interface? Is there one that is suitable? I was wondering why are we defining the new interface `EventRecorder` and concluded that it's for returning it here, right? Anything preventing us from returning just an upstream Go interface?

adri1197: Good question @matheuscscp — A couple of clarifications:

- On returning an upstream interface from tools/events: the blocker is that our recorder exposes both `Eventf` and
`AnnotatedEventf`, but upstream deliberately splits these across two interfaces:
  - `events.EventRecorder` → `Eventf` only.
  - `events.AnnotatedEventRecorder` → `AnnotatedEventf` only (documented as an "optional extension of EventRecorder").

- There is no single upstream interface that carries both methods, so returning one forces a tradeoff:
  - Return `events.EventRecorder`: drops `AnnotatedEventf` from the returned type. Therefore, callers that need would have to type-assert to `events.AnnotatedEventRecorder`. Since annotation scraping (<group>/ keys off the object) is the main reason this recorder exists, hiding it behind a type assertion is a regression.
  - Return `events.AnnotatedEventRecorder`: keeps full capability (`Eventf` is just ```AnnotatedEventf(obj, related, nil, ...))```, but drops the convenience `Eventf` from the surface, which is the more common call in controllers.

So yes — the custom `EventRecorder` interface exists precisely to expose both methods behind one type, matching
what the concrete recorder does and what consumers embed.

If minimizing our exported surface is the priority, the cleanest upstream-only option is to return `events.EventRecorder` and let callers type-assert for annotations. But I'd lean toward keeping the custom interface for the ergonomics. Happy to go either way — what's your view? 😃

matheuscscp: Then we can use interface embedding for this:

```go
type Recorder interface { // lets rename EventRecorder just to Recorder, the package is already called events
	events.EventRecorder
	events.AnnotatedEventRecorder
}
```

adri1197: Good point! Done! 😃

matheuscscp: Thanks!

matheuscscp (approved):
LGTM! 🚀 

Thanks @adri1197 for pushing through! This PR has come a long way! It has even triggered upstream changes in Kubernetes. You should feel proud!

stefanprodan (approved):
LGTM

Thanks @adri1197 🏅

PR #1166, opened 2026-04-13
deps: update google.golang.org/api to v0.275.0

guidopili:
The `github.com/fluxcd/pkg/auth` package is currently using `v0.261.0` of the `google.golang.org/api` dependency, with an indirect dependency that is affected by [CVE-2026-33186](https://nvd.nist.gov/vuln/detail/CVE-2026-33186). 

This PR upgrades the library to the latest available version, bumping the indirect dependency to a not-affected version.


stefanprodan:
We'll bump all deps before Flux 2.9 release end of Q2, no need to do it now, more CVEs will come.

PR #1170, opened 2026-04-14
tar: Add support for creating gzip-compressed tarballs

### LINKED ISSUE #323: tar: Add TarGzip func
Following up from refactoring the untar pkg, we could add a `.TarGzip` func to tar-gzip directories.

> Can we add a TarGzip function that takes a directory and generates a tgz file? We have that in the tests and I think it would be nice to offer both functions in this new package.
source: https://github.com/fluxcd/pkg/pull/221#discussion_r786033016

Depends on https://github.com/fluxcd/pkg/pull/221 being merged.

stefanprodan:
Add `Tar` func that writes a sanitized archive of a directory tree, skipping symlinks and other non-regular,
non-directory entries. The `Tar` function accepts the following options:
- `WithSkipGzip` to write  a plain tar stream instead of a gzip-compressed tarball
- `WithFilter` to exclude entries by `path` or `FileInfo`

The tar package now includes symlink resolution utilities for archiving, which should be used in:
-  `flux build artifact --resolve-symlinks`
-  `flux push artifact --resolve-symlinks`

This PR also refactors the existing tests and reorganises the package structure, adds proper doc.go with usage examples.

Closes: #323


tar/options.go
```diff
@@ -0,0 +1,79 @@
...
+limitations under the License.
+*/
+
+package tar
+
+import "os"
+
+// TarOption configures the behavior of Tar and Untar. Options are
+// silently ignored by operations they do not apply to.
+type TarOption func(*tarOpts)
```
matheuscscp: Unless `Option` already exists in this package for other purposes, I'd rename this to `Option`, otherwise we have stuttering: `tar.TarOption` (I see `TarOption` already exists, but I think it'd be ok to rename)

stefanprodan: This has the potential to break all consumers, but since the goal is to release `v1.0.0` of this package, we can actually break it.

stefanprodan: Renamed to `Option`

matheuscscp (approved):
LGTM! 🚀

PR #1174, opened 2026-04-15
ssa: introduce support for migrating API version

matheuscscp:
Part of: https://github.com/fluxcd/flux2/issues/5715

Supersedes: #1131

This PR has tests that reproduce the issue with external-secrets with `MigrateAPIVersion` set to `false`, then fixes the issue when `MigrateAPIVersion` is set to `true`.

The new options introduced for the `PatchMigrateToVersion` function are for scoping the migration down to what Flux controllers actually do: they do not touch subresources and should not touch managed field entries belonging to other managers (the cleanup of kubectl and override managers is responsible for that, in `cleanupMetadata`).

### DIFF
--- a/ssa/manager_apply.go
+++ b/ssa/manager_apply.go
@@ -68,6 +68,16 @@ type ApplyOptions struct {
 	// CustomStageKinds defines a set of Kubernetes resource types that should be applied
 	// in a separate stage after CRDs and before namespaced objects.
 	CustomStageKinds map[schema.GroupKind]struct{} `json:"customStageKinds,omitempty"`
+
+	// MigrateAPIVersion, when enabled, rewrites every managed fields entry
+	// on the existing object to match the API version of the applied object
+	// before each apply.
+	//
+	// This is needed after a CRD adds a new version that introduces fields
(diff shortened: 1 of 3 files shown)


ssa/manager_apply.go
```diff
@@ -108,6 +113,15 @@ func (m *ResourceManager) Apply(ctx context.Context, object *unstructured.Unstru
 		return m.changeSetEntry(object, SkippedAction), nil
 	}
 
+	var patched bool
+	if opts.MigrateAPIVersion && getError == nil {
+		var err error
+		patched, err = m.migrateAPIVersion(ctx, existingObject, object.GetAPIVersion())
+		if err != nil {
+			return nil, err
```
stefanprodan:
```suggestion
			return nil, fmt.Errorf("%s failed to migrate API version: %w", utils.FmtUnstructured(existingObject), err)
```

matheuscscp: Fixed, thanks!

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #1178, opened 2026-04-19
tests/intergration: cleanup AWS ENIs

matheuscscp:
Fix race leaving ENIs behind.

Tested here: https://github.com/fluxcd/pkg/actions/runs/24649785720


stefanprodan (approved):
LGTM

Thanks @matheuscscp

PR #1183, opened 2026-04-21
runtime/cel: allow empty kind in status reader

matheuscscp:
Part of: https://github.com/fluxcd/flux2/issues/5858


stefanprodan (approved):
LGTM

Please add this to the roadmap for Flux 2.9

PR #1186, opened 2026-04-23
Remove autoscaling/v2beta2 from HPA normalization

### LINKED ISSUE #1185: Normalize package does not compile against Kubernetes 1.36
Hello! I'm trying to upgrade to Kubernetes 1.36, and it looks like the Normalize package won't compile against it.

`k8s.io/api v0.36.0` removed the `autoscaling/v2beta2` Go package so anyone consuming ssa/normalize will run into compilation errors:

```bash
normalize/normalize.go:24:2: no required module provides package k8s.io/api/autoscaling/v2beta2; to add it:                                     
        go get k8s.io/api/autoscaling/v2beta2
```

It's probably safe to remove that normalization case since https://github.com/kubernetes/kubernetes/issues/118293 was fixed, and the buggy Kubernetes version is quite old and no longer supported.

guineveresaenger:
This pull request removes normalization logic for `v2beta2 HorizontalPodAutoscaler` which was removed as an API in kubernetes v1.36.

I used Claude Code a bit to establish context but the changes were all done by me.

Fixes #1185.


stefanprodan:
Flux support Kubernetes N-2 versions, so we need to check when the HPA v2beta2 bug was fixed.

stefanprodan:
So looks like the fix was shipped in Kubernetes 1.28 and backported to 1.27 and 1.26 so we should be good merging this.

matheuscscp (approved):
LGTM

PR #1187, opened 2026-04-24
runtime/cel: fix requiring int `.status.observedGeneration`

### LINKED ISSUE #5862: Fix requirement for integer `.status.observedGeneration` for CEL healthchecks
### Discussed in https://github.com/fluxcd/flux2/discussions/4508

<div type='discussions-op-text'>

<sup>Originally posted by **hrvatskibogmars** December 21, 2023</sup>
> Hello,
> 
> I'm currently encountering a type mismatch error when using **Istio** + **Flux** and would appreciate any guidance or suggestions to resolve it. The issue arises with the status.observedGeneration field, which is expected to be an integer (int64), but it's being interpreted or provided as a string (string). Here's the specific error message I'm receiving:
> 
> `Warning HealthCheckFailed 2m32s (x17 over 83m) kustomize-controller (combined from similar events): health check failed after 5m0.055393312s: timeout waiting for: [VirtualService/default/vs-front-imba status: 'Unknown': looking up status.observedGeneration from resource: .status.observedGeneration accessor error: 3 is of the type string, expected int64]`
> 
> virtual-service.yaml
> 
> ```
> apiVersion: networking.istio.io/v1alpha3
> kind: VirtualService
> metadata:
>   name: vs-api-imba
>   namespace: default
> spec:
>   gateways:
>   - gw-global-lb-portal-imba
>   hosts:
>   -imba.com
>   http:
>   - match:
>     - uri:
>         prefix: /
>     route:
>     - destination:
>         host: api
>         port:
>           number: 80
> ```

-------

https://github.com/fluxcd/flux2/discussions/4508#discussioncomment-16666205

> Hi @stefanprodan,
> 
> is there a way to bypass this type mismatch using `spec.healthCheckExprs`? We are using a custom CRD where `.status.observedGeneration` is intentionally typed as string, which causes Flux's built-in assessor to fail with:
> 
> `status: 'Unknown': .status.observedGeneration accessor error: 1 is of the type string, expected int64]`

matheuscscp:
Fixes: https://github.com/fluxcd/flux2/issues/5862

### DIFF
--- a/runtime/cel/status_evaluator.go
+++ b/runtime/cel/status_evaluator.go
@@ -88,18 +88,14 @@ func (s *StatusEvaluator) Evaluate(ctx context.Context, u *unstructured.Unstruct
 	unsObj := u.UnstructuredContent()
 
 	// Check if the object has the field status.observedGeneration
-	// and if it differs from metadata.generation, in which case we
-	// return status InProgress.
+	// as an int64 and if it differs from metadata.generation, in which
+	// case we return status InProgress. If the field is not an int64
+	// (e.g. some CRDs type it as a string), we skip this built-in check
+	// and rely solely on the user-provided CEL expressions.
 	observedGeneration, ok, err := unstructured.NestedInt64(unsObj, "status", "observedGeneration")
-	if err != nil {
-		return nil, err
-	}
-	if ok {
+	if err == nil && ok {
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

PR #1191, opened 2026-04-28
ssa: fix incorrect object reference on SkippedAction

matheuscscp:
Part of: https://github.com/fluxcd/kustomize-controller/issues/1647


matheuscscp:
I tested this manually in a kind cluster where I reproduced the issue and the fix works. Also added a test.

stefanprodan (approved):
LGTM

PR #1193, opened 2026-05-01
ssa: remove namespace from skipped cluster-scoped objects

matheuscscp:
Part of: https://github.com/fluxcd/kustomize-controller/issues/1647


matheuscscp:
We are not going to fix this, because the tradoff does not make sense. Using RESTMapper here requires elevated RBAC and will therefore not work under impersonation, plus uses a ton of memory and is a big hit on the API server. Also, the conditions that need to align for the issue to surface make it extremely rare:

* object must be cluster-scoped
* object must have `ssa: Ignore` or equivalent
* a kustomize feature that adds a namespace must be used

The impact of the bug is not real, as those objects are not healtchecked and are not garbage-collected. The only downside is showing a slightly wrong entry in the inventory.

PR #1208, opened 2026-05-21
fix: preserve all labels including invalid ones

raffis:
See https://github.com/fluxcd/flux2/pull/5906


matheuscscp (commented):
Same comments as in the other PR

stefanprodan:
This behavior is very intentional, the SSA package is used in all controllers and this change breaks them. For the CLI we could have some validation step set there not in this package.

PR #1211, opened 2026-05-21
auth/utils/cioidc: introduce high-level OIDC transport for CI systems

matheuscscp:
Moving this logic from `flux-mirror`.

This will be useful for `flux-mirror sync` and `flux push artifact`, but not for `flux trigger receiver` (which is one of the reasons why I'm keeping `auth/actionsoidc` and `auth/utils/cioidc` separate). The other reason is following the pattern of high-level abstractions placed under `auth/utils`, while the low-level ones, with more involved provider integration, as their own packages directly under `auth`.

### DIFF
--- a/auth/utils/cioidc/cioidc.go
+++ b/auth/utils/cioidc/cioidc.go
@@ -1,0 +1,187 @@
+/*
+Copyright 2026 The Flux authors
+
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

PR #1214, opened 2026-05-23
auth: introduce JWT signing with JWK for CI systems

matheuscscp:
Also move `auth/utils/cioidc` to `auth/utils/cijwt`.

This will be useful for `flux-mirror sync` and `flux push artifact`.

Similar to how the transport injects OIDC tokens retrieved from CI platforms like GitHub, GitLab and Forgejo (which are JWT tokens by the definition of OIDC), this PR introduces support for "bring your own key" JWT authentication in the `auth/jwt` package.

`WithHostJWK()` accepts a target host, a private JWK, and the `iss`, `aud` and `sub` claims. The transport will issue one JWT per request. This adds minimal overhead per request since the parameters are all in memory and no network calls are made for issuing a token (as opposed for example to `actionsoidc`). Because it's much cheaper to issue tokens like this, we make them 60sec-lived, which is awesome!

I have specific plans to use `flux-mirror sync` with this feature in my company.

### DIFF
--- a/auth/go.mod
+++ b/auth/go.mod
@@ -26,6 +26,7 @@ require (
 	github.com/fluxcd/pkg/apis/meta v1.27.0
 	github.com/fluxcd/pkg/cache v0.14.0
 	github.com/fluxcd/pkg/ssh v0.25.0
+	github.com/go-jose/go-jose/v4 v4.1.4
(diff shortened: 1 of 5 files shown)


stefanprodan (approved):
LGTM 🚀

PR #1217, opened 2026-05-25
fix(tools): respect manual major version bumps in go.mod

### LINKED ISSUE #1172: Improve `make release` to support major version bumps
After releasing  `tar` v1 GA we noticed that `make release` errors out requiring changes back to `v0` after the prep PR was merged with our manual bump from `v0` to `v1`:

https://github.com/fluxcd/pkg/pull/1169/commits

```
$ make release
Current branch: main
Release mode: minor
Bumped tar: 0.17.0 => 0.18.0 in modules: artifact, http/fetch, oci
Error: modules need to be bumped, please run 'make prep' first and open a pull request
make: *** [Makefile:155: release] Error 1
```

We need to change the code under `cmd/` invoked by `make release` to make it tolerate bumps like that where the current version in the module references across the codebase is greater than the tag it thinks should be the correct one. In this case, it should have seen that `v1.0.0` is greater than `v0.18.0` and should have accepted the release operation using the greater tag, i.e. `v1.0.0`.

Iam-Karan-Suresh:
fix(tools): respect manual major version bumps in go.mod

Previously, the release tool computed module version bumps and blindly 
replaced existing versions in go.mod files using regex. This caused 
manual major version bumps (e.g., v1.0.0) to be incorrectly downgraded 
to computed minor bumps (e.g., v0.18.0), resulting in false-positive 
modifications that blocked the release process.

This commit updates the apply function in module_bump.go to:
- Extract the existing version string from the target go.mod file.
- Parse both the existing and computed versions using semver.
- Dynamically adopt the existing version if it is mathematically 
  greater than the internally computed version.

Fixes #1172

### DIFF
--- a/cmd/internal/module_bump.go
+++ b/cmd/internal/module_bump.go
@@ -95,9 +95,22 @@ func (m *ModuleBump) apply(ctx context.Context, targetModule string, dryRun bool
 		return false, fmt.Errorf("failed to read %s: %w", gomod, err)
 	}
 	oldContent := string(b)
-	if !m.regex.MatchString(oldContent) {
+	// Extract current version from the go.mod line and compare with the computed new version.
+	matches := m.regex.FindStringSubmatch(oldContent)
+	if len(matches) == 0 {
 		return false, nil
 	}
+	if len(matches) > 1 {
+		if currentVer, err := semver.NewVersion(matches[1]); err == nil {
+			if newVer, err := semver.NewVersion(m.newVersion); err == nil {
+				if currentVer.GreaterThan(newVer) {
+					// Use the higher version already present (e.g., a manual major bump).
+					m.newVersion = currentVer.String()
+				}
+			}
+		}
+	}
+
 	bumpString := fmt.Sprintf("github.com/fluxcd/pkg/%s v%s", m.module, m.newVersion)
 	newContent := m.regex.ReplaceAllString(oldContent, bumpString)
 	if oldContent == newContent {


cmd/internal/module_bump.go
```diff
@@ -98,6 +98,19 @@ func (m *ModuleBump) apply(ctx context.Context, targetModule string, dryRun bool
 	if !m.regex.MatchString(oldContent) {
 		return false, nil
 	}
+// Extract current version from the go.mod line and compare with the computed new version.
+	matches := m.regex.FindStringSubmatch(oldContent)
```
matheuscscp: Why match twice? This call could replace the one right above, and `if len(matches) == 0` we `return false, nil`

cmd/internal/module_bump.go
```diff
@@ -95,9 +95,22 @@ func (m *ModuleBump) apply(ctx context.Context, targetModule string, dryRun bool
 		return false, fmt.Errorf("failed to read %s: %w", gomod, err)
 	}
 	oldContent := string(b)
-	if !m.regex.MatchString(oldContent) {
+	// Extract current version from the go.mod line and compare with the computed new version.
+	matches := m.regex.FindStringSubmatch(oldContent)
+	if len(matches) == 0 {
 		return false, nil
 	}
+	if len(matches) > 1 {
```
matheuscscp, line 103: Can `len(matches) == 1`?

PR #1218, opened 2026-05-25
chartutil: add Literal mode to ValuesReference (helm --set-literal semantics)

### LINKED ISSUE #1317: Feature Request: use helm --set-literal in valuesFrom
Currently, you need to escape some characters manually if you want to use valuesFrom + targetPath to insert a string from a configmap into a helm value that contains characters such as `,`

https://github.com/fluxcd/helm-controller/issues/853
https://github.com/fluxcd/flux2/issues/1756

This was solved in helm by creating the `--set-literal` flag: https://github.com/helm/helm/pull/9182

But I don't think you can set this in fluxcd's helm controller.

Would it be a good idea to set this by default?

gecube:
## Problem

When a `HelmRelease` consumes a `ConfigMap` or `Secret` via `valuesFrom` with a `targetPath` set, the referenced value is currently passed to Helm's `strvals.ParseInto` — the same parser as `helm --set`. That treats `,`, `[`, `]`, `{`, `}`, `=` and `\` in the value as syntactic metacharacters, so arbitrary file content (Spring `application.yml` with flow sequences, HOCON `application.conf`, JSON blobs, multi-line YAML with trailing commas, …) is misparsed or fails outright with errors like:

```
error parsing index: strconv.Atoi: parsing " \"prometheus\", \"health\", \"info\" ": invalid syntax
```
```
key "efm" has no value (cannot end with ,)
```

The single/double-quote workaround introduced in helm-controller#298 (`'<value>'` → `ParseIntoString`) is impractical for delivering raw config files via kustomize `configMapGenerator` / `secretGenerator`: the generated CM data is the file content as-is, with no opportunity to wrap it without sidecar files or build steps. Flux's kustomize-controller [explicitly disables Kustomize plugins](https://github.com/fluxcd/pkg/blob/main/kustomize/kustomize_generator.go) (`PluginConfig: kustypes.DisabledPluginConfig()`), so a transformer-based workaround isn't available either.

## Solution

Adds a `Literal bool` field to `meta.ValuesReference`. When `Literal: true` together with `TargetPath`, `ChartValuesFromReferences` calls a new `ReplacePathLiteralValue` helper instead of `ReplacePathValue`. The helper pre-escapes strvals metacharacters in the value, then delegates to `strvals.ParseIntoString` — that combination yields:

- Value preserved **verbatim** (no `,`/`[`/`{`/`=` interpretation; no type coercion to int/bool/list/map).
- Path escapes still honoured (`externalConfig.application\.yml.content` resolves to the literal key `application.yml`, which the alternative `strvals.ParseLiteralInto` does not).

`Literal` has no effect when `TargetPath` is empty (the root YAML-merge path is unchanged). Default `Literal: false` is fully backward compatible.

### Why not `strvals.ParseLiteralInto`?

It is the obvious choice and was the first thing I tried. It correctly preserves value metacharacters, but does **not** support `\.` escapes in the path:

```
externalConfig.application\.yml.content=hello
  → ParseIntoString:   externalConfig.application.yml.content = "hello"   ✓
  → ParseLiteralInto:  externalConfig["application\"]["yml"]["content"] = "hello"   ✗
```

Charts that use filenames as map keys (a common Helm pattern for per-file config) need the `\.` escape, so I went with pre-escaping the value and routing through `ParseIntoString`. Documented in the function doc comment.

## API

```yaml
spec:
  valuesFrom:
    - kind: ConfigMap
      name: my-service-config
      valuesKey: application.yml
      targetPath: 'externalConfig.application\.yml.content'
      literal: true   # NEW
```

## Tests

`chartutil/values_test.go` gains:

- **`TestChartValuesFromReferences`** — new cases:
  - `non-literal target path mangles value with commas` — pinning the current bug as a regression test.
  - `literal target path preserves helm-set metacharacters` — Spring-style YAML with flow sequence.
  - `literal target path preserves multi-line HOCON with equals and braces` — HOCON content read from a Secret.
  - `literal flag without targetPath is ignored (root YAML merge)` — documents the no-effect branch.
- **`TestReplacePathLiteralValue`** — 10 unit cases covering commas, braces, brackets, equals, multi-line YAML, HOCON, JSON, boolean-shaped strings, and the `\.` path escape.

All existing tests continue to pass.

## Backward compatibility

- New optional field with `omitempty`; defaults to `false`.
- DeepCopy is a no-op (`*out = *in`) — `ValuesReference` is a plain value type, no regeneration needed.
- Existing references behave identically (call site path picks `ReplacePathValue` when `Literal: false`).

## Follow-up

This PR ships the API + behaviour in `fluxcd/pkg`. A follow-up PR in `fluxcd/helm-controller` will:

1. Bump the `fluxcd/pkg` dependency to a release containing this change.
2. Regenerate `config/crd/bases/helm.toolkit.fluxcd.io_helmreleases.yaml` so the field is openAPI-validated by the API server.
3. Add a `valuesFrom literal` section to `docs/spec/v2/helmreleases.md`.

## Related

- Resolves [fluxcd/helm-controller#1317](https://github.com/fluxcd/helm-controller/issues/1317) (request for `--set-literal` semantics in `valuesFrom + targetPath`)
- Addresses parts of [#460](https://github.com/fluxcd/helm-controller/issues/460) (YAML / raw value in `valuesFrom`)
- Practical examples discussed in [#853](https://github.com/fluxcd/helm-controller/issues/853), [flux2#1756](https://github.com/fluxcd/flux2/issues/1756), where users hit the manual-escape wall.
- Helm landed `--set-literal` upstream in [helm#9182](https://github.com/helm/helm/pull/9182) (v3.12); this brings the equivalent to Flux.

### DIFF
--- a/apis/meta/reference_types.go
+++ b/apis/meta/reference_types.go
@@ -218,6 +218,17 @@ type ValuesReference struct {
 	// transient error will still result in a reconciliation failure.
 	// +optional
 	Optional bool `json:"optional,omitempty"`
+
+	// Literal marks this ValuesReference as a literal value. When set in
+	// combination with TargetPath, the referenced value is merged at the target
+	// path without interpreting Helm's `--set` syntax (commas, brackets, dots,
+	// equal signs, etc.), mirroring the behavior of `helm --set-literal`. This
+	// is the only safe way to inject arbitrary file content (config files, JSON
+	// blobs, multi-line strings containing special characters) through
+	// `valuesFrom`. Has no effect when TargetPath is empty: in that mode the
+	// referenced value is always YAML-merged at the root.
+	// +optional
+	Literal bool `json:"literal,omitempty"`
 }
 
 // GetValuesKey returns the defined ValuesKey, or the default ('values.yaml').
--- a/chartutil/values.go
+++ b/chartutil/values.go
@@ -236,7 +236,11 @@ func ChartValuesFromReferences(ctx context.Context, log logr.Logger, client kube
 			// TODO(hidde): this is a bit of hack, as it mimics the way the option string is passed
 			// 	to Helm from a CLI perspective. Given the parser is however not publicly accessible
 			// 	while it contains all logic around parsing the target path, it is a fair trade-off.
-			if err := ReplacePathValue(result, ref.TargetPath, string(valuesData)); err != nil {
+			merger := ReplacePathValue
+			if ref.Literal {
+				merger = ReplacePathLiteralValue
+			}
+			if err := merger(result, ref.TargetPath, string(valuesData)); err != nil {
 				return nil, NewErrValuesReference(namespacedName, ref, ErrValueMerge, err)
 			}
 			continue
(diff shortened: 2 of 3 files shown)


stefanprodan:
Does this solves https://github.com/fluxcd/flux2/issues/2625?

gecube:
@stefanprodan Hi! I think that yes, I did not find that issue before, but came from the other end - I got a very weird error when trying to put the configmap in yaml format as a key with valuesFrom. So now it looks like that in fact it is the same thing

stefanprodan:
Please add a test that replicates https://github.com/fluxcd/flux2/issues/2625 and let's see if it passes

gecube:
Ok, thanks, I will do it tomorrow and return asap

gecube:
@stefanprodan please check

matheuscscp:
@gecube Please rebase and force-push :pray:

gecube:
@matheuscscp done, thanks

matheuscscp:
@gecube Strange, I still see the `Update branch` button... Maybe you just squashed? What I usually do is the following:

`git pull --rebase origin main`

Assuming `origin` is the `github.com/fluxcd/pkg` remote (not your fork)

matheuscscp (approved):
LGTM

PR #1221, opened 2026-05-28
auth/utils/cijwt: introduce `WithHostTokenFunc`

matheuscscp:
Parameterize the API for JWTs that are issued through a network call. `actionsoidc.FetchToken` will now be passed in transparently.

This moves the 50% lifetime cache logic to the transport, to avoid all the different `FetchToken` functions that will be passed repeating the JWT parsing for extracting the expiration (for the cache logic).

This will be useful in `flux push artifact` and `flux-mirror sync` for fetching OIDC tokens from cloud providers like GCP and Azure (AWS does not support it). In this case, the token will be used to authenticate on a third party service that accepts generic OIDC (these are not IAM credentials for targeting GCP/Azure services, those are covered by `auth/gcp` and `auth/azure` instead).

### DIFF
--- a/auth/actionsoidc/actionsoidc.go
+++ b/auth/actionsoidc/actionsoidc.go
@@ -32,9 +32,6 @@ import (
 	"net/url"
 	"os"
 	"strings"
-	"time"
-
-	"github.com/golang-jwt/jwt/v5"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


stefanprodan (approved):
LGTM

PR #1224, opened 2026-06-01
Add DriftIgnoreRules to DiffOptions for drift detection

dipti-pai:
Wire DriftIgnoreRules into the Diff() method so that the CLI diff path (flux diff kustomization) can leverage the same field-level ignore rules already supported by the apply path.

When DriftIgnoreRules is set, Diff() compiles the rules once, delegates to hasDriftedWithIgnore instead of hasDrifted, and strips ignored fields from the returned objects so that the caller's diff output only shows meaningful drift.

Assisted-by: GitHub Copilot/Claude Opus 4.6

Related to: https://github.com/fluxcd/flux2/issues/5884

### DIFF
--- a/ssa/manager_diff.go
+++ b/ssa/manager_diff.go
@@ -25,6 +25,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/fluxcd/pkg/ssa/errors"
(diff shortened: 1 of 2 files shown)


matheuscscp (approved):
LGTM

matheuscscp:
@dipti-pai `ssa/v0.75.0` is tagged :ok_hand:

Issue #603, opened 2023-07-08
feature request: add support for providing propagationPolicy (i.e orphan) when deleting immutable resources


mdnfiras:
When setting the `kustomize.toolkit.fluxcd.io/force: enabled` label for statefulsets, flux will delete the statefulset if it detects an immutable field error. This will delete all the statefulset's pods as a result due to kubernetes garbage collection.

As a workaround, an admission webhook can mutate pods' `metadata.ownerReferences` to be empty during pod creation, but admission webhooks are risky: if they fail (service is down, or throwing unhandled exceptions) they block the creation of pods, and if set to be ignored upon failure then we would risk flux deleting a statefulset and its pods all at once.

I think that a proper solution would be to add support to set the `propagationPolicy` of the `deleteOptions` in the delete request that flux makes. This can be set for example through an annotation like `kustomize.toolkit.fluxcd.io/propagationPolicy: orphan` in the same resource that has `kustomize.toolkit.fluxcd.io/force: enabled`:
* if the label is set, then the annotation is checked
* if the annotation does not exist, the default policy is used, which is `background`
* if it is set to `background`, `foreground` or `orphan` then this value will be used in the delete request
* if it is set to anything other than those values, the controller can throw an error during reconciliation dry-run.

what do you think about this feature?

kingdonb:
Is one of the intended or possible uses of a `foreground` propagationPolicy to be able to wait for deletes? I think I have seen users request control over the deletion order when it comes to dependency and dependsOn, I just wonder if this might address that concern when it gets into kustomize-controller.

stefanprodan:
No this will not address dependsOn delete ordering. Waiting for resources to be deleted is not a viable option IMO. I’m for adding an annotation to control the propagation policy, but it can’t be only for force as proposed here, it would cover deletion no matter why it happens, force apply, GC, Kustomization deletion, etc.

mdnfiras:
@kingdonb actually my main concern is to have the `orphan` option, since I intend to use the new *immutable field resource deletion feature* of flux with statefulsets, but as flux deletes and re-creates a statefulset it will result in an instant deletion of all its pods at the same time.

@stefanprodan 
> Waiting for resources to be deleted is not a viable option IMO
> The annotation would cover deletion no matter why it happens

I agree with both statements, so I guess `foreground` is out of the question since it waits for owned resources to be deleted first, so its kinda "blocking". 

[...]

PR #1227, opened 2026-06-02
Support force delete propagation policy

martinohansen:
## What changed

Adds support for `kustomize.toolkit.fluxcd.io/propagationPolicy` when force-applying immutable resource changes. The annotation accepts `background` and `orphan`; when it is absent, the delete request leaves `PropagationPolicy` unset so the Kubernetes API server default applies.

## Why

Force apply previously hardcoded background deletion. This prevented users from orphaning dependents during forced replacement and also meant Flux carried its own default instead of relying on Kubernetes when no annotation was set.

## Validation

- `KUBEBUILDER_ASSETS=... go test -run TestForceApplyDeleteOptions -count=1`
- `go test -c`
- `git diff --check`

`make test-ssa` was attempted, but the full envtest suite failed in existing Kubernetes 1.36 force-apply and API migration cases unrelated to the focused propagation-policy coverage.

### DIFF
--- a/ssa/manager_apply.go
+++ b/ssa/manager_apply.go
@@ -42,6 +42,8 @@ import (
 	"github.com/fluxcd/pkg/ssa/utils"
 )
 
+const forceApplyPropagationPolicyAnnotation = "kustomize.toolkit.fluxcd.io/propagationPolicy"
+
 // ApplyOptions contains options for server-side apply requests.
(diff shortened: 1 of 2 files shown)


martinohansen:
Fixes https://github.com/fluxcd/pkg/issues/603

PR #1229, opened 2026-06-04
feat(api): extend DependencyReference type to support any Kubernetes resource

vecil:
### Summary

The shared `DependencyReference` type now supports depending on any Kubernetes object, improving flexibility in release ordering and readiness handling.

Dependencies can reference arbitrary Kubernetes resources and are considered satisfied based on either existence or readiness (via the new `ready` field). The `runtime/dependency` sort package has been updated to operate on typed references, so dependencies across different `apiVersion`/`kind` combinations are correctly ordered and cycle-checked. The new `TypedNamespacedObjectReference` type carries the same shape as the previous `NamespacedObjectReference` plus the `apiVersion` and `kind` fields, and `MakeDependsOn` now parses the new `apiVersion/Kind/[namespace/]name[:ready][@readyExpr]` format.

Changes include:

- Extend `DependencyReference` with optional `apiVersion`, `kind`, and `ready` fields, and add a `TypedNamespacedObjectReference` type for typed cross-resource references
- Extend `MakeDependsOn` to parse `apiVersion/Kind/[namespace/]name[:ready][@readyExpr]` strings, with backward-compatible parsing of the existing `name` and `namespace/name` formats
- Render the new fields in `DependencyReference.String()` so dependencies round-trip correctly
- Update `runtime/dependency.Sort` to return `TypedNamespacedObjectReference`, require `GetAPIVersion()`/`GetKind()` on the `Dependent` interface, and only default the dependency namespace when the kind matches the parent
- Improve tests to cover arbitrary resource types, the `ready` toggle, and cross-kind dependency graphs

Relates fluxcd/flux2#5879


runtime/dependency/sort.go
```diff
@@ -39,40 +41,46 @@ const (
...
-// NamespacedObjectReference based on their dependencies. It performs a
+// TypedNamespacedObjectReference based on their dependencies. It performs a
 // topological sort using a depth-first search algorithm, which has
 // runtime complexity of O(|V| + |E|), where |V| is the number of
 // vertices (objects) and |E| is the number of edges (dependencies).
 //
 // Reference:
 // https://en.wikipedia.org/wiki/Topological_sorting#Depth-first_search
-func Sort(objects []Dependent) ([]meta.NamespacedObjectReference, error) {
+func Sort(objects []Dependent) ([]meta.TypedNamespacedObjectReference, error) {
```
matheuscscp, line 41: Is `NamespacedObjectReference` still used anywhere after this change?

apis/meta/reference_types.go
```diff
@@ -43,18 +65,50 @@ func (in NamespacedObjectReference) String() string {
...
-	// object that contains the reference.
+	// Namespace of the resource to depend on, defaults to the namespace of the Flux
+	// Applier API resource (Kustomization or HelmRelease) that contains the reference.
 	// +optional
 	Namespace string `json:"namespace,omitempty"`
 
+	// Ready checks if the resource Ready status condition is true, defaults to
+	// true when the dependency is a Flux Applier API resource (Kustomization or HelmRelease).
+	// +optional
+	Ready *bool `json:"ready,omitempty"`
```
matheuscscp: We thought about this field and I think we can leave it out since the default for Flux appliers would be `true` while for RSET the default is `false`. If one  wishes to check only for existence, they  can set `ReadyExpr` to the expression `true`.

apis/meta/reference_types.go
```diff
@@ -43,18 +65,50 @@ func (in NamespacedObjectReference) String() string {
...
+	// APIVersion of the resource to depend on, defaults to the API group version of the
+	// Flux Applier API resource (Kustomization or HelmRelease) that contains the reference
+	// when the dependency is of the same kind.
+	// +optional
+	APIVersion string `json:"apiVersion,omitempty"`
+
+	// Kind of the resource to depend on, defaults to the kind of the
+	// Flux Applier API resource (Kustomization or HelmRelease) that contains the reference.
+	// +optional
+	Kind string `json:"kind,omitempty"`
```
matheuscscp, line 94: Because most API types would normally require a custom `ReadyExpr` anyway, we would like to make it so that if the API  group (ignoring the version) or the kind differ from the type managed by the respective controller, then `ReadyExpr` becomes mandatory.

This  means everything except `Kustomization` will require `ReadyExpr` in kustomize-controller, and everything except `HelmRelease` will require `ReadyExpr` in helm-controller.

Maybe we can add xvalidation rule with a CEL expression to the `DependsOn` field in  each controller to enforce this.

runtime/dependency/sort.go
vecil: It will remain in use in the `image-reflector-controller` API (via the `ImageRepositoryRef` field type) and its tests, the `image-automation-controller` tests, and the Flux CLI `create image policy` command (through the `image-ref` argument).

matheuscscp: Thanks for the investigation!

apis/meta/reference_types.go
matheuscscp: @stefanprodan Vincent brought a good point here, if the feature gate `AdditiveCELDependencyCheck` is enabled then it's impossible to check only for existence without this field

vecil: There may be very rare use cases where one would want to check only for existence though... I understand the motivation for simplifying the API change proposal.

apis/meta/reference_types.go
vecil: If the `ReadyExpr-only` approach is adopted, this proposal would provide a clear separation of behavior between HR-to-HR/KS-to-KS and cross-resource dependencies, and would reduce the cognitive load of writing such dependency rules 👍

matheuscscp:
@vecil I think we can try and implement as much as possible of this feature in this repo. Besides the new types/fields in `apis/meta`, we can add also more logic to `runtime/dependency` as much as possible to slim down the code in the controllers. Unfotunately we will need the xvalidations in the respective `DependsOn` struct fields of each controller due to them being different, as they depend on the respective type managed by that controller. But I think a large part of the code could be implemented in `runtime/dependency`, maybe even existing code in the controllers.

One more thing: we need the schema validations that will be applied via xvalidations in the `DependsOn` struct fields of each controller to also be checked in the `Reconcile()` loop, this is a standard we have all across Flux to keep validation tight and error messages clear. This is so old API servers not supporting xvalidation still get guarded asynchronously after admission.

apis/meta/reference_types.go
vecil: Following agreement that `AdditiveCELDependencyCheck` is additive, meaning the built-in readiness check is not intended to be bypassed when enabled, the proposed `ready` field has been dropped from the change proposal.

apis/meta/reference_types.go
vecil: On second thought, I have the feeling this proposal would introduce additional complexity when defining dependencies on resource types other than Appliers. Built-in readiness checks based on `fluxcd/cli-utils/pkg/kstatus/status` types and functions may already cover many use cases.

However, the proposed xvalidation rule introduces a limitation: it would require `readyExpr` to always be set, even when no custom check is desired. This prevents relying solely on built-in readiness checks in cases where they would otherwise be sufficient.

- When the `AdditiveCELDependencyCheck` feature gate is enabled, built-in readiness checks will always run alongside required `readyExpr`, with only the overhead of specifying a value (e.g. `"true"`).
- When the gate is disabled, requiring `readyExpr` via xvalidation would prevent users from relying on built-in readiness checks alone.

PR #1238, opened 2026-06-08
fix(kustomize): reset global OpenAPI schema for each build

### LINKED ISSUE #1237: Custom `openapi:` configuration can silently corrupt strategic-merge-patch for future builds
A custom `openapi:` schema in one Kustomization corrupts strategic-merge in later builds (kustomize OpenAPI global is never reset between builds)

# What happens

kustomize-controller builds every Kustomization in one long-lived process. kustomize's OpenAPI schema (used for strategic-merge keys) is a process-global that `fluxcd/pkg/kustomize.Build` never resets between builds.

So if one Kustomization sets a **trimmed** custom schema (via `openapi: { path: ... }`) that omits built-in types, and it builds first after a restart, every later build loses strategic-merge for those types and **replaces lists instead of merging them**. For us this rendered a `StatefulSet` with **no container `image`** and dropped sidecars → `spec.containers[0].image: Required value` → outage. Flux reported success.

It stays broken for the life of the process (a plain `flux reconcile` re-applies the corrupt build). Only a restart, or a later build with a full schema / `openapi: version:`, fixes it.

# Reproduce

Two dirs built in order through the real `fluxcd/pkg/kustomize.Build`:

- `poisoner/`: `kustomization.yaml` with `openapi: { path: schema.json }`, where `schema.json` is a CRD-only schema (no `StatefulSet`).
- `victim/`: a normal StatefulSet (image only in the base) + a strategic-merge patch that omits `image`. No `openapi:`.

```go
openapi.ResetOpenAPI()                               // fresh process
kustomize.Build(filesys.MakeFsOnDisk(), "poisoner")  // sets trimmed schema
m, _ := kustomize.Build(filesys.MakeFsOnDisk(), "victim")
y, _ := m.AsYaml()
// y has no "image:" — the base container was dropped (list-replace)
```

Build `victim` first and it's fine. Reproduces on `fluxcd/pkg/kustomize` v1.32.0 (kustomize v0.21.1)

# Fix

Reset the global around each build. Builds are already serialised by `kustomizeBuildMutex`, so this is safe:

```go
func Build(fs filesys.FileSystem, dirPath string) (resmap.ResMap, error) {
    kustomizeBuildMutex.Lock()
    defer kustomizeBuildMutex.Unlock()
    defer openapi.ResetOpenAPI()   // <-- add: don't leak this build's schema into the next
    ...
}
```

dan-j:
The included test fails if `openapi.ResetOpenAPI()` is not called, not sure on the test implementation doing string matching, but I couldn't see an existing dependency that would give better reflection.

cc @matheuscscp 

Fixes #1237

### DIFF
--- a/kustomize/kustomize_generator.go
+++ b/kustomize/kustomize_generator.go
@@ -36,6 +36,7 @@ import (
 	"sigs.k8s.io/kustomize/api/resmap"
 	kustypes "sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
+	"sigs.k8s.io/kustomize/kyaml/openapi"
 	"sigs.k8s.io/yaml"
 
 	"github.com/fluxcd/pkg/apis/kustomize"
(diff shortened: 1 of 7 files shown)


stefanprodan (approved):
LGTM

Thanks @dan-j

matheuscscp (approved):
LGTM! 🚀 

Nice one!

PR #1239, opened 2026-06-10
oci: set created timestamp in the config

matheuscscp:
Fixes: https://github.com/fluxcd/flux2/discussions/5934


stefanprodan (approved):
LGTM

PR #1244, opened 2026-06-12
auth/utils: ensure registry credentials are always fresh

matheuscscp:
This PR fixes a bug **(not a vuln!)**, but I have never seen it happen, there are no user reports that I'm aware of.

Long-running operations using artifact registry credentials for multiple OCI/HTTP calls can hit 401, since credentials are statically resolved by `auth.GetArtifactRegistryCredentials()`. That function returns a user+pass pair immediately (due to how `auth/{aws|azure|gcp}` are implemented for ECR/ACR/GAR). In other words, the `Authorize()` method of the returned `authn.Authenticator` will return a cached user+pass pair that never changes. This fix delays the call to `auth.GetArtifactRegistryCredentials()` to happen on demand inside `authn.Authenticator.Authorize()`, which ensures that multiple calls will always return a fresh credential. This is because `auth.GetArtifactRegistryCredentials()` has a smart cache inside that keeps track of the credential lifetime. This makes it never return an expired credential, as a new one is created when the cached one is expired (the cache knows when).

This fix is super good for `flux-mirror sync`, which calls `auth/utills.GetArtifactRegistryCredentials()` (not the low-level one from `auth`) for ECR/ACR/GAR only once during the process startup and hands the `authn.Authenticator` off to `go-containerregistry`. Before this fix, `go-containerregistry` would call `Authorize()` again internally upon a 401 and get the same credentials acquired during process startup. This PR causes `Authorize()` to return valid creds on every call.

I don't think SC or IRC would ever hit this bug, since their operations are not super long. The reconcilers just perform a handful of calls, and no OCI artifact is large enough that it takes 2-digit minutes to reconcile. But this PR fixes the whole thing anyway.

### DIFF
--- a/auth/utils/registry.go
+++ b/auth/utils/registry.go
@@ -24,6 +24,17 @@ import (
 	"github.com/fluxcd/pkg/auth"
 )
 
+// delegatedAuthenticator is an authn.Authenticator that delegates
+// calls for Authorization to a provided function.
+type delegatedAuthenticator struct {
+	authorizationFunc func() (*authn.AuthConfig, error)
+}
+
+// Authorization implements authn.Authenticator.
+func (d *delegatedAuthenticator) Authorization() (*authn.AuthConfig, error) {
+	return d.authorizationFunc()
+}
+
 // GetArtifactRegistryCredentials retrieves the registry credentials for the
 // specified artifact repository and provider.
 func GetArtifactRegistryCredentials(ctx context.Context, providerName string,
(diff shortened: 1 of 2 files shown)


matheuscscp:
Note: We already do this for other operations that are known to be long-running, like reconciling YAML on remote clusters:

https://github.com/fluxcd/pkg/blob/3f9f27f0ba23cf7953c25b7eba7ccac1b525f60e/auth/utils/restconfig.go#L126-L160

matheuscscp:
Integration tests are running:

* AWS: https://github.com/fluxcd/pkg/actions/runs/27447762507
* Azure: https://github.com/fluxcd/pkg/actions/runs/27447765194
* GCP: https://github.com/fluxcd/pkg/actions/runs/27447768254

Edit: All passed!

matheuscscp:
I have deployed this change through a source-controller release candidate in my EKS clusters (where all my sources are OCIRepository) and they are all healthy. Will keep it there until I can switch to the 2.9 release.

stefanprodan (approved):

PR #1246, opened 2026-06-16
kustomize: introduce option for always substituting

matheuscscp:
Part of: https://github.com/fluxcd/kustomize-controller/issues/1011

Also adding tests for https://github.com/fluxcd/kustomize-controller/pull/1671.


stefanprodan (approved):
LGTM

PR #1260, opened 2026-07-08
kustomize: load builtin schemas after reset

matheuscscp:
xref: https://github.com/fluxcd/flux2/issues/5981 (Avoiding `Fixes` to keep the issue open until a release)

Turns out we dont need to traverse the artifact and look for the `openapi` field to merge them onto the builtin schemas. We can just load them after the reset.


kustomize/kustomize_generator.go
```diff
@@ -651,12 +630,17 @@ func adaptSelector(selector *kustomize.Selector) (output *kustypes.Selector) {
 // buildMutex protects against kustomize concurrent map read/write panic
 var kustomizeBuildMutex sync.Mutex
 
-// SecureBuild wraps krusty.MakeKustomizer with the following settings:
+func resetOpenAPIWithBuiltins() {
+	openapi.ResetOpenAPI()
+	_ = openapi.Schema()
```
matheuscscp, line 635: This simple `openapi.Schema()` call is the actual fix for https://github.com/fluxcd/flux2/issues/5972

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #1266, opened 2026-07-14
kustomize: preserve existing image fields when merging spec.images

gecube:
Fixes the image tag loss reported in fluxcd/flux2#5787.

When the `kustomization.yaml` at the Kustomization's `spec.path` already contains an `images` entry for the same image name, the generator replaced the whole entry with the one from `spec.images`, silently dropping any field set only in the existing entry. With `spec.images` setting only `newName`, the `newTag` applied by the `kustomization.yaml` was lost and the build produced an untagged image reference — diverging from how a kustomize overlay handles a partial image override (root cause analysis in https://github.com/fluxcd/flux2/issues/5787#issuecomment-4965566628).

This changes the generator to merge the fields into the existing entry instead. `newName` is merged independently, while `newTag` and `digest` are updated as a single unit: setting one clears the other. The pairing mirrors the kustomize image transformer, which replaces both original values whenever either of them is overridden ([`api/filters/imagetag/updater.go#L39-L53`](https://github.com/kubernetes-sigs/kustomize/blob/api/v0.21.1/api/filters/imagetag/updater.go#L39-L53)), and the `Image.Digest` API contract ("If digest is present NewTag value is ignored"). Without the pairing, an existing `digest` merged with an incoming `newTag` would produce `name:tag@digest`, where the stale digest silently wins at pull time while the manifest reads as updated.

| # | existing entry | `spec.images` | before (replace) | after (merge) |
|---|---|---|---|---|
| 1 | `newTag: v1.2.3` | `newName: reg/app` | `reg/app` — tag lost | `reg/app:v1.2.3` |
| 2 | `newName: reg/app` | `newTag: v2.0.0` | `app:v2.0.0` — newName lost | `reg/app:v2.0.0` |
| 3 | `digest: sha256:d1` | `newName: reg/app` | `reg/app` — digest lost | `reg/app@sha256:d1` |
| 4 | `digest: sha256:d1` | `newTag: v2.0.0` | `app:v2.0.0` | `app:v2.0.0` — digest cleared, not kept |
| 5 | `newTag: v1.2.3` | `digest: sha256:d2` | `app@sha256:d2` | `app@sha256:d2` — tag cleared, not kept |
| 6 | `newName: reg/app`, `digest: sha256:d1` | `newTag: v2.0.0` | `app:v2.0.0` — newName lost | `reg/app:v2.0.0` — newName kept, digest cleared |
| 7 | `newTag: v1.2.3` | `newTag: v2.0.0`, `digest: sha256:d2` | `app:v2.0.0@sha256:d2` | `app:v2.0.0@sha256:d2` — both set explicitly, both applied |
| 8 | `newName: reg/app`, `newTag: v1.2.3` | *(only `name`)* | `app` — all overrides lost | `reg/app:v1.2.3` — no-op |

The cross-axis rows (1–3, 6) are the actual bug fix: `newName` selects *where* the image lives, `newTag`/`digest` select *which version*, and a partial override of one axis no longer clobbers the other. On the version axis (rows 4–5) the merge intentionally behaves like the old replace, so switching between tag and digest pinning never leaves a stale value of the other field behind.

`Test_Images` covers every row above — asserting both the merged `kustomization.yaml` entry and the final image reference rendered through a real kustomize build — plus a full override of all fields and appending an entry with no existing match. The behavior was also reproduced end-to-end against kustomize-controller v1.8.5 and v1.9.3 (https://github.com/gecube/flux2-5787-repro).

### DIFF
--- a/kustomize/kustomize_generator.go
+++ b/kustomize/kustomize_generator.go
@@ -299,16 +299,27 @@ func (g *Generator) GenerateManifest(dirPath string) ([]byte, string, Action, er
 	}
 
 	for _, image := range images {
-		newImage := kustypes.Image{
-			Name:    image.Name,
-			NewName: image.NewName,
-			NewTag:  image.NewTag,
-			Digest:  image.Digest,
-		}
 		if exists, index := checkKustomizeImageExists(kus.Images, image.Name); exists {
-			kus.Images[index] = newImage
+			// Merge the individual fields into the existing entry, so that
+			// fields set only in the kustomization.yaml (e.g. newTag) are
+			// preserved when the Kustomization overrides a subset of them,
+			// matching the behavior of a kustomize overlay.
+			if image.NewName != "" {
+				kus.Images[index].NewName = image.NewName
+			}
+			if image.NewTag != "" {
+				kus.Images[index].NewTag = image.NewTag
+			}
+			if image.Digest != "" {
+				kus.Images[index].Digest = image.Digest
+			}
 		} else {
-			kus.Images = append(kus.Images, newImage)
+			kus.Images = append(kus.Images, kustypes.Image{
+				Name:    image.Name,
+				NewName: image.NewName,
+				NewTag:  image.NewTag,
+				Digest:  image.Digest,
+			})
 		}
 	}
 
--- a/kustomize/kustomize_generator.go
+++ b/kustomize/kustomize_generator.go
@@ -307,10 +307,8 @@ func (g *Generator) GenerateManifest(dirPath string) ([]byte, string, Action, er
 			if image.NewName != "" {
 				kus.Images[index].NewName = image.NewName
 			}
-			if image.NewTag != "" {
+			if image.NewTag != "" || image.Digest != "" {
 				kus.Images[index].NewTag = image.NewTag
-			}
-			if image.Digest != "" {
 				kus.Images[index].Digest = image.Digest
 			}
 		} else {
--- a/kustomize/kustomize_generator.go
+++ b/kustomize/kustomize_generator.go
@@ -307,6 +307,15 @@ func (g *Generator) GenerateManifest(dirPath string) ([]byte, string, Action, er
 			if image.NewName != "" {
 				kus.Images[index].NewName = image.NewName
 			}
+			// NewTag and Digest both select the image version, so they are
+			// updated as a single unit: setting one clears the other. This
+			// mirrors the kustomize image transformer, which replaces both
+			// original values whenever either is overridden (see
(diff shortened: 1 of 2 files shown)


kustomize/kustomize_generator.go
```diff
@@ -299,16 +299,27 @@ func (g *Generator) GenerateManifest(dirPath string) ([]byte, string, Action, er
...
+			// matching the behavior of a kustomize overlay.
+			if image.NewName != "" {
+				kus.Images[index].NewName = image.NewName
+			}
+			if image.NewTag != "" {
+				kus.Images[index].NewTag = image.NewTag
+			}
+			if image.Digest != "" {
+				kus.Images[index].Digest = image.Digest
+			}
```
matheuscscp: We should probably always update tag and digest together:

```suggestion
			if image.NewTag != "" || image.Digest != "" {
				kus.Images[index].NewTag = image.NewTag
				kus.Images[index].Digest = image.Digest
			}
```

gecube: Applied, thanks — and I think this is more than a consistency nit. It makes the merge semantics match what kustomize itself does at the string level: the image transformer replaces tag and digest together whenever either of them is overridden ([`updater.go#L39-L53`](https://github.com/kubernetes-sigs/kustomize/blob/api/v0.21.1/api/filters/imagetag/updater.go#L39-L53), *"overriding tag or digest will replace both original tag and digest values"*), and the `Image.Digest` API doc already promises *"If digest is present NewTag value is ignored"*. With the two independent `if`s, an existing `digest` merged with an incoming `newTag` would have produced `name:tag@digest`, where the stale digest silently wins at pull time while the manifest reads as updated — probably the worst failure mode for a GitOps tool, since nothing errors and the drift is invisible in Git.

I've extended `Test_Images` with the cross tag↔digest cases (both directions, `newName` isolation from the pair update, both fields set together, and a name-only no-op entry) and updated the table in the PR description to document the pairing.

One case I'd like an explicit decision on: existing `newTag: v1.2.3` + incoming `digest` only (row 5 in the description). As implemented, the tag is cleared and the entry renders `app@sha256:…`. The alternative — keeping the tag so it renders `app:v1.2.3@sha256:…` — would still pull the correct image (the digest wins), but leaves a tag that no longer necessarily corresponds to the digest, which reads as misleading metadata. I went with clearing, consistent with updater.go; if you'd rather preserve the tag as a human-readable hint there, happy to change it — users who want the `tag@digest` form can always set both fields explicitly in `spec.images`.

stefanprodan (approved):
LGTM

Thanks @gecube

matheuscscp (approved):
LGTM

PR #1274, opened 2026-07-29
Fix Azure e2e tests after new major version with breaking changes

matheuscscp:
`hashicorp/azurerm` shipped v5.0 yesterday with lots of breaking changes. This PR mitigates the breakage by pinning to below v5.0. Later we can upgrade to the new version with the required changes.


matheuscscp:
Running here: https://github.com/fluxcd/pkg/actions/runs/30452834381

stefanprodan (approved):

PR #1276, opened 2026-08-06
ssa: Add pre-dry-run field manager takeover

### LINKED ISSUE #1722: ssa: support taking over field managers before dry-run
## Summary

`ResourceManager.Apply` / `ApplyAll` [cleans up field managers](https://github.com/fluxcd/pkg/blob/main/ssa/manager_apply.go#L154) (`cleanupMetadata` → `PatchReplaceFieldsManagers`) only after a successful server-side dry-run. When a stale/foreign field manager co-owns a field such that the merged object is invalid, the dry-run fails first and `Apply` returns an error. As a result, `--override-manager` in kustomize-controller can resolve valid-but-wrong co-ownership, but cannot recover an object that is wedged at dry-run validation, even though stripping the offending manager would make the merge valid.

A workload is managed by Flux via server-side apply. During an incident an operator disables Flux and patches the workload with client-side PUTs from a legacy controller (operation: Update). Re-enabling Flux reconciliations leaves a stale `legacy-controller / Update` managedFields entry co-owning a container's identity ("." + f:name) but NOT its required f:image. When a later build removes that container, server-side apply cannot prune it (the legacy Update manager still owns the element) but drops the field Flux solely owned (f:image), leaving a half-removed, invalid container: `spec.template.spec.containers[N].image: Required value`

This fails at the dry-run on every reconcile, before any write. The standard post-dry-run FieldManagers cleanup never runs, so --override-manager cannot recover it.

## Potential fix

Add a separate, opt-in list to run the field-manager takeover before `dryRunApply`, reusing the existing `PatchReplaceFieldsManagers`. 

The [sequence of steps that lead to the issue](https://github.com/fluxcd/pkg/pull/1276/changes#diff-49950ef3bf2043a603f8c7a8fbc6c60db938e0806a64bfc507fa6e6f759610df) and the proposed fix are in the draft PR https://github.com/fluxcd/pkg/pull/1276. 

### Why a separate list 

We discussed this in the dev meeting on August 6th with @matheuscscp and @stealthybox , and running the takeover before the dry-run sacrifices a real property of the current design where a failed/invalid dry-run, `Apply` returns at `NewDryRunErr` and `cleanupMetadata` never runs, so managedFields are left untouched when the apply doesn't proceed. A pre-dry-run takeover mutates ownership first, so if the dry-run then fails for an unrelated reason (a different missing required field, a webhook rejection), ownership has already been reassigned on an object whose desired state never applied.

Ther always-on default managers that kustomize-controller already reclaims post-dry-run (`kubectl` Apply/Update, `before-first-apply`, and the controller's own `Update`) stay on the post-dry-run path to minimize the blast radius of the change.

### kustomize-controller wiring

A dedicated flag `--override-manager-before-dry-run=<manager>` populates `FieldManagersBeforeDryRun` with only those managers (both Apply and Update ops). It is kept separate from `--override-manager` so operators opt specific known-problematic managers into the stronger, pre-validation behavior; the default managers and ordinary `--override-manager` entries keep today's post-dry-run semantics.

dipti-pai:
Add ApplyCleanupOptions.FieldManagersBeforeDryRun, a separate opt-in list of field managers whose ownership is reclaimed before the server-side dry-run in Apply and ApplyAll.

The existing FieldManagers takeover runs only after a successful dry-run (cleanupMetadata). When a stale or foreign manager co-owns a required field such that the merged object is invalid, the dry-run fails first and the takeover is never reached, leaving the object wedged on every reconcile. Taking over the named managers before the dry-run repairs ownership so the merge is valid.

This is kept separate from FieldManagers because a pre-dry-run takeover mutates managedFields even if the apply later fails, unlike the post-dry-run cleanup. Only managers explicitly identified as problematic should be listed, so the default managers keep their transactional post-dry-run behavior.

Assisted-by: GitHub Copilot/Claude Opus 4.8

Fixes: https://github.com/fluxcd/kustomize-controller/issues/1722

### DIFF
--- a/ssa/manager_apply.go
+++ b/ssa/manager_apply.go
@@ -102,6 +102,18 @@ type ApplyCleanupOptions struct {
 	// FieldManagers defines which `metadata.managedFields` managers should be removed from in-cluster objects.
 	FieldManagers []FieldManager `json:"fieldManagers,omitempty"`
 
+	// FieldManagersBeforeDryRun defines which `metadata.managedFields` managers
+	// are taken over on the in-cluster object BEFORE the server-side dry-run.
+	//
+	// Use this to reclaim a required field co-owned by a stale or foreign manager
+	// that would otherwise be orphaned on prune and fail the dry-run before the
+	// post-dry-run cleanup can run.
+	//
+	// It is a deliberately narrow, opt-in list, separate from FieldManagers,
+	// because taking over ownership before the dry-run mutates managedFields even
+	// if the apply later fails — unlike the post-dry-run cleanup.
+	FieldManagersBeforeDryRun []FieldManager `json:"fieldManagersBeforeDryRun,omitempty"`
+
 	// Exclusions determines which in-cluster objects are skipped from cleanup
 	// based on the specified key-value pairs.
 	Exclusions map[string]string `json:"exclusions"`
@@ -136,6 +148,20 @@ func (m *ResourceManager) Apply(ctx context.Context, object *unstructured.Unstru
 		if err != nil {
 			return nil, fmt.Errorf("%s failed to migrate API version: %w", utils.FmtUnstructured(existingObject), err)
 		}
+	}
+
+	// Take over fields held by stale or foreign managers before the dry-run so
+	// that the merged object is valid. Without this, a manager co-owning a
+	// required field can leave that field orphaned when the applier prunes it,
+	// failing the dry-run before the post-dry-run cleanup can run.
+	if len(opts.Cleanup.FieldManagersBeforeDryRun) > 0 && getError == nil {
+		tookOver, err := m.takeoverManagedFields(ctx, object, existingObject,
+			opts.Cleanup.FieldManagersBeforeDryRun, opts.Cleanup)
+		if err != nil {
+			return nil, fmt.Errorf("%s metadata.managedFields takeover failed: %w",
+				utils.FmtUnstructured(existingObject), err)
+		}
+		patched = patched || tookOver
 	}
 
 	dryRunObject := object.DeepCopy()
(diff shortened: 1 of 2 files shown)


ssa/manager_apply_takeover_test.go
matheuscscp, line 1: We should add tests for ApplyAll as well. Always prefer ApplyAll, as that's what kustomize-controller uses.

ssa/manager_apply.go
```diff
@@ -154,6 +181,18 @@ func (m *ResourceManager) Apply(ctx context.Context, object *unstructured.Unstru
 		}
 	}
 
+	// Take over fields held by stale or foreign managers before the dry-run so
+	// that the merged object is valid. Without this, a manager co-owning a
+	// required field can leave that field orphaned when the applier prunes it,
+	// failing the dry-run before the post-dry-run cleanup can run.
+	tookOver, err := m.takeoverFieldManagersBeforeDryRun(ctx, object, existingObject,
```
matheuscscp, line 205: Also in ApplyAll.

On top of explaining the behavior, this comment should have the same structure of the comment we added for `MigrateAPIVersion`: it should contain the harm it can cause to justify why it should not always be enabled, only in recovery situations (turn on to recover the cluster, then turn off).

ssa/manager_apply.go
```diff
@@ -281,6 +320,16 @@ func (m *ResourceManager) ApplyAll(ctx context.Context, objects []*unstructured.
 					}
 				}
 
+				// Take over fields held by stale or foreign managers before the
+				// dry-run so that the merged object is valid (see Apply for details).
+				tookOver, err := m.takeoverFieldManagersBeforeDryRun(ctx, object, existingObject,
```
matheuscscp, line 363: Ditto.

matheuscscp (commented):
Thanks @dipti-pai, looks good overall!

dipti-pai:
> Thanks @dipti-pai, looks good overall!

Thanks @matheuscscp. Addressed the review comments.

matheuscscp (approved):
LGTM! 🚀 

Thanks @dipti-pai!

Issue #1294, opened 2026-09-08
knownhosts: HostKeyAlgorithms are not scoped to the target host, making SSH host key verification fail non-deterministically


adonaicosta:
### Describe the bug

`ssh/knownhosts.New()` returns the host key algorithms to advertise in
`ssh.ClientConfig.HostKeyAlgorithms`, but `hostKeyAlgorithms()` collects them from
**every** entry in the `known_hosts` blob, without matching them against the host being
dialled, and builds the slice by ranging over a map, so the order is random on every call:

https://github.com/fluxcd/pkg/blob/main/ssh/knownhosts/knownhosts.go#L168-L180

```go
func (db *inMemoryHostKeyDB) hostKeyAlgorithms() []string {
	var algos []string

	uniq := make(map[string]struct{})
	for _, hk := range db.hostKeys {   // every host in the blob
		uniq[hk.key.Type()] = struct{}{}
	}
	for k := range uniq {              // map range: random order
		algos = append(algos, k)
	}

	return algos
}
```

When a single `known_hosts` blob covers more than one host and those hosts have different
sets of key types, the client advertises algorithms it holds no entry for on the host it is
actually connecting to. The server is free to pick one of them (RFC 4253: the first
algorithm in the client list that the server also supports), and `checkAddr` then rejects
it, by design:

https://github.com/fluxcd/pkg/blob/main/ssh/knownhosts/knownhosts.go#L349-L352

```go
	// If the remote host starts using a different, unknown key type, we
	// also interpret that as a mismatch.
	if known, ok := knownKeys[remoteKey.Type()]; !ok || !keyEq(known, remoteKey) {
		return keyErr
	}
```

Because the advertised order is randomised per call, the same GitRepository succeeds or
fails depending on which permutation `New()` happens to produce. Users see a `GitRepository`
that flips between `Ready=True` and

```
failed to checkout and determine revision: unable to list remote for 'ssh://git@github.com/org/repo.git': ssh: handshake failed: knownhosts: key mismatch
```

with no change to the secret, the network or the server, which reads like an egress or DNS
problem and is very hard to diagnose.

This is a regression introduced by #943 (`ssh v0.19.0`, `git/gogit v0.35.0`, shipped in
Flux v2.7.0), which added `HostKeyAlgorithms` to fix fluxcd/flux2#5385. The fix is right for
the single-host case; it just needs to be scoped to the host being dialled.

### Steps to reproduce

[...]

PR #1295, opened 2026-09-08
knownhosts: scope HostKeyAlgorithms to the target host

adonaicosta:
hostKeyAlgorithms collected the key types of every entry in the known_hosts
blob, without matching them against the host being dialled, and built the
slice by ranging over a map. A blob covering several hosts with different
key-type sets therefore advertised algorithms unknown for the host being
connected to, in a random order on every call. The server may pick one of
them, which checkAddr then rejects as a key mismatch, so verification
succeeded or failed depending on the permutation.

Match the entries against the target address, skip certificate authority
entries, which hold the signing key rather than a host key algorithm the
server can offer, and sort the result so the negotiation is deterministic.
Return nil when no entry matches, leaving the negotiation to the defaults of
golang.org/x/crypto/ssh.

New() now takes the connection target, and gogit passes AuthOptions.Host.

Regression from #943, which is correct for a single host per blob.

Signed-off-by: Adonai Costa <adonai.costa@gmail.com>

### DIFF
--- a/git/gogit/transport.go
+++ b/git/gogit/transport.go
@@ -73,7 +73,7 @@ func transportAuth(opts *git.AuthOptions, fallbackToDefaultKnownHosts bool) (tra
 		var callback gossh.HostKeyCallback
 		var hkAlgos []string
 		if len(opts.KnownHosts) > 0 {
-			callback, hkAlgos, err = knownhosts.New(opts.KnownHosts)
+			callback, hkAlgos, err = knownhosts.New(opts.KnownHosts, opts.Host)
 			if err != nil {
 				return nil, err
 			}
--- a/ssh/knownhosts/knownhosts.go
+++ b/ssh/knownhosts/knownhosts.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"sort"
 	"strings"
 
 	"golang.org/x/crypto/ssh"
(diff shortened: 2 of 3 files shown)


adonaicosta:
Fixes the non-deterministic `knownhosts: key mismatch` described in (https://github.com/fluxcd/pkg/issues/1294).

`ssh/knownhosts.hostKeyAlgorithms()` collected the key types of **every** entry in the
`known_hosts` blob, without matching them against the host being dialled, and built the
slice by ranging over a map:

```go
	uniq := make(map[string]struct{})
	for _, hk := range db.hostKeys {   // every host in the blob
		uniq[hk.key.Type()] = struct{}{}
	}
	for k := range uniq {              // map range: random order
		algos = append(algos, k)
	}
```

A blob covering several hosts with different key-type sets therefore advertises algorithms
for which the database holds no entry for the host actually being connected to. The server
is free to pick one (RFC 4253), and `checkAddr` then rejects it as designed. Because the
order is randomised per call, the same GitRepository flips between `Ready=True` and
`ssh: handshake failed: knownhosts: key mismatch` with nothing changing around it.

### Changes

- `hostKeyAlgorithms` now takes an `addr` and returns only the algorithms held for it,
  sorted so the negotiation order is deterministic.
- Certificate authority entries are skipped: they hold the key that signs host
  certificates, not a host key algorithm the server can offer.
- Returns `nil` when no entry matches the host, leaving the negotiation to the defaults of
  `golang.org/x/crypto/ssh` rather than advertising a list that cannot match.
- `New` takes the connection target (`host`, with or without a port); `git/gogit` passes
  `AuthOptions.Host`, which `Validate()` already requires for the SSH transport.

`New` is a signature change, but `git/gogit/transport.go` is its only caller in the tree,
and #943 already changed this signature one release ago.

### Behaviour

Same `known_hosts` blob (GitLab with three key types, GitHub with one), same real
`github.com`, 9 consecutive dials:

| | before | after |
|---|---|---|
| host key accepted | 1/9 | 9/9 |
| advertised algos | random permutation of all 3 types | `[ecdsa-sha2-nistp256]` every time |

fluxcd/flux2#5385, the issue that #943 fixed, stays fixed: a blob holding only
`ssh-rsa` for its host still advertises `ssh-rsa`.

### Tests

`TestHostKeyAlgorithmsScopedToHost` covers host scoping, the `host` vs `host:port` forms, a
non-matching port, an unknown host, and repeats each case 20 times to catch the random
order. `TestHostKeyAlgorithmsIgnoresCertAuthority` covers the `@cert-authority` case. Both
fail against the previous implementation:

```
--- FAIL: TestHostKeyAlgorithmsScopedToHost (0.00s)
    knownhosts_test.go:355: host github.com: got [ecdsa-sha2-nistp256 ssh-ed25519], want [ecdsa-sha2-nistp256]
--- FAIL: TestHostKeyAlgorithmsIgnoresCertAuthority (0.00s)
    knownhosts_test.go:370: got [ssh-ed25519 ecdsa-sha2-nistp256], want [ecdsa-sha2-nistp256]
```

`go test ./...` passes in `ssh`, and `go build ./... && go vet ./gogit/ && go test ./gogit/`
passes in `git`.

Issue #1291, opened 2026-09-04
feat(auth/githubapp): support client ID as alternative to numeric app ID for JWT iss claim


Erik-Schuetze:
Title: feat(auth/githubapp): support client ID as alternative to numeric app ID for JWT iss claim

---

GitHub's documentation now recommends using the client ID (e.g. `Iv23liXXXXXXX`) instead of
the numeric app ID as the `iss` claim when generating JWTs for GitHub App authentication:
[docs page for github.com](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app) 
[docs page for GHE 3.20](https://docs.github.com/en/enterprise-server@3.20/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app) 

Currently `auth/githubapp/client.go` only accepts a numeric `githubAppID`, parsed with
`strconv.ParseInt` and used as the integer JWT issuer.

**Problem**

Users migrating from PATs to GitHub Apps are storing credentials using the client ID because
that is what GitHub now surfaces and recommends. Having to separately look up the numeric app ID
just for Flux is unnecessary friction, and the numeric ID is an internal identifier that GitHub
themselves are steering users away from.

**Proposed change**

Add an optional `githubAppClientID` secret key. If present and `githubAppID` is absent, use it
as the string `iss` claim in the JWT. The GitHub API accepts both forms.

Concretely in `auth/githubapp/client.go`:
- Add `KeyAppClientID = "githubAppClientID"` constant
- Store it as a `clientID string` field on `Client`
- In `WithAppData`, populate it when present
- In `New`, accept either `appID != 0` or `clientID != ""` as valid
- In `createJWT`, use the client ID string as `iss` when app ID is zero

No breaking change -- `githubAppID` continues to work exactly as today.

**Context**

This came up while migrating a Flux `GitRepository` from a PAT to a GitHub App on a GitHub
Enterprise Server 3.20 instance. When creating the GitHub App, the UI prominently shows the
client ID (e.g. `Iv23liXXXXXXX`) as the app identifier. The numeric app ID is a secondary field
that users have to actively look for. Supporting `githubAppClientID` in the secret would let users
use the credential GitHub presents to them directly, without an extra lookup step.

I would be happy to submit a PR if the approach looks reasonable.

matheuscscp:
The approach looks reasonable to me.

[...]

PR #1296, opened 2026-09-08
Add `githubAppClientID` as an alternative to `githubAppID`

Erik-Schuetze:
Refs: https://github.com/fluxcd/pkg/issues/1291

GitHub accepts either the numeric AppID or the ClientID of GH Apps as the JWT `iss` claim, and recommends the ClientID in the official docs ([GitHub](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app), [GHE](https://docs.github.com/en/enterprise-server@3.20/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app)). `pkg` currently only accepts the numeric AppID, so this PR adds `githubAppClientID` as an alternative across `auth/githubapp` (the JWT consumer) and `runtime/secrets` (the secret writer and reader). Existing secrets using only AppID are unaffected.

AppID and ClientID are treated as mutually exclusive. This is enforced in `New`, `MakeGitHubAppSecret`, and `GitHubAppDataFromSecret`, mirroring the existing `installationOwner` / `installationID` handling. 
They identify the same app for a single `iss` claim, so this keeps the behaviour unambiguous and avoids silently ignoring one identity when both are supplied. 
The ClientID is treated as a regular string, since GitHub documents no format for it (The ClientIDs I checked on GHE and GitHub also differed from each other with no common prefix). 
As `pkg` is a GA dependency consumed by all Flux controllers, `MakeGitHubAppSecret` keeps `appID` as a positional parameter and adds `clientID` via an option, because any change to the positional parameters of this exported function would break every existing consumer. 
If the preferred solution is to change to symmetric options (`WithGitHubAppID` + `WithGitHubAppClientID`) and accept the function signature break, I'm happy to switch.

Two behaviour changes to note, given the GA compatibility guarantee:
- changed the error message of `MakeGitHubAppSecret` from `"githubAppID is required"` to `"exactly one of githubAppID or githubAppClientID must be provided"`.
- `buildCacheKey` now includes the ClientID, which causes a new token to be requested on version upgrade (a one-time, self-healing re-fetch of short-lived installation tokens).

I applied the change across `auth/githubapp` and `runtime/secrets` in `pkg` so the new ClientID is usable end-to-end. If you prefer I can split it into two PRs. 

I added Tests for client-ID-only, both-set (rejected), and neither-set across writer, reader, and JWT issuer generation; `make test-auth` and `make test-runtime` pass.

### DIFF
--- a/runtime/secrets/factory.go
+++ b/runtime/secrets/factory.go
@@ -238,29 +238,42 @@ func WithGitHubAppBaseURL(baseURL string) GitHubAppOption {
 	}
 }
 
+// WithGitHubAppClientID sets the client ID for the GitHub App secret.
+func WithGitHubAppClientID(clientID string) GitHubAppOption {
+	return func(data map[string]string) {
+		data[KeyGitHubAppClientID] = clientID
+	}
+}
+
 // MakeGitHubAppSecret creates a Kubernetes secret for GitHub App authentication.
 //
-// The function requires appID and privateKey to be non-empty.
+// The function requires privateKey to be non-empty, and exactly one of appID or clientID.
 // Exactly one of installationOwner or installationID must be provided.
 // Optional baseURL can be provided for GitHub Enterprise Server instances.
 // The resulting secret will be of type Opaque.
 func MakeGitHubAppSecret(name, namespace, appID, privateKey string,
 	opts ...GitHubAppOption) (*corev1.Secret, error) {
 
-	if err := validateRequired(appID, KeyGitHubAppID); err != nil {
-		return nil, err
-	}
 	if err := validateRequired(privateKey, KeyGitHubAppPrivateKey); err != nil {
 		return nil, err
 	}
 
 	data := map[string]string{
-		KeyGitHubAppID:         appID,
 		KeyGitHubAppPrivateKey: privateKey,
+	}
+	if appID != "" {
+		data[KeyGitHubAppID] = appID
 	}
 
 	for _, opt := range opts {
 		opt(data)
+	}
+
+	_, hasAppID := data[KeyGitHubAppID]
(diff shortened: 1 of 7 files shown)


runtime/secrets/factory.go
```diff
@@ -238,31 +238,44 @@ func WithGitHubAppBaseURL(baseURL string) GitHubAppOption {
...
+func WithGitHubAppClientID(clientID string) GitHubAppOption {
+	return func(data map[string]string) {
+		data[KeyGitHubAppClientID] = clientID
+	}
+}
+
 // MakeGitHubAppSecret creates a Kubernetes secret for GitHub App authentication.
 //
-// The function requires appID and privateKey to be non-empty.
+// The function requires privateKey to be non-empty, and exactly one of appID or clientID.
```
matheuscscp, line 250: Please remove the `appID` positional argument and make it an option too

Erik-Schuetze:
appID is now a WithGitHubAppID option. I left the validation unchanged as it was already checking for exactly one of appID/clientID.

As discussed this is a breaking change to the exported MakeGitHubAppSecret signature (positional appID dropped), so downstream callers will need updating.

The force push was just to add the missing DCO sign-off to the commit.

matheuscscp:
@Erik-Schuetze Please rebase with main (no merge commits!), squash all commits into one, and force-push

Erik-Schuetze:
Done

matheuscscp (approved):
LGTM! 🚀 

Thanks @Erik-Schuetze, nice addition!

PR #1298, opened 2026-09-14
kustomize: add ContextWithSecretVarsCollector for LoadVariables

dipti-pai:
SubstituteVariables calls LoadVariables internally, once per built resource, to resolve spec.postBuild.substituteFrom Secret values. A caller that needs those same values for another purpose had to call LoadVariables a second time.

Add ContextWithSecretVarsCollector, which registers a callback on the context that LoadVariables invokes with the secret vars it resolves, in addition to returning them as before. The callback receives an independent copy of the map, so it is unaffected by the inline postBuild.substitute overlay that SubstituteVariables merges in afterwards.

Assisted-by: GitHub Copilot/Claude Sonnet 5


matheuscscp (approved):
LGTM

